PR #12, opened 2022-08-16
Add conda nvcc installation instructions to compile_from_source.md

aninrusimha:
If users don't have sudo access, or want to have different cuda versions for each conda environment, it can be helpful to install nvcc on a per environment basis. It is non-intuitive which of the many conda package (nvcc, cuda_runtime, etc.) you need to install, so I thought it would be useful to include that info in the source compilation instructions. This is how I installed the package locally, and can upload my conda env file if its wanted :)


TimDettmers:
This is a great addition, thank you! One issue with the conda install is that only a limited number of CUDA versions are supported.

We also offer a script to download and install a variety of CUDA versions locally: [cuda_install.sh](https://github.com/TimDettmers/bitsandbytes/blob/main/cuda_install.sh). Could you add instructions on how to use this?

PR #22, opened 2022-08-22
Make 8-bit matrix multiplication compatible with cpuonly builds

mryab:
While trying to use the latest release of bitsandbytes in hivemind, I faced an error after installing the library in a cpuonly mode (both from PyPI and from source): https://github.com/learning-at-home/hivemind/runs/7961913748?check_suite_focus=true#step:8:56

The gist of this error seems to be unprotected usage of `torch.cuda.get_device_capability()`, which fails for non-GPU builds of PyTorch. To handle this, I've moved the corresponding imports in the root of bitsandbytes inside the COMPILED_WITH_CUDA block and also moved the relevant function/class declarations inside the similar block to avoid accidental imports.

The solution appears to work locally (at least the tests pass), happy to improve on the PR if any additional work is needed.

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -1,10 +1,12 @@
 import operator
-import torch
-import bitsandbytes as bnb
-import bitsandbytes.functional as F
-
 from dataclasses import dataclass
 from functools import reduce  # Required in Python 3
+
+import torch
+
+import bitsandbytes.functional as F
+from bitsandbytes.cextension import COMPILED_WITH_CUDA
+
 
 # math.prod not compatible with python < 3.8
 def prod(iterable):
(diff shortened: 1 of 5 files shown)


bitsandbytes/autograd/_functions.py
```diff
@@ -165,231 +167,228 @@ def backward(ctx, grad_output):
...
 
-        return grad_A, grad_B, None, grad_bias, None
+            # 4. Mixed-precision decomposition matmul
+            if coo_tensorA is not None and subA is not None:
+                output += torch.matmul(subA, state.subB)
 
+            # 5. Save state
+            ctx.state = state
 
-matmul = MatMul8bitLt.apply
```
mryab, line 381: This gets immediately redefined below

bitsandbytes/functional.py
```diff
@@ -1685,18 +1684,6 @@ def double_quant(
...
-    if major < 7:
-        print(
-            f"Device with CUDA capability of {major} not supported for 8-bit matmul. Device has no tensor cores!"
-        )
-        assert major >= 7
-
-    if major == 7: return 'col_turing'
-    elif major == 8: return 'col_ampere'
-    else: return 'col_turing'
-
```
mryab, line 1699: This is already defined above (L186)

TimDettmers:
Thank you, Max! We chatted about this and I used your suggestions and fixed this in 7e0fb655e1e040221054886fbee9d5682aa6e4e2 and ee5b947e63c2340405f25e4e83066f39292bc0ed and it has been deployed already.

PR #23, opened 2022-08-23
Memory efficient 8bit backward

dbaranchuk:


bitsandbytes/autograd/_functions.py
```diff
@@ -366,10 +357,24 @@ def backward(ctx, grad_output):
...
+                if state.has_fp16_weights:
+                    CBt = state.CBt
+                else:
+                    # Restore CBt from CB
+                    assert state.CBt is None, "CBt should not be stored in state"
+                    CB = state.CB.half()
+                    SCB = state.SCB.unsqueeze(1).half()
+                    SCBt = state.SCBt.unsqueeze(1).half()
+                    Bt = (CB * SCB).t().contiguous()
+                    CBt = (Bt / SCBt).t().to(torch.int8)
```
justheuristic, line 369: [opinion] this block of code looks like the most memory-intensive; perhaps we can optimize some of it with in-place divisions / transpositions (please ping me if you're interested)

TimDettmers:
This version of backward had too high of an error to be useful. As such, we will not merge it into main. See #33 for the new memory efficient tbackwards that uses fp16 computation and which seems to work fine.

PR #33, opened 2022-09-11
Memory efficient backward

dbaranchuk:
TODO #1: double check that both memory_efficient_backward and has_fp16_weights options work properly. 
TODO #2: make a clearer PR description.

This PR provides two features:

1. Memory efficient backward option:

* Stores int8 weights in row-major format;
* In forward pass, we have to cast row-major matrix to Turing/Ampere at each iteration. This leads to noticeable computational overhead for inference. Thus, we suggest using this option only for training. 
* In backward pass, we transform row-major weight matrix to the fp16 weights and perform fp16 matmul with fp16 grad_outputs. Note that we do not store fp16 weights and just efficiently compute them on the fly. Therefore, overall computational and memory overheads are negligible. 

2. Cast inputs and outputs of the 8bit layer to fp16 and back to the initial input dtype, respectively. This allows us to use models of arbitrary dtypes without their conversion to fp16.


bitsandbytes/autograd/_functions.py
```diff
@@ -225,9 +226,10 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState()):
         input_shape = A.shape
         if state.outlier_pool is None:
             state.outlier_pool = GlobalOutlierPooler.get_instance()
-        assert (
-            A.dtype == torch.float16
-        ), f"The input data type needs to be fp16 but {A.dtype} was found!"
+
+        # Cast A to fp16
+        A_dtype = A.dtype
+        A = A.to(torch.float16)
```
TimDettmers, line 232: This is exactly what we talked about, this should take care of bfloat16! One thing we need to think about is some kind of warning or something when this is first executed to make people aware that a cast happens and the operation quantization is performed in fp16.

Additionally, it would be great to have bf16 tests that verify everything works correctly with those inputs. I think you just need to change one [line](https://github.com/TimDettmers/bitsandbytes/blob/main/tests/test_autograd.py#L256) (and check if everything else is correctly executed).

TimDettmers (changes requested):
Looks good to me overall. Please include also a test that tests for minimal difference between fp16 gradients and int8->fp16 gradients. You can add these tests either to the [existing tests](https://github.com/TimDettmers/bitsandbytes/blob/main/tests/test_autograd.py#L297) or create a new one.

bitsandbytes/autograd/_functions.py
justheuristic: - added a warnings.warn the first time A is cast to a different dtype
- added FP16 tests; fine-tuned typecasts to fit into *almost* all thresholds
  - increased atol 0.0175 -> 0.02 in one specific case for BF16 (this does not affect FP16)
  - e.g. grad_bias is now computed on natural grad_output before it is cast (and before it loses precision)

bitsandbytes/autograd/_functions.py
```diff
@@ -307,7 +308,13 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState()):
...
+
+        if bias is None or bias.dtype == torch.float16:
+            output = F.mm_dequant(out32, Sout32, SCA, state.SCB, bias=bias)
+            output = output.to(A.dtype)
+        else:  # apply bias separately
+            output = F.mm_dequant(out32, Sout32, SCA, state.SCB, bias=None)
+            output = output.to(A.dtype).add_(bias)
 
         # 4. Mixed-precision decomposition matmul
         if coo_tensorA is not None and subA is not None:
```
justheuristic, line 320: curiously, when i tried to replace the next line from
`output += torch.matmul(subA, state.subB)`

to 
`output.addmm_(subA, state.subB)`

the precision would drop. I have no idea why.

justheuristic:
Ran all tests 10 times to check for stability

image
image

bitsandbytes/autograd/_functions.py
```diff
@@ -375,21 +378,14 @@ def backward(ctx, grad_output):
...
                 gradA32, SgradA32 = F.igemmlt(C32grad, state.CxBt, Sgrad, state.SBt)
-                grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape)
+                grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape).to(ctx.dtype_A)
+
             elif state.CB is not None:
-                CB = state.CB.half()
-                SCB = (state.SCB.unsqueeze(1) / 127.0).half()
-                CB *= SCB
-                grad_A = torch.mm(grad_output, CB).view(ctx.grad_shape)
+                CB = state.CB.to(ctx.dtype_A, copy=True).mul_(state.SCB.unsqueeze(1).div(127.0))
```
TimDettmers, line 384: Not sure how PyTorch implements div, but multiplication is about 30x faster than division. Since we apply it over a matrix this might make a tiny but significant difference. So `.mul(1/127.0)` might be better here.

tests/test_modules.py
```diff
@@ -545,11 +552,30 @@ def test_linear8bitlt_no_fp16_weights(threshold):
...
+        o1 = mlp(b1)
+        assert o1.dtype == torch.float16
+        assert o1.requires_grad
+        grad_proj = torch.randn_like(o1)
+
+        mlp.zero_grad()
+        (o1 * grad_proj).sum().backward()
+        grad_ref = grad_proj.flatten(2) @ w2.half() @ w1.half()
+        scale = grad_ref.abs().mean()
+        assert torch.allclose(b1.grad, grad_ref, rtol=0, atol=0.05 * scale)
```
TimDettmers, line 572: I remember I had some tests that were using relative difference normalized by the standard deviation, which is similar to this. What is the range of errors that you see? It might also be good to test for a maximum of `k` elements that exceed a threshold. This helps to differentiate worse-case vs general performance.

bitsandbytes/autograd/_functions.py
TimDettmers: I cannot remember if I stumbled upon the same thing. I remember trying to make this matrix multiplication more efficient but failed. What is the increase that you see in errors? 

It does not make much sense to me since in cuBLAS you perform (A @ B) + D = C and the results of A @ B is in fp32 so the entire operation should be _more_ precise. The same goes for fused multiply-add in general, which is more precise than multiplication followed by addition. It might be some weird tensor core issue, but it makes no sense to me. 

If the error is only smaller some of the time and it has more variance, it would still be okay to have this. I believe it would be a good chunk faster.

TimDettmers (commented):
Looks all good to me. Let's discuss this briefly tomorrow. I am curious if we can get the `.addmm_` to work. Otherwise, just a couple of questions on the test performance. Overall great work! Thank you so much, Yozh!

bitsandbytes/autograd/_functions.py
justheuristic: applied, thanks

tests/test_modules.py
justheuristic: got it, i've added a separate assert with outliers:
```
        torch.testing.assert_allclose(b1.grad, grad_ref, rtol=0, atol=0.05 * scale)
        idx = torch.isclose(b1.grad, grad_ref, atol=0.01 * scale, rtol=0.1)
        assert (idx == 0).sum().item() <= b1.numel() * 0.005
```

TimDettmers (approved):

TimDettmers:
All good! Thank you both!

PR #74, opened 2022-10-27
Fix call to get_instance

manugarri:
parenthesis were missing to the function call `get_instance`


ulmewennberg:
This is addressed by https://github.com/TimDettmers/bitsandbytes/pull/72

manugarri:
glad it got fixed! better add some tests for a package that its used all over the GenAI world :D

PR #75, opened 2022-10-27
Typo that triggers error in conda env

Dango233:
Should be `CUDASetup.get_instance().add_log_entry` instead of `CUDASetup.get_instance.add_log_entry`


Dango233:
Addressed by #72 

Hence closed

PR #87, opened 2022-11-04
Add `device` and `dtype` parameters to `StableEmbedding`

lostmsu:
This allows using `StableEmbedding` anywhere where `torch.nn.Embeding` could be used


TimDettmers:
Thank you! Also, the layer norm cast to default precision is a great addition!

Issue #97, opened 2022-11-18
ERROR: Your GPU does not support Int8 Matmul!


lixiepeng:
- nvidia-smi
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 515.65.01    Driver Version: 515.65.01    CUDA Version: 11.7     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla P40           Off  | 00000000:02:00.0 Off |                  Off |
| N/A   29C    P8    10W / 250W |     24MiB / 24576MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+
|   1  Tesla P40           Off  | 00000000:03:00.0 Off |                  Off |
| N/A   26C    P8     9W / 250W |     40MiB / 24576MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
+-----------------------------------------------------------------------------+
```
- python -m bitsandbytes
```
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
++++++++++++++++++++ DEBUG INFORMATION +++++++++++++++++++++
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

++++++++++ POTENTIALLY LIBRARY-PATH-LIKE ENV VARS ++++++++++
'CONDA_EXE': '~miniconda3/bin/conda'
'CONDA_PREFIX': '~/miniconda3'
'CONDA_PYTHON_EXE': '~/miniconda3/bin/python'
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

WARNING: Please be sure to sanitize sensible info from any such env vars!

++++++++++++++++++++++++++ OTHER +++++++++++++++++++++++++++
COMPILED_WITH_CUDA = True
COMPUTE_CAPABILITIES_PER_GPU = ['6.1', '6.1']
[...]

PR #110, opened 2022-12-29
Improve cc version detection for cublaslt

BlackHC:
The current code fails to detect cublaslt for new cc versions (e.g. 8.9 fails the check).

This fixes issue #97.

### DIFF
--- a/bitsandbytes/cuda_setup/main.py
+++ b/bitsandbytes/cuda_setup/main.py
@@ -143,7 +143,14 @@ def evaluate_cuda_setup():
         cuda_setup.add_log_entry(f'CUDA SETUP: Detected CUDA version {cuda_version_string}')
 
     # 7.5 is the minimum CC vor cublaslt
-    has_cublaslt = cc in ["7.5", "8.0", "8.6"]
+    has_cublaslt = False
+    if cc is not None:
+        cc_major, cc_minor = cc.split('.')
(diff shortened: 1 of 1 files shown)


TimDettmers:
Thank you for this contribution!

gururise:
I don't think this PR quite fixes issue #97, which referenced lack of Int8 Matmul error being thrown on the Nvidia Tesla P40.

Still getting the following error:
```
=============================================
ERROR: Your GPU does not support Int8 Matmul!
=============================================

python: /home/gene/dockerx/temp/bitsandbytes/csrc/ops.cu:408: int igemmlt(cublasLtHandle_t, int, int, int, const int8_t*, const int8_t*, void*, float*, int, int, int) [with int FORMATB = 3; int DTYPE_OUT = 32; int SCALE_ROWS = 0; cublasLtHandle_t = cublasLtContext*; int8_t = signed char]: Assertion `false' failed.
```

I created a new issue #118 to make it clear that problem still exists.

BlackHC:
@gururise 

The following has helped me better understand what was going on, but your error sounds very different to the one I fixed with my PR.

```
logging.basicConfig(level=logging.DEBUG)
from bitsandbytes.cextension import CUDASetup
CUDASetup.get_instance().print_log_stack()
```

Best wishes,
 Andreas

PR #115, opened 2023-01-04
Fix for python 3.7

kashif:
and when CC is less than 7.5

I believe this is not a big change to fix #84

PR #117, opened 2023-01-05
Fix undefined variable in cuda_setup

indigoviolet:
It looks like this instance was missed in some refactor


kashif:
@indigoviolet this is what i did in #115

TimDettmers:
I merged #115 and this should be fixed now. Thanks for your contribution!

PR #127, opened 2023-01-22
issues/30 Windows Compiling

centerionware:
Summary:
Added cmake files
Rename pythonInterface.c to pythonInterface.cpp
Added template<I, T> struct InstrFloatTraits { typedef T vec_t; }
Added #ifndef _MSC_VER to not include unistd.h when MSVC is the compiler used
Added batch script to run all steps to build - including downloading pthread dependency and building it. The only thing it doesn't do by itself is launch the visual studio build environment. 
Modified Makefile for pythonInterface.c->.cpp rename

The change in ops.cu isn't necessary, it just removes the warning when compiling in release mode because assert is ignored in windows on release so it yells that the variable is never used.

*NEEDS TESTING!* Creating pull request to gain awareness, hopefully someone else can verify it works(specifically the batch script. I ran it by hand and copied to the script after. I haven't run the script alone yet.). Also, unsure of if InstrFloatTraits should be defined as above, or the way it is in the #30 issue..


kjerk:
Thanks for getting the ball rolling, this is exciting! I've been using bitsandbytes precompiled for CUDA 11.6 for a while but with the newer Lovelace cards we need 11.8 working properly, so getting the builds working would be great.

Following along your example Windows_Compile_Howto.bat, everything works up until the actual compilation for me and then I get failures still. Building pthread worked. Errors in kernels.cu it looks like:
```
bitsandbytes\csrc\kernels.cu(2667): error : no instance of function template "kspmm_coo_very_sparse_naive" matches the specified type [C:\Dev\Clo
ned\bitsandbytes\build\csrc\bitsandbytes.vcxproj]
```
[Entire output from clean to build](https://gist.github.com/kjerk/13afdec31920a6c83ba27f6fe6f3c6d9)

image

Environment:
* Visual Studio 2022 Command Prompt (Tried 2019 as well, exact same result). Bootstrapped into cmd with `call "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"`
* CMAKE 3.25.2
* Conda: None, Python 3.10 only (Cuda/Cudnn managed on path).
* CUDA 11.8 / cuDNN 8.7.0 on path (Verified with nvcc --version)
* CUDA_HOME / CUDA_PATH / CMAKE_CUDA_COMPILER all defined in Environment variables for safety.

centerionware:
I installed 11.8 - and 12.0 to try, and got the same results as you. 
Am not sure the problem is with the build scripts however, and think it may a compatibility issue with new compilers being more aggressive checking for __restrict__ const in general. 
I managed to fix this issue by opening the files Kernel.cu and ops.cu, and switching the parts that error. 
Not sure if this would be proper PR for this problem.
from:
```float *dequant_stats```
to:
```float * __restrict__ const dequant_stats```

kjerk:
Okay thanks to your comment I was able to take this all the way to the end. There are still a few ugly steps to get the glue back in place, but at least the separate parts are on the right track.

1. Older workaround for the python code, changing this line:
https://github.com/TimDettmers/bitsandbytes/blob/1341fb44ad03dfefe88970411c96966cd3fe43b4/bitsandbytes/cuda_setup/main.py#L123
to `self.lib = ct.cdll.LoadLibrary(str(binary_path))` to avoid the 'cannot enumerate WinPath' error.

2. And then in evaluate_cuda_setup() around line 368
https://github.com/TimDettmers/bitsandbytes/blob/1341fb44ad03dfefe88970411c96966cd3fe43b4/bitsandbytes/cuda_setup/main.py#L360
Hard coding the dll return.
`return 'bitsandbytes.dll', None, None, None, None`

3. `pip install .` , then manually placing bitsandbytes.dll in the site-packages package.

image

Then I was able to instantiate `Adam8bit` in a project and use it in a training job (successfully, gradients update etc etc)

This is with nightly Pytorch 2.0.0, CUDA 11.8, and cuDNN 8.7 so all the newest of the new, looking good.

centerionware:
Great to hear it's working! 
Because of this I may add another quick edit to this too. I learned doing this is - from what I can figure out of how the whole build system is interacting together - when installing cuda it installs stuff into visual studio about cuda, which cmake can query. so if I install 3 different versions of cuda on the same computer, I can target a specific cuda install by creating a fully seperate build/ directory, and doing first step as  ```cmake .. -T cuda=11.3``` for example instead of just ```cmake ..```

Ken1256:
Can you upload CUDA 11.7 and 11.8 precompiled for windows?

whitepapercg:
> Can you upload CUDA 11.7 and 11.8 precompiled for windows?

I agree, I tried to compile it myself, but to no avail

nicknitewolf:
I have a working version on my github but I am still unsure how to add the runtime dll dependencies of the final bitsandbytes dll other than copying them to the same folder

swumagic:
Bitsandbytes was not supported windows before, but my method can support windows.（yuhuang）
1  open folder J:\StableDiffusion\sdwebui，Click the address bar of the folder and enter CMD
or WIN+R, CMD  。enter，cd /d J:\StableDiffusion\sdwebui
2  J:\StableDiffusion\sdwebui\py310\python.exe -m pip uninstall bitsandbytes

3 J:\StableDiffusion\sdwebui\py310\python.exe -m pip uninstall bitsandbytes-windows

4 J:\StableDiffusion\sdwebui\py310\python.exe -m pip install https://github.com/jllllll/bitsandbytes-windows-webui/releases/download/wheels/bitsandbytes-0.41.1-py3-none-win_amd64.whl

Replace your SD venv directory file（python.exe Folder） here（J:\StableDiffusion\sdwebui\py310）

TimDettmers:
Sorry for taking forever for this. We want to add windows support soon. Could you please contribute to discussions at #788 and #876. Thank you!

PR #159, opened 2023-02-21
Implement proper serialization of Linear8bitLt

mryab:
As mentioned in https://github.com/huggingface/transformers/issues/20247 and as often requested by users of https://github.com/bigscience-workshop/petals, it would be very useful to save and load 8-bit quantized parameters of `Linear8bitLt` (e.g., to save bandwidth when downloading large models only for inference). This PR implements that feature by adding the missing `SCB` set of float16 quantized weights to the state dict of the layer. 

Importantly,
1. One needs to first trigger the quantization by calling layer.cuda()
2. To load the weights into the model, one also needs to call layer.cuda() before that to allocate the necessary buffers
3. If `has_fp16_weights` is True, then 8-bit quantized weights are not stored in the state — thus not serialized
4. Loading 8-bit weights into a layer with `has_fp16_weights==True` is not supported, as it would require dequantizing the weights

The test I've implemented handles all combinations of use cases and verifies that loading the serialized weights either results in an exception (if rule 2 is violated) or the same output as before serialization.

cc @borzunov @younesbelkada

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -223,7 +223,7 @@ matmul_cublas = MatMul8bit.apply
 
 @dataclass
 class MatmulLtState:
-    tile_indices: Optional[torch.Tensor] = None
+    _tile_indices: Optional[torch.Tensor] = None
     force_no_igemmlt: bool = False
     CB = None
     CxB = None
@@ -262,6 +262,15 @@ class MatmulLtState:
             "col_ampere",
         ), f"please find this assert and manually enter tile size for {self.formatB}"
         return (8, 32) if self.formatB == "col_turing" else (32, 32)
+
+    @property
+    def tile_indices(self):
+        if self._tile_indices is None:
+            device = self.CxB.device
+            transform = lambda x: F.transform(x.to(device), from_order="row", to_order=self.formatB)[0].to(x.device)
+            with torch.no_grad():
+                self._tile_indices = get_inverse_transform_indices(transform, self.get_tile_size()).to(device)
+        return self._tile_indices
 
 
 class MatMul8bitLt(torch.autograd.Function):
@@ -455,13 +464,6 @@ class MatMul8bitLt(torch.autograd.Function):
                 CB = state.CB.to(ctx.dtype_A, copy=True).mul_(state.SCB.unsqueeze(1).mul(1.0 / 127.0))
                 grad_A = torch.matmul(grad_output, CB).view(ctx.grad_shape).to(ctx.dtype_A)
             elif state.CxB is not None:
-
-                if state.tile_indices is None:
-                    order, tile_size = state.formatB, state.get_tile_size()
-                    transform = lambda x: F.transform(x.cuda(), from_order="row", to_order=order)[0].to(x.device)
-                    with torch.no_grad():
-                        state.tile_indices = get_inverse_transform_indices(transform, tile_size).to(state.CxB.device)
-
                 CB = (
                     undo_layout(state.CxB, state.tile_indices)
                     .to(ctx.dtype_A)
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -224,6 +224,34 @@ class Linear8bitLt(nn.Linear):
 
         self.weight = Int8Params(self.weight.data, has_fp16_weights=has_fp16_weights, requires_grad=has_fp16_weights)
 
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
+        super()._save_to_state_dict(destination, prefix, keep_vars)
+
+        # we only need to save SCB as extra data, because CB for quantized weights is already stored in weight.data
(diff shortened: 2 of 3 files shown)


bitsandbytes/autograd/_functions.py
```diff
@@ -455,13 +464,6 @@ def backward(ctx, grad_output):
                 CB = state.CB.to(ctx.dtype_A, copy=True).mul_(state.SCB.unsqueeze(1).mul(1.0 / 127.0))
                 grad_A = torch.matmul(grad_output, CB).view(ctx.grad_shape).to(ctx.dtype_A)
             elif state.CxB is not None:
-
-                if state.tile_indices is None:
-                    order, tile_size = state.formatB, state.get_tile_size()
-                    transform = lambda x: F.transform(x.cuda(), from_order="row", to_order=order)[0].to(x.device)
```
mryab, line 461: There was a slight possibility of a bug before: .cuda() might transfer the tensor to GPU0 instead of the GPU we're using for weights

younesbelkada:
Thanks a mile for this PR, definitely looking forward to merge this as it will help push 8bit weights on the Hub as well, as requested by the community
I will try this PR whenever I can on a dummy int8 model using `transformers`

bitsandbytes/autograd/_functions.py
```diff
@@ -263,6 +263,15 @@ def get_tile_size(self):
         ), f"please find this assert and manually enter tile size for {self.formatB}"
         return (8, 32) if self.formatB == "col_turing" else (32, 32)
 
+    @property
+    def tile_indices(self):
```
BlackSamorez, line 267: Why not use `cached_property`?

mryab: As far as I know, it is [only available](https://docs.python.org/3/library/functools.html#functools.cached_property) in Python 3.8: hence, the code won't work on systems with older Python. Still, thanks for the suggestion — I haven't actually thought of it when working on the PR!

younesbelkada (approved):
Thanks so much for this !
I tried it and it works on `transformers`, this PR would unlock the possibility to push 8bit models on the Hub for `transformers` users 
https://github.com/huggingface/transformers/pull/22177

TimDettmers (changes requested):
Thank you, Max! This is an awesome PR with a great feature. It almost looks ready to merge for me. I think there is one case that is currently missing though, and that is if a user has a GPU with compute capability less than 7.5 (GTX 1080 or older) only `.CB` is not `None` and `.CxB` is None. 

You can create a test case for this with a >= 7.5 GPU by using the `state.force_no_igemmlt` variable. See this [Linear8bitLt](https://github.com/TimDettmers/bitsandbytes/blob/main/tests/test_linear8bitlt.py#L40) test case for an example. Could you check your implementation for this case, make the necessary changes and report back? Thank you!

TimDettmers:
Thank you, Max, for this PR. If the test passes, this looks good to me. I will merge and push out a release soon.

mryab:
Hi @TimDettmers, thanks for the review! I've implemented the `force_no_igemmlt` flag as you requested and ran the tests on a 2080 machine: there are some combinations of test cases that fail, mainly if force_no_igemmlt differs before and after the serialization:

```
================================================================================================== short test summary info ===================================================================================================
FAILED tests/test_linear8bitlt.py::test_linear_serialization[False-False-False-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[False-False-False-True-False] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[False-True-False-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[False-True-False-True-False] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-False-False-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-False-False-True-False] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-False-True-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-False-True-True-False] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-True-False-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-True-False-True-False] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-True-True-False-True] - AssertionError: assert False
FAILED tests/test_linear8bitlt.py::test_linear_serialization[True-True-True-True-False] - AssertionError: assert False
=============================================================================================== 12 failed, 22 passed in 6.48s ================================================================================================
```

As such, I think that users with older GPUs might face some inconsistencies in behavior. I was thinking of adding an extra check during serialization along the lines of what you described: we can declare the feature to be a prototype and release a follow-up fix if necessary though.

PR #132, opened 2023-02-02
fix: fixes cuda_setup issue

telpirion:
The `is_cublasLt_compatible` function in cuda_setup/main.py has an uninitialized variable.

PR #163, opened 2023-02-24
fix cuda_steup undefined err

cauyxy:


justheuristic:
Probably related: https://github.com/TimDettmers/bitsandbytes/pull/132

TimDettmers:
I merged #115 instead because it was the first PR. Thanks for your contribution!

PR #220, opened 2023-03-25
Initial CMake support

niclimcy:
- For initial win32 support
- Change pythonInterface file extension to cpp because it is C++
- Requires external dependency pthread-win32 for Windows builds


rickardp:
Consider changing phthread into canonical std::thread.  There's also some scary malloc going on in that code, maybe take the opportunity to use the c++ primitives?

TimDettmers:
Thank you for this contribution! We will probably want to change to cmake and this would be an important starting point. How does this relate to #908 and #229. Thanks @rickardp for your feedback

niclimcy:
> Thank you for this contribution! We will probably want to change to cmake and this would be an important starting point. How does this relate to #908 and #229. Thanks @rickardp for your feedback

Those were seperate attempts, I think you should choose the PR that is the most complete (definitely not mine).

Issue #156, opened 2023-02-21
bitsandbytes/libbitsandbytes_cpu.so: undefined  symbol: cget_col_row_stats


XixuHu:
Hi, I came across this problem when I try to use bitsandbytes to load a big model from huggingface, and I cannot fix it. My CUDA version is 12.0 and my torch version is 1.13.1+cu116. I would like to know if there is any way to solve this problem? Thanks!

chris-aeviator:
facing the same issue when trying to resolve https://github.com/TimDettmers/bitsandbytes/issues/134 with a downgrade to 0.35

JanKalo:
Any news about this issue?

Jacobsolawetz:
Hit it as well 

```
(env) ubuntu@ip-172-31-7-128:~/uss-vision-deploy-demo$ pip list
Package                  Version    
------------------------ -----------
accelerate               0.16.0     
aiohttp                  3.8.4      
aiosignal                1.3.1      
async-timeout            4.0.2      
attrs                    22.2.0     
bitsandbytes             0.37.0     
boto3                    1.26.83    
botocore                 1.29.83    
certifi                  2022.12.7  
charset-normalizer       3.0.1      
click                    8.1.3      
datasets                 2.10.1     
deeplake                 3.2.12     
dill                     0.3.6      
entrypoints              0.4        
filelock                 3.9.0      
frozenlist               1.3.3      
fsspec                   2023.3.0   
hub                      3.0.1      
huggingface-hub          0.12.1     
humbug                   0.2.8      
idna                     3.4        
jmespath                 1.0.1      
loralib                  0.1.1      
multidict                6.0.4      
multiprocess             0.70.14    
numcodecs                0.11.0     
numpy                    1.24.2     
nvidia-cublas-cu11       11.10.3.66 
nvidia-cuda-nvrtc-cu11   11.7.99    
nvidia-cuda-runtime-cu11 11.7.99    
nvidia-cudnn-cu11        8.5.0.96   
packaging                23.0       
pandas                   1.5.3      
pathos                   0.3.0      
peft                     0.2.0.dev0 
Pillow                   9.4.0      
pip                      20.0.2     
pkg-resources            0.0.0      
pox                      0.3.2      
ppft                     1.7.6.6    
psutil                   5.9.4      
pyarrow                  11.0.0     
PyJWT                    2.6.0      
python-dateutil          2.8.2      
pytz                     2022.7.1   
PyYAML                   6.0        
[...]

PR #260, opened 2023-04-05
Fixed typo libsbitsandbytes_cpu.so

rapsealk:
Hello bitsandbytes!

While I was investigating on some problem running `bitsandbytes`, I think I found a typo.

There seems nowhere using `libsbitsandbytes_cpu.so`(lib+s+bitsandbytes_cpu.so), but `libbitsandbytes_cpu.so`(lib+bitsandbytes_cpu.so) is being used.

Thanks in advance!


xNul:
Does this fix this bug? https://github.com/TimDettmers/bitsandbytes/issues/156#issuecomment-1462329713

rapsealk:
@xNul Sorry, but I guess it won't. Actually it substitutes `libbitsandbytes.so` with `libsbitsandbytes.so` by itself.

xNul:
@rapsealk All good, thanks for the response

TimDettmers:
Thanks for the PR. I need to look further into #156 and other related bugs. Thank you for your patience.

PR #292, opened 2023-04-10
Support nvidia16 GPUs

justheuristic:
There is [a set of GPUs](https://en.wikipedia.org/wiki/GeForce_16_series) that are technically Turing, but they lack tensor cores and some other Turing features. Turns out, they are fairly popular and people host 3-5B models on those cards. This PR makes them compatible with load_in_8bit.


TimDettmers:
Excellent, thank you so much for this PR!

PR #296, opened 2023-04-11
Add support for AMD GPUs

agrocylo:
I made a fork of bitsandbytes to add support for ROCm HIP, it is currently based on 0.37.2.

It was made using [hipify_torch](https://github.com/ROCmSoftwarePlatform/hipify_torch) as a base and modifying the generated files.

It's probably not mergeable as is, but could be used to discuss how best to implement it, as it would be beneficial for users to have AMD GPUs supported officially.

The problem is that I'm not sure what the best way to keep in sync change in cuda files, as adding a burden to duplicate code in both cuda and hip files is undesirable. 

There is also another fork with the same purpose: https://github.com/broncotc/bitsandbytes-rocm, but it doesn't seem maintained, I added the author as a co-author on some commits where I was inspired by his changes.


deftdawg:
There is also @0cc4m 's [rocm branch](https://github.com/0cc4m/bitsandbytes-rocm/tree/rocm)  which was last touched a bit more recently...

PR #303, opened 2023-04-12
Change printing to logging

JulesGM:
Makes it so we can filter the cuda setup messages at least


JulesGM:
Is this not happening? I see the print statement is still there
https://github.com/TimDettmers/bitsandbytes/blob/main/bitsandbytes/cuda_setup/main.py#L147

JulesGM:
It's literally a one-second change to make

JulesGM:
@TimDettmers

Issue #324, opened 2023-04-17
New OOM bug introduced in bitsandbytes 0.38.x?


KukumavMozolo:
Hi there, apparently 0.38.1 or 0.38.0 introduced a bug that increases memory consumption by a lot when trying to save a model.
For details see this post in the [alpaca-lora githubg](https://github.com/tloen/alpaca-lora/issues/350).

The bug doesn't happen when using bitsandbytes==0.37.2

Note, while this script says otherwise i am pretty sure i have cuda 11.7 installed

` python -m bitsandbytes

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please submit your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
CUDA SETUP: CUDA version lower than 11 are currently not supported for LLM.int8(). You will be only to use 8-bit optimizers and quantization routines!!
CUDA SETUP: CUDA runtime path found: ...lib/libcudart.so
CUDA SETUP: Highest compute capability among GPUs detected: 8.6
CUDA SETUP: Detected CUDA version 102
CUDA SETUP: Required library version not found: libbitsandbytes_cuda102.so. Maybe you need to compile it from source?
CUDA SETUP: Defaulting to libbitsandbytes_cpu.so...

================================================ERROR=====================================
CUDA SETUP: CUDA detection failed! Possible reasons:
1. CUDA driver not installed
2. CUDA not installed
3. You have multiple conflicting CUDA libraries
4. Required library not pre-compiled for this bitsandbytes release!
CUDA SETUP: If you compiled from source, try again with `make CUDA_VERSION=DETECTED_CUDA_VERSION` for example, `make CUDA_VERSION=113`.
CUDA SETUP: The CUDA version for the compile might depend on your conda install. Inspect CUDA version via `conda list | grep cuda`.
================================================================================

CUDA SETUP: Something unexpected happened. Please compile from source:
git clone git@github.com:TimDettmers/bitsandbytes.git
cd bitsandbytes
CUDA_VERSION=102
python setup.py install
CUDA SETUP: CUDA version lower than 11 are currently not supported for LLM.int8(). You will be only to use 8-bit optimizers and quantization routines!!
CUDA SETUP: CUDA runtime path found: .../lib/libcudart.so
CUDA SETUP: Highest compute capability among GPUs detected: 8.6
CUDA SETUP: Detected CUDA version 102
[...]

PR #330, opened 2023-04-20
Make Linear8bitLt serialization more memory-efficient

mryab:
Previously, the Linear8bitLt serialization code relied on cloning the weights if they were transposed to the Ampere/Turing format at the forward pass. This leads to a 3x increase in peak memory footprint for the weight matrix, which might be quite noticeable for large models. The PR changes that to undoing the weights and then transforming them back to the original format.

The following script now consumes 9 vs 15 MB of peak memory, which might resolve https://github.com/TimDettmers/bitsandbytes/issues/324.
```
import torch

from bitsandbytes.nn.modules import Linear8bitLt, Int8Params


with torch.profiler.profile(schedule=torch.profiler.schedule(wait=1, warmup=1, active=10, repeat=2),
                            on_trace_ready=torch.profiler.tensorboard_trace_handler('./log/chkp9'),
                            profile_memory=True, with_stack=True) as prof:
    x = torch.randn(3, 1024, dtype=torch.half, device="cuda")
    linear_custom = Linear8bitLt(
        1024,
        3072,
        has_fp16_weights=False,
        threshold=6.0,
    )
    linear_custom.cuda()
    out=linear_custom(x)

    for _ in range(50):
        linear_custom.state_dict()
        prof.step()
print(torch.cuda.max_memory_reserved())
```


otto-dev:
Thank you for looking into this, I ran into the same issue. I built your branch locally (cherry-picked to master) with `make cuda12x CUDA_VERSION=121`, and this patch did not fix the bug reported here in the following comment: https://github.com/tloen/alpaca-lora/issues/344#issuecomment-1510438597. I'm getting the same error with exactly `68.00 MiB` just like reported by the other user. Just thought I'd let you know. Appreciate your efforts!

Issue #336, opened 2023-04-23
WSL CUDA install on Windows 10: GPU not detected


ksylvan:
# nvidia-smi output

```text
Sun Apr 23 01:54:14 2023
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 530.30.04              Driver Version: 531.29       CUDA Version: 12.1     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                  Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf            Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA GeForce RTX 4090         On | 00000000:01:00.0  On |                  Off |
|  0%   41C    P8               21W / 450W|    971MiB / 24564MiB |      1%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|    0   N/A  N/A        27      G   /Xwayland                                 N/A      |
+---------------------------------------------------------------------------------------+
```

# Output of bitsandbytes startup

Followed the instructions in <https://github.com/oobabooga/text-generation-webui> `README.md` for the WSL
Conda install.

```
$ python -m bitsandbytes

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please run

python -m bitsandbytes

 and submit this information together with your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
bin /home/kayvan/.miniconda3/envs/textgen/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so
/home/kayvan/.miniconda3/envs/textgen/lib/python3.10/site-packages/bitsandbytes/cextension.py:33: UserWarning: The installed version of bitsandbytes was compiled without GPU support. 8-bit optimizers, 8-bit multiplication, and GPU quantization are unavailable.
  warn("The installed version of bitsandbytes was compiled without GPU support. "
CUDA SETUP: WARNING! libcuda.so not found! Do you have a CUDA driver installed? If you are on a cluster, make sure you are on a CUDA machine!
[...]

PR #337, opened 2023-04-24
Solution to GPU not detected in Windows WSL

ksylvan:
This is the solution to https://github.com/TimDettmers/bitsandbytes/issues/336 and a few other issues about GPU not being
seen in Windows WSL.

TL;DR: The CUDA driver installed on Windows host is stubbed inside the WSL 2 as libcuda.so and we need to include
*that* directory location in `LD_LIBRARY_PATH`


nivibilla:
This is actually written in the Nvidia documentation, however is missing in the wsl installation page.

https://docs.nvidia.com/cuda/cuda-installation-guide-linux/#mandatory-actions

ksylvan:
Yes, @nivibilla - It's in the Linux sections - but not the windows WSL sections.

FYI, the WSL Ubuntu environment uses `/etc/ld.so.cache.d/` files for the majority of those librarues, so setting them in LD_LIBRARY_PATH is not strictly needed, but nothing mentions the need to include  `/usr/lib/wsl/lib` directory.

nivibilla:
You're right, it should be mentioned in the docs for clarity. I didn't have issues with installing cuda this way and it worked for me. Just wanted to let you know.

ksylvan:
Can we get this merged? It's simply an addition to `errors_and_solutions.md`. @TimDettmers

TimDettmers:
Thank you, this is a great addition to the docs! Sorry for taking so long to respond here. Could you add this commit to the latest branch on main. I made a mess with a force push when pushing private code for the 4-bit beta.

PR #385, opened 2023-05-11
When using sudo to run the command, the system variable environment is reset and the original environment variable cannot be obtained

dailingx:
I get an error while running the program on the linux, 
```
===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please run

python -m bitsandbytes

 and submit this information together with your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
bin /home/stable-diffusion/kohya_ss/venv/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cpu.so
CUDA_SETUP: WARNING! libcudart.so not found in any environmental path. Searching in backup paths...
```
This CUDA_SETUP WARNING looks like the CUDA information was not found. Following the error log, I found this file: https://github.com/TimDettmers/bitsandbytes/blob/9e7cdc9ea95e9756d9f5621a0e2c7e2538363fae/bitsandbytes/cuda_setup/main.py. The determine_cuda_runtime_lib_path method in this file will use the LD_LIBRARY_PATH environment variable to get the CUDA installation path. But I use sudo to run the program, which causes non-basic environment variables to be reset, including LD_LIBRARY_PATH and CONDA_PREFIX, os.environ can only get the environment variables in the current shell, so the original environment variables of the system cannot be obtained.
I try to use sudo -i env to obtain the original environment variables of the system, and fill in the values ​​of CONDA_PREFIX and LD_LIBRARY_PATH that do not exist in env_vars, which can effectively solve my problem.
I think this is a useful optimization point to avoid some potential anomalies brought by the sudo command.

### DIFF
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -54,9 +54,13 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
 
 def restore_original_system_env_vars(env_vars: Dict[str, str]):
     """
+        If the python program was started with `sudo`,
         Use the `sudo -i env` command to obtain the original environment variables of the system, including CONDA_PREFIX, LD_LIBRARY_PATH...
         Then replace the environment variables obtained from the current shell, to obtain correct and complete environment variables
     """
+    if not os.geteuid() == 0:
+        return
+
     output = subprocess.check_output(['sudo', '-i', 'env'])
     output_str = output.decode('utf-8')
     env_lines = output_str.split('\n')
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -64,10 +64,7 @@ def restore_original_system_env_vars(env_vars: Dict[str, str]):
     if os.geteuid() != 0:
         return
 
-    output = subprocess.check_output(['sudo', '-i', 'env'])
-    output_str = output.decode('utf-8')
-    env_lines = output_str.split('\n')
-    for env_line in env_lines:
+    for env_line in subprocess.check_output(['sudo', '-i', 'env'], encoding="utf-8").splitlines():
         if '=' in env_line:
             key, value = env_line.split('=', 1)
             if (key == "CONDA_PREFIX" or key == "LD_LIBRARY_PATH") and key not in env_vars:
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -1,6 +1,7 @@
 import os
 import subprocess
 from typing import Dict
+from distutils.spawn import find_executable
 
 
 def to_be_ignored(env_var: str, value: str) -> bool:
@@ -54,10 +55,12 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
 
 def restore_original_system_env_vars(env_vars: Dict[str, str]):
     """
-        If the python program was started with `sudo`,
+        If the system supports `sudo` command, and the python program was started with `sudo`,
         Use the `sudo -i env` command to obtain the original environment variables of the system, including CONDA_PREFIX, LD_LIBRARY_PATH...
         Then replace the environment variables obtained from the current shell, to obtain correct and complete environment variables
     """
+    if find_executable('sudo') is None:
+        return
     if not os.geteuid() == 0:
         return
 
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -53,6 +53,10 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
     }
 
 def restore_original_system_env_vars(env_vars: Dict[str, str]):
+    """
+        Use the `sudo -i env` command to obtain the original environment variables of the system, including CONDA_PREFIX, LD_LIBRARY_PATH...
+        Then replace the environment variables obtained from the current shell, to obtain correct and complete environment variables
+    """
     output = subprocess.check_output(['sudo', '-i', 'env'])
     output_str = output.decode('utf-8')
     env_lines = output_str.split('\n')
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -1,6 +1,7 @@
 import os
 import subprocess
 from typing import Dict
+import shutil
 
 
 def to_be_ignored(env_var: str, value: str) -> bool:
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -63,5 +63,5 @@ def restore_original_system_env_vars(env_vars: Dict[str, str]):
     for env_line in env_lines:
         if '=' in env_line:
             key, value = env_line.split('=', 1)
-            if key in env_vars:
+            if (key == "CONDA_PREFIX" or key == "LD_LIBRARY_PATH") and key not in env_vars:
                 env_vars[key] = value
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -61,7 +61,7 @@ def restore_original_system_env_vars(env_vars: Dict[str, str]):
     """
     if find_executable('sudo') is None:
         return
-    if not os.geteuid() == 0:
+    if os.geteuid() != 0:
         return
 
     output = subprocess.check_output(['sudo', '-i', 'env'])
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -58,7 +58,7 @@ def restore_original_system_env_vars(env_vars: Dict[str, str]):
         Use the `sudo -i env` command to obtain the original environment variables of the system, including CONDA_PREFIX, LD_LIBRARY_PATH...
         Then replace the environment variables obtained from the current shell, to obtain correct and complete environment variables
     """
-    if find_executable('sudo') is None:
+    if shutil.which('sudo') is None:
         return
     if os.geteuid() != 0:
         return
--- a/bitsandbytes/cuda_setup/env_vars.py
+++ b/bitsandbytes/cuda_setup/env_vars.py
@@ -56,7 +56,7 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
 def restore_original_system_env_vars(env_vars: Dict[str, str]):
     """
         If the system supports `sudo` command, and the python program was started with `sudo`,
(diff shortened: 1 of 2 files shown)


dailingx:
I think this situation needs to be taken into account, what do you think?

bitsandbytes/cuda_setup/env_vars.py
```diff
@@ -50,3 +51,17 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
         for env_var, value in os.environ.items()
         if is_relevant_candidate_env_var(env_var, value)
     }
+
+def restore_original_system_env_vars(env_vars: Dict[str, str]):
+    """
+        Use the `sudo -i env` command to obtain the original environment variables of the system, including CONDA_PREFIX, LD_LIBRARY_PATH...
+        Then replace the environment variables obtained from the current shell, to obtain correct and complete environment variables
+    """
+    output = subprocess.check_output(['sudo', '-i', 'env'])
```
TimDettmers, line 60: This should only be executed if we are running on a linux system (bitsandbytes will soon support windows).

TimDettmers: I thought that `sudo -i env` does not require a password, but it does. I think with that we cannot have this in the main code-path that is executed by everyone. 

Is there a way to detect if the python program was started with sudo? I think if we could detect this and then apply this fix conditionally dependent on that this would be a good solution.

Otherwise, can you think of different solutions to this issue that does not require calling sudo?

TimDettmers (changes requested):
Currently, the main problem is that this solution needs a sudo password to run. Not all users use sudo, as such, it should only be executed when a user used sudo. Can we detect this or maybe there is another solution that does not require a password?

bitsandbytes/cuda_setup/env_vars.py
dailingx: I use `os.geteuid()` to ensure that the python program is started with `sudo`. If the program was started with `sudo`, this value will be 0 (the user ID of the superuser).

WaferJay:
`LD_LIBRARY_PATH` and `CONDA_PREFIX` environment variables can be set by hand code to resolve them in `sudo`.

This may help:

```sh
sudo LD_LIBRARY_PATH="$LD_LIBRARY_PATH" CONDA_PREFIX="$CONDA_PREFIX" python -m bitsandbytes
```

bitsandbytes/cuda_setup/env_vars.py
dailingx: And I use `find_executable('sudo')` to ensure the system supports sudo command. When running on a Windows system, this code will not be executed.

dailingx:
> Currently, the main problem is that this solution needs a sudo password to run. Not all users use sudo, as such, it should only be executed when a user used sudo. Can we detect this or maybe there is another solution that does not require a password?



> Currently, the main problem is that this solution needs a sudo password to run. Not all users use sudo, as such, it should only be executed when a user used sudo. Can we detect this or maybe there is another solution that does not require a password?



> Currently, the main problem is that this solution needs a sudo password to run. Not all users use sudo, as such, it should only be executed when a user used sudo. Can we detect this or maybe there is another solution that does not require a password?

How about this solution code? Is there anything else I need to do?

dailingx:
> Currently, the main problem is that this solution needs a sudo password to run. Not all users use sudo, as such, it should only be executed when a user used sudo. Can we detect this or maybe there is another solution that does not require a password?

How about my solution code? Is there anything else I need to do?

dailingx (commented):
Now, this code just be executed when a  user system support sudo and user used sudo

bitsandbytes/cuda_setup/env_vars.py
```diff
@@ -50,3 +52,23 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
...
+    """
+    if find_executable('sudo') is None:
+        return
+    if not os.geteuid() == 0:
+        return
+
+    output = subprocess.check_output(['sudo', '-i', 'env'])
+    output_str = output.decode('utf-8')
+    env_lines = output_str.split('\n')
+    for env_line in env_lines:
```
akx, line 70: More succinctly:
```suggestion
    for env_line in subprocess.check_output(['sudo', '-i', 'env'], encoding="utf-8").splitlines():
```

akx (commented):
Beyond the inline comments, I'm not sure this PR does what you expect it to.

As I understand it, `sudo -i env` will run `env` as root, with a login session (`-i` being `--login`), so this will get you values as if you'd logged in as `root` directly; it will _not_ get you the environment variables of the session before you had `sudo`ed, so e.g. `CONDA_PREFIX` would not be one of the activated environment.

The correct fix is to configure your system's `sudo` to keep environment variables correctly: https://stackoverflow.com/a/8636711

dailingx:
Time has passed too long, I might use the sudo command to run because some underlying dependencies may require this permission. Nonetheless, I think we should consider this case.

dailingx:
Now, I have re-optimized the code, switching from distutils to shutil. Is there anything else I need to do?

akx:
@dailingx Well, again, I wonder if this patch works at all; see below:

```bash
# pretend we activate a Conda environment
regular-user@b8013804a7de:~$ export CONDA_PREFIX=foobar

# check that a regular Python sees that environment variable
regular-user@b8013804a7de:~$ python3
>>> import os
>>> os.environ.get("CONDA_PREFIX")
'foobar'

# now run Python using sudo
regular-user@b8013804a7de:~$ sudo python3
[sudo] password for regular-user: ...
>>> import os

# as expected, the environment variable isn't set because of keep-env
>>> os.environ.get("CONDA_PREFIX")

# now run the equivalent of what this PR does to find whether `CONDA_PREFIX` is set:
>>> import subprocess
>>> "CONDA_PREFIX" in subprocess.check_output(["sudo", "-i", "env"], encoding="UTF-8")
False
```

dailingx:
image

dailingx:
This image depicts the complete process of my execution, indicating that this code is worked on my end. Have there been any additional steps or operations on your side? This doesn't seem quite normal.

dailingx:
Oh! I see. I think this might be because when you use sudo, it requires a password, but when using sudo -i env in the program, you're unable to input the password. I will think about how to  handle this case.

dailingx:
Oh, perhaps we can bypass this case. The code is valid only when executed as the root user, so I utilized os.geteuid() upon method invocation to verify root user status. If not executed as the root user, my code will not to run.

dailingx:
image

@akx Just like the description of this method. what do you think?

dailingx:
@akx This is because you are not the root user, so I use os.geteuid() at the entry point of my code to filter out non-root users. You can switch to root user and try again, it will work at all.

akx:
>  You can switch to root user and try again, it will work at all.

In the transcript above, I did switch to root with `sudo`. But if you mean I should start off as `root` and have set that envvar as `root`:

```
root@b8013804a7de:/# id
uid=0(root) gid=0(root) groups=0(root)
root@b8013804a7de:/# export CONDA_PREFIX=foobar
root@b8013804a7de:/# python -c 'import os;print(os.environ.get("CONDA_PREFIX"))'
foobar
root@b8013804a7de:/# sudo python -c 'import os;print(os.environ.get("CONDA_PREFIX"))'
None
root@b8013804a7de:/# sudo python -c 'import subprocess;print("CONDA_PREFIX" in subprocess.check_output(["sudo", "-i", "env"], encoding="UTF-8"))'
False
root@b8013804a7de:/# python -c 'import subprocess;print("CONDA_PREFIX" in subprocess.check_output(["sudo", "-i", "env"], encoding="UTF-8"))'
False
root@b8013804a7de:/#
```

IOW, I really do not think this patch works in the general case.

bitsandbytes/cuda_setup/env_vars.py
```diff
@@ -56,16 +56,19 @@ def get_potentially_lib_path_containing_env_vars() -> Dict[str, str]:
...
     if os.geteuid() != 0:
         return
 
-    for env_line in subprocess.check_output(['sudo', '-i', 'env'], encoding="utf-8").splitlines():
-        if '=' in env_line:
-            key, value = env_line.split('=', 1)
-            if (key == "CONDA_PREFIX" or key == "LD_LIBRARY_PATH") and key not in env_vars:
-                env_vars[key] = value
+    sudo_env_output = subprocess.check_output(['sudo', '-i', 'env'], encoding="utf-8")
+    current_env_output = subprocess.check_output(['env'], encoding="utf-8")
```
akx, line 68: This is the same as just reading `os.environ`.

dailingx: image
if run with sudo, this is not the same as just reading os.environ.

akx: You're not testing for the right thing. 😁 

The environment variable `SUDO_COMMAND` includes the string `test_1` in your case, so that experiment will always return true:

```
# sudo python -c 'import subprocess;print("hello hello!");print(subprocess.check_output(["env"]))'
hello hello!
b'HOSTNAME=315f5cc030af\nLANG=C.UTF-8\nTERM=xterm\nPATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin\nMAIL=/var/mail/root\nLOGNAME=root\nUSER=root\nHOME=/root\nSHELL=/bin/bash\nSUDO_COMMAND=/usr/local/bin/python -c import subprocess;print("hello hello!");print(subprocess.check_output(["env"]))\nSUDO_USER=root\nSUDO_UID=0\nSUDO_GID=0\n'
```
If I were to test whether `hello hello` appears in the output of `env`, it'd be true.

akx:
@dailingx Please see my reply in the review. You're not testing for the right thing there; see for yourself with
```
sudo python -c 'import subprocess;print(subprocess.check_output(["env"], encoding="UTF-8"))'
```
and see whether you can find `test1=test` there. 😄

dailingx:
@akx You are right. But this is unbelievable. Why does using `sudo python -c 'import subprocess;print("test_1" in subprocess.check_output(["env"], encoding="UTF-8"))'` return True?

dailingx:
@akx You are right! I understand, thank you very much for your answer, I learned a lot.

dailingx:
@akx The originally submitted code was useful to me. This may be due to a problem with the CONDA_PREFIX environment variable when it was written. Although I don't know what happened, after activating conda, I did not find this variable in `env`, and It was found in ` sudo -i env`. like this:
image
image

dailingx:
If you know why, I'd love to hear it and thank you very much.

akx:
@dailingx I can only imagine that's related to your machine's `sudo` or root environment configuration. For instance, if your `root` user has `.bashrc` or other profile things that activate a Conda environment, that would happen with `sudo -i`.

dailingx:
@akx Thank you for bringing the machine environment issue to my attention. I appreciate your advice and your kind demeanor. I will investigate this further. In the meantime, I will close this PR.

PR #469, opened 2023-06-01
Add `device` parameter to `Linear` subclasses and `Embedding`

shadeMe:
Useful when one wants to initialize the layers on the `meta` device without using the `init_on_empty` helper in the `transformers` package.


vince62s:
+1
it will make  nn.utils.skip_init() work

vince62s:
@shadeMe  I opened this https://discuss.pytorch.org/t/what-does-nn-utils-skip-init-really-do/192370
because I am still having an issue with nn.utils.skip_init() applied on Linear4bit
did you manage to make it work ?

PR #488, opened 2023-06-06
Added install requirements to setup

ccoulombe:
When installing or distributing, ensure needed requirements are present.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ setup(
     url="https://github.com/TimDettmers/bitsandbytes",
     packages=find_packages(),
     package_data={"": libs},
+    install_requires=['torch', 'numpy', 'scipy', 'pandas', 'matplotlib'],
     long_description=read("README.md"),
     long_description_content_type="text/markdown",
     classifiers=[
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ setup(
     url="https://github.com/TimDettmers/bitsandbytes",
     packages=find_packages(),
     package_data={"": libs},
-    install_requires=['torch', 'numpy', 'scipy', 'pandas', 'matplotlib'],
+    install_requires=['torch', 'numpy', 'scipy'],
+    extras_require={'benchmark': ['pandas', 'matplotlib']},
     long_description=read("README.md"),
     long_description_content_type="text/markdown",
     classifiers=[


TimDettmers:
Thanks for this contribution! When working with `setup.py` in other projects, I sometimes noticed that my pytorch gets uninstalled and a different version installed. I think this only happens if a specific version is pinned. Is there any other scenario where this addition could cause the wrong pytorch version to be installed? I could imagine a new OS/VM and one installed bitsandbytes first and it installs a different pytorch version as one intended.

@Titus-von-Koeller @younesbelkada, what is your opinion?

akx:
@TimDettmers If I may chip in, that shouldn't be an issue here – there is no version requirement here, so any version of `torch` installed should satisfy the requirement. (Of course `setup.py` in itself is antiquated, and should be replaced with a PEP 517 build backend, but I see there's parallel work to do `cmake` things and so on, so that'd probably conflict.)

setup.py
```diff
@@ -27,6 +27,7 @@ def read(fname):
     url="https://github.com/TimDettmers/bitsandbytes",
     packages=find_packages(),
     package_data={"": libs},
+    install_requires=['torch', 'numpy', 'scipy', 'pandas', 'matplotlib'],
```
akx, line 30: Are you absolutely sure `pandas` and `matplotlib` are _hard_ required for all use of the library? If not, they should be declared in extras, not here.

akx, line 30: `pandas` and `matplotlib` are only required for the benchmarking scripts; they should be required only as extras:
```suggestion
    install_requires=['torch', 'numpy', 'scipy'],
    extras_require={'benchmark': ['pandas', 'matplotlib']},
```

akx (changes requested):
Not good to hard-require pandas and matplotlib.

ccoulombe:
@TimDettmers Pip may install a different pytorch version if another package installed in the same virtual environement as bitsandbyte require (strictly) a different version. Otherwise, since the requirement is `torch`, it means that no specific version is required by bitsandbytes.

On HPC systems where administrators maintains a wheelhouse, or to simply avoid user issues when installing/using packages, it is easier to have the main package request its dependencies.

As for `matplotlib` or `pandas`, I've accepted @akx suggestion

Titus-von-Koeller:
You're right, the build part of bnb might see overhauls soon, but I don't see any reason not to merge this. Makes sense. Thanks for your contribution, really appreciated 🤗

akx:
This was mis-merged somehow 😞 (see #981, #983).

PR #494, opened 2023-06-07
Update README.md

pranavgitb11:
Changed misleading Hardware requirements from "(a GPU from 2018 or older)" to "(a GPU from 2018 or newer)"


TimDettmers:
Good catch!

PR #506, opened 2023-06-10
Add support for CUDA 12x

PriNova:
This is a PR to add CUDA 12x support.

With this change, I managed to install bitsandbytes on WSL2 (Ubuntu 22.04.1)

### DIFF
--- a/bitsandbytes/cuda_setup/main.py
+++ b/bitsandbytes/cuda_setup/main.py
@@ -31,7 +31,7 @@ from .env_vars import get_potentially_lib_path_containing_env_vars
 # libcudart.so is missing by default for a conda install with PyTorch 2.0 and instead
 # we have libcudart.so.11.0 which causes a lot of errors before
 # not sure if libcudart.so.12.0 exists in pytorch installs, but it does not hurt
-CUDA_RUNTIME_LIBS: list = ["libcudart.so", 'libcudart.so.11.0', 'libcudart.so.12.0']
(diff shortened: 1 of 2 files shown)


bitsandbytes/cuda_setup/main.py
```diff
@@ -75,6 +75,8 @@ def generate_instructions(self):
             make_cmd += ' make cuda110'
         elif self.cuda_version_string[:2] == '11' and int(self.cuda_version_string[2]) > 0:
             make_cmd += ' make cuda11x'
+        elif self.cuda_version_string == '121':
```
rapsealk, line 78:
```suggestion
        elif self.cuda_version_string[:2] == '12' and 1 >= int(self.cuda_version_string[2]) >= 0:
```
I suggest this change to support `12.0` either.

PriNova:
Thank you @rapsealk for your suggestion

bitsandbytes/cuda_setup/main.py
TimDettmers: I think this is the only line necessary for the support of the automatic debugging string. The other parts should work for cuda 12.x.

PriNova: Yes, as long as nothing will be changed from PyTorch or Cuda side it should work for future releases, too.
(I wonder, if this was a question from your point of view?)

TimDettmers:
Thank you for your contribution! I have been a bit slow on PRs and I had a look before at #703 and reviewed that one. @PriNova can you have a look at #703 and add to the discussion there? I am particularly curious about the requirement to add the paths for `libcudart.so.12.1`. Thank you!

PR #525, opened 2023-06-20
Added scipy to requirements.txt

dulalbert:
Added scipy to requirements.txt as it is used but not added to requirements

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 lion-pytorch
 pytest
+scipy
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
     license="MIT",
     keywords="gpu optimizers optimization 8-bit quantization compression",
     url="https://github.com/TimDettmers/bitsandbytes",
+    install_requires=['scipy'],
     packages=find_packages(),
     package_data={"": libs},
     long_description=read("README.md"),


shadeMe:
+1

This breaks the conditional importing of `bitsandbytes`, i.e.,
```python
try:
    import bitsandbytes

    has_bitsandbytes = True
except ImportError:
    bitsandbytes = None  # type: ignore
    has_bitsandbytes = False
```
`has_bitsandbytes` will be `False` even if the user has the library installed (but not `scipy`) since the `ImportError` gets masked.

danieldk:
I think that the dependency should also be added to the `install_requires`, otherwise `scipy` doesn't get installed when `pip install`-ing `bitsandbytes`.

nazarov-yuriy:
@TimDettmers 
Could we merge it?
Apparently https://github.com/TimDettmers/bitsandbytes/commit/8a20cd864b3c890a51e27cb2f460382b75734b4a is not enough.

rasbt:
Should be a quick merge in case this got forgotten 😅

mnoukhov:
@TimDettmers is this ok to merge?

https://github.com/TimDettmers/bitsandbytes/commit/8a20cd864b3c890a51e27cb2f460382b75734b4a is indeed not enough. Anyone installing bitsandbytes through pip e.g. `pip install bitsandbytes` will not install `scipy` and code will error.
`scipy` needs to be added to the `install_requires`

mnoukhov:
This still needs to be addressed. Easy fix, should be merged

TimDettmers:
Sorry for the delay on this. Thanks for contributing this fix!

PR #527, opened 2023-06-20
add sort to get_compute_capabilities

Ronnoc:
In `get_compute_capability`, the last item of ccs (`ccs[-1]`) will be selected as the maximum compute capability.
But `ccs` are not sorted in `get_compute_capabilities`
This PR adds sorting to `get_compute_capabilities`

### DIFF
--- a/bitsandbytes/cuda_setup/main.py
+++ b/bitsandbytes/cuda_setup/main.py
@@ -355,6 +355,7 @@ def get_compute_capabilities(cuda):
         # 2. call extern C function to determine CC
         check_cuda_result(cuda, cuda.cuDeviceComputeCapability(ref_major, ref_minor, device))
         ccs.append(f"{cc_major.value}.{cc_minor.value}")
+    ccs.sort(key=lambda v: tuple(map(int, str(v).split("."))))
 
     return ccs
 


Ronnoc:
@TimDettmers Please help review this Pull Request. Thank you.

bitsandbytes/cuda_setup/main.py
```diff
@@ -348,13 +348,17 @@ def get_compute_capabilities(cuda):
 
     check_cuda_result(cuda, cuda.cuDeviceGetCount(ct.byref(nGpus)))
     ccs = []
+    ccs_int_tuple = []
```
Dessix, line 351:
```suggestion
```

Dessix, line 351:
```suggestion
```

bitsandbytes/cuda_setup/main.py
```diff
@@ -348,13 +348,17 @@ def get_compute_capabilities(cuda):
...
         check_cuda_result(cuda, cuda.cuDeviceGet(ct.byref(device), i))
         ref_major = ct.byref(cc_major)
         ref_minor = ct.byref(cc_minor)
         # 2. call extern C function to determine CC
         check_cuda_result(cuda, cuda.cuDeviceComputeCapability(ref_major, ref_minor, device))
-        ccs.append(f"{cc_major.value}.{cc_minor.value}")
+        ccs_int_tuple.append((int(cc_major.value), int(cc_minor.value)))
+    ccs_int_tuple.sort()
+    for (major, minor) in ccs_int_tuple:
+        ccs.append(f"{major}.{minor}")
```
Dessix, line 361:
```suggestion
        ccs.append(f"{cc_major.value}.{cc_minor.value}")
    ccs.sort(key=lambda v: tuple(map(int, str(v).split("."))))
```

Dessix (changes requested):
Try sorting using sort-by functionality.

Dessix (approved):

TimDettmers:
Thank you for catching this! I think a question is if we should select the highest or lowest compute capability. The lowest would give the most broad support for multiple GPU. However, it is common that one uses an old GPU for monitors and a new GPU for deep learning work. As such, one would want to select the highest compute capability. @Titus-von-Koeller @younesbelkada any thoughts?

TimDettmers:
I will merge #703 instead as it is more comprehensive, but lets chat a bit about which how we should sort the compute capability

Ronnoc:
closed this PR as it's included in another one

Titus-von-Koeller:
> I think a question is if we should select the highest or lowest compute capability. The lowest would give the most broad support for multiple GPU. However, it is common that one uses an old GPU for monitors and a new GPU for deep learning work. As such, one would want to select the highest compute capability

I agree, highest is a good choice. We'll see if this is contrary to people's needs or expectations. If that is the case, we can add some logic to make it configurable.

Thanks a lot @Ronnoc for your contribution!

Issue #528, opened 2023-06-20
RuntimeError: Something when wrong when trying to find file. Maybe you do not have a linux system?


mmderakhshani:
Hi, I have installed bitsandbytes using following commands:

```
pip install -q bitsandbytes datasets accelerate loralib
pip install -q git+https://github.com/huggingface/transformers.git@main git+https://github.com/huggingface/peft.git
```

But when I run `python -m bitsandbytes`, I get the following error. Do you have any idea how to solve it? For more information, please see the following:

Python version
```
Python 3.8.5 (default, Sep  4 2020, 07:30:14) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
```

Pytorch/cuda version:
```
>>> torch.__version__
'2.0.1+cu117'
>>> torch.version.cuda
'11.7'
```

===================================BUG REPORT===================================
Welcome to bitsandbytes. For bug reports, please run

python -m bitsandbytes

 and submit this information together with your error trace to: https://github.com/TimDettmers/bitsandbytes/issues
================================================================================
bin /anaconda/envs/mlenv/lib/python3.8/site-packages/bitsandbytes/libbitsandbytes_cuda116_nocublaslt.so
/anaconda/envs/mlenv/lib/python3.8/site-packages/bitsandbytes/cuda_setup/main.py:149: UserWarning: Found duplicate ['libcudart.so', 'libcudart.so.11.0', 'libcudart.so.12.0'] files: {PosixPath('/anaconda/envs/mlenv/lib/libcudart.so.11.0'), PosixPath('/anaconda/envs/mlenv/lib/libcudart.so')}.. We'll flip a coin and try one of these, in order to fail forward.
Either way, this might cause trouble in the future:
If you get `CUDA error: invalid device function` errors, the above might be the cause and the solution is to make sure only one ['libcudart.so', 'libcudart.so.11.0', 'libcudart.so.12.0'] in the paths that we search based on your env.
  warn(msg)
CUDA SETUP: CUDA runtime path found: /anaconda/envs/mlenv/lib/libcudart.so.11.0
CUDA SETUP: Highest compute capability among GPUs detected: 3.7
CUDA SETUP: Detected CUDA version 116
[...]

PR #531, opened 2023-06-21
fix install bug on __main.py__ #528

hzg0601:
In a multi-user linux system, some users are not allowed to access other user's directory, however, the function `execute_and_return` try to do so, which would cause Permission denied error. 
So the function `find_file_recursive` raise `RuntimeError('Something when wrong when trying to find file. Maybe you do not have a linux system?')`  whenever `len(err)>0` is not logic. In this case, the  `Permission denied` should be excluded.
Presumably, this commit may partly solve the issue #528

### DIFF
--- a/bitsandbytes/__main__.py
+++ b/bitsandbytes/__main__.py
@@ -33,8 +33,14 @@ def execute_and_return(command_string: str) -> Tuple[str, str]:
 def find_file_recursive(folder, filename):
     cmd = f'find {folder} -name {filename}'
     out, err = execute_and_return(cmd)
+
     if len(err) > 0:
-        raise RuntimeError('Something when wrong when trying to find file. Maybe you do not have a linux system?')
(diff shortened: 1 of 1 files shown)


ToddMorrill:
This is an issue I'm facing in Azure ML Studio. Manually running the failed call to `find_file_recursive('/usr/local', '*cuda*so')` I have:
```
$ find /usr/local -name *cuda*so
find: ‘/usr/local/qualys’: Permission denied
/usr/local/cuda-11.3/targets/x86_64-linux/lib/stubs/libcuda.so
/usr/local/cuda-11.3/targets/x86_64-linux/lib/libcudart.so
```
I can also confirm that @hzg0601's patch works.

TimDettmers:
Sorry for taking so long on this. This has been fixed in a duplicate PR that I already merged a couple of minutes ago.

PR #563, opened 2023-07-04
Fixed missing `Embedding` export

neel04:
A very small fix - `__init__.py` has a missing `Embedding` import; So one can't access that module under `bitsandbytes.nn`


TimDettmers:
Good catch! Sorry for taking so long.

PR #567, opened 2023-07-06
[BugFix] replace view+continuous with reshape 

### LINKED ISSUE #550: view size is not compatible with input tensor's size and stride
bitsandbytes/autograd /_functions.py

300         # 1. Quantize A
301         if len(A.shape) == 3:
302             A = A.view(-1, A.shape[-1]).contiguous()

Is this a bug? Change it to :A = A.contiguous().view(-1, A.shape[-1])

 an error before changing:

view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...)  instead

normal after  change

apbard:
fixes #550

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -322,7 +322,7 @@ class MatMul8bitLt(torch.autograd.Function):
 
         # 1. Quantize A
         if len(A.shape) == 3:
-            A = A.view(-1, A.shape[-1]).contiguous()
+            A = A.reshape(-1, A.shape[-1])
         CA, CAt, SCA, SCAt, coo_tensorA = F.double_quant(A.to(torch.float16), threshold=state.threshold)
 
         if state.threshold > 0.0 and coo_tensorA is not None:


Gorluxor:
Is there something holding back this PR? 
Are there any performance reasons why reshape is not being used?
I have the same concerns with #550 when using InstructBLIP.

Yang-HangWA:
@Gorluxor  view() merely creates a view of the original tensor, without changing the data in memory, but it requires the input tensor to be contiguous. If not, you can use the contiguous() method to convert it.
reshape() will create a view or a copy of the original tensor as needed. It does not require the input tensor to be contiguous, but it may increase memory overhead.

apbard:
> @Gorluxor view() merely creates a view of the original tensor, without changing the data in memory, but it requires the input tensor to be contiguous. If not, you can use the contiguous() method to convert it. reshape() will create a view or a copy of the original tensor as needed. It does not require the input tensor to be contiguous, but it may increase memory overhead. I think this modification should be effective, I also encountered the same error.

In theory, reshape should be equivalent to view+continuous. In this case, as both were used, there should not be any memory overhead

Gorluxor:
Agree with @apbard; if reshape == continues + view performance-wise, then there shouldn't be performance implications though this should be tested. It may be faster when the memory is already contiguous, and the .continuous() does nothing but return the tensor (per [pytorch documentation](https://pytorch.org/docs/stable/generated/torch.Tensor.contiguous.html)). 
But then the problem would indicate that memory is not contiguous when we reach this step.

TimDettmers:
Thank you for the PR. As discussed, this can have a small performance hit if the tensor is indeed 2D and contiguous. But as far as I know, only a few frameworks do this (for example, metaseq). I think in this case correctness is much more important than performance in edge cases and it's better to merge this. Thank you for your contribution!

PR #570, opened 2023-07-06
Check for existence of key before accessing it.

Marenz:


bitsandbytes/__main__.py
```diff
@@ -63,7 +63,7 @@ def generate_bug_report_information():
         print('')
 
     print_header("LD_LIBRARY CUDA PATHS")
-    lib_path = os.environ['LD_LIBRARY_PATH'].strip()
+    lib_path = os.environ['LD_LIBRARY_PATH'].strip() if 'LD_LIBRARY_PATH' in os.environ else ''
```
rapsealk, line 66:
```suggestion
    lib_path = os.environ.get('LD_LIBRARY_PATH', '').strip()
```

How about using `dict.get()` rather than `key in dict` if it is to get a value?

TimDettmers:
A good fix, thank you!

TimDettmers:
It appears this has been already fixed before, I think I was to slow to merge this, sorry :sweat_smile:

Marenz:
The commented suggestion would have been better anyway, :)

PR #580, opened 2023-07-11
Check LD_LIBRARY_PATH on setup

gembancud:
Issue:
Instructions when compiling to source require setting the destination of binaries in env var LD_LIBRARY_PATH.
This env is not checked during library setup, so unshipped self-compiled binaries will remain undetected.
Soln:
Adds a check to LD_LIBRARY PATH.


TimDettmers:
I have to check this more carefully when I have more time. Thank you for your contribution! Sorry for taking so long to look at this.

akx:
Related: https://github.com/TimDettmers/bitsandbytes/pull/1041#issuecomment-1931634146

matthewdouglas:
I am going to close this as it is stale and we've had a lot of refactoring done around this area. #1041 took care of a lot of this!

When building from source with current instructions using CMake, the libbitsandbytes binary is placed in the same directory as the bitsandbytes package, and we look there by default to load it.

PR #587, opened 2023-07-13
replace private with public https repo URL

BramVanroy:
The information log uses git@github format but that's not the public format URL (you need contributor access to clone in this way). This PR replaces it with the public https URL that should work regardless of any authentication.

Error trace before the change
```
$ git clone git@github.com:TimDettmers/bitsandbytes.git
Cloning into 'bitsandbytes'...
Warning: Permanently added 'github.com,140.82.121.3' (ECDSA) to the list of known hosts.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Solved after changing the command to https


TimDettmers:
This looks right to me, thank you!

PR #595, opened 2023-07-15
Fix  import crash caused by __main__.py reference to nonexistent cuda_setup.main.get_cuda_lib_handle

ihsanturk:
get_cuda_lib_handle no longer exists and is no longer necessary for determining compute capability, and is causing a crash on importing bitsandbytes. this fix removes all lingering reference to it


TimDettmers:
Thank you!

PR #613, opened 2023-07-19
Fix uncaught error due to except clause ordering

VikParuchuri:
PermissionError is a subclass of OSError.  If you put the OSError except clause first, it will always catch all PermissionErrors.  Reordering the clauses ensures that PermissionErrors are correctly ignored.

Without this, all PermissionErrors will raise.


TimDettmers:
Thank you for your contribution! Sorry for being slow on this. I have merged a duplicate pull request.

PR #621, opened 2023-07-21
improve `make clean` target

### LINKED ISSUE #532: `make clean` does not remove shared libraries
I would expect there to be a single make target that removes all build artifacts, resetting the source directory to a clean state. It would usually be called 'clean', although sometimes 'distclean' or 'mrproper' are available as more aggressive options. `git clean` is not always an option, e.g. if the source was extracted from a tarball, or if there are untracked files that should be kept.

While testing build configurations, assuming 'clean' was sufficient, I got misleading results because '.so' files were left in the source directory. Apparently I should have run `make clean cleanlibs`, but I'm not familiar with the idea of a 'cleanlibs' target, so I only figured that out by reading the Makefile:

https://github.com/TimDettmers/bitsandbytes/blob/4395d68cf6beda911c1c3e5633debf15652e2902/Makefile#L127-L134

The current 'clean' target produces confusing output in many cases, which doesn't help the situation:
```
$ make cuda12x_nomatmul CUDA_VERSION=121
(output omitted)
$ pip install .
(output omitted)
$ make clean
Makefile:11: WARNING: CUDA_VERSION not set. Call make with CUDA string, for example: make cuda11x CUDA_VERSION=115 or make cpuonly CUDA_VERSION=CPU
rm build/*
rm: cannot remove 'build/bdist.linux-x86_64': Is a directory
rm: cannot remove 'build/lib': Is a directory
make: *** [Makefile:132: clean] Error 1
$ echo $?
2
```

cebtenzzre:
Make `make clean` remove all build artifacts, and do not warn about CUDA_VERSION when simply running 'clean'.

Fixes #532

---

This seems like the least surprising way for `make clean` to behave:
```
$ make cuda12x_nomatmul CUDA_VERSION=121
$ git status --ignored -uall
On branch fix-make-clean
Your branch is ahead of 'origin/main' by 1 commit.
  (use "git push" to publish your local commits)

Ignored files:
  (use "git add -f <file>..." to include in what will be committed)
        bitsandbytes/libbitsandbytes_cuda121_nocublaslt.so
        build/kernels.o
        build/link.o
        build/ops.o

nothing to commit, working tree clean
$ make clean
rm -rf build/* *.egg*
rm -f bitsandbytes/libbitsandbytes*.so
$ git status --ignored -uall
On branch fix-make-clean
Your branch is ahead of 'origin/main' by 1 commit.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,10 @@ ifeq ($(CUDA_HOME),)
 endif
 
 ifndef CUDA_VERSION
+ifneq ($(MAKECMDGOALS),clean)
 $(warning WARNING: CUDA_VERSION not set. Call make with CUDA string, for example: make cuda11x CUDA_VERSION=115 or make cpuonly CUDA_VERSION=CPU)
 CUDA_VERSION:=
+endif
 endif
 
 
@@ -135,10 +137,5 @@ $(ROOT_DIR)/dependencies/cub:
 	cd dependencies/cub; git checkout 1.11.0
 
 clean:
-	rm build/*
-
-cleaneggs:
-	rm -rf *.egg*
-
-cleanlibs:
-	rm ./bitsandbytes/libbitsandbytes*.so
+	rm -rf build/* *.egg*
+	rm -f bitsandbytes/libbitsandbytes*.so


TimDettmers:
Neat addition, thank you!

PR #622, opened 2023-07-21
Make sure bitsandbytes handles permission errors in the right order

fozziethebeat:
When trying to run bitsandbytes on GCP, I ran into an unexpected permission error during cuda driver lookup

```
  File "/opt/conda/envs/cubrio-trainer/lib/python3.9/site-packages/bitsandbytes/cuda_setup/main.py", line 201, in remove_non_existent_dirs
    raise exc
  File "/opt/conda/envs/cubrio-trainer/lib/python3.9/site-packages/bitsandbytes/cuda_setup/main.py", line 197, in remove_non_existent_dirs
    if path.exists():
  File "/opt/conda/envs/cubrio-trainer/lib/python3.9/pathlib.py", line 1424, in exists
    self.stat()
  File "/opt/conda/envs/cubrio-trainer/lib/python3.9/pathlib.py", line 1232, in stat
    return self._accessor.stat(self)
PermissionError: [Errno 13] Permission denied: '/root/google_vm_config.lock'
```

I looked at the code and everything appeared correct until I discovered that `PermissionError` is a sub-type of `OSError` and thus the first except clause would catch any permission errors.  This fixes that by swapping the except clauses.

### DIFF
--- a/bitsandbytes/cuda_setup/main.py
+++ b/bitsandbytes/cuda_setup/main.py
@@ -196,11 +196,13 @@ def remove_non_existent_dirs(candidate_paths: Set[Path]) -> Set[Path]:
         try:
             if path.exists():
                 existent_directories.add(path)
+        except PermissionError as pex:
+            # Handle the PermissionError first as it is a subtype of OSError 
+            # https://docs.python.org/3/library/exceptions.html#exception-hierarchy
+            pass
         except OSError as exc:
             if exc.errno != errno.ENAMETOOLONG:
                 raise exc
-        except PermissionError as pex:
-            pass
 
     non_existent_directories: Set[Path] = candidate_paths - existent_directories
     if non_existent_directories:


limjiayi:
I can also confirm that this is the right fix, can this be merged soon?

xaptronic:
Hmm - this is my fix for the same problem: https://github.com/TimDettmers/bitsandbytes/pull/715

Could ignoring permission errors mask legit errors?

fozziethebeat:
I've only had this problem on Google VMs, so your fix looks just as good to me.  I figured since the code is already written this way, ordering it to work as intended should be fine.

RemiLacroix-IDRIS:
> Could ignoring permission errors mask legit errors?

I doubt it, the goal of that function is to remove paths that cannot be used to find the CUDA libs anyway.

madhavthaker1:
I'm also running into the same error. Are there plans to merge this fix?

fozziethebeat:
As far as I can tell, this repo has been frozen since early August.  looks like this won't be getting merged anytime soon.

asaiacai:
This also fixes my issue on GCP. Will this be merged soon?

RemiLacroix-IDRIS:
This definitely needs to be addressed.

TimDettmers:
Sorry for the delay on this and thank you for your contribution!

PR #710, opened 2023-08-16
Add version attribute as per Python convention

rasbt:
Just a simple PR so that bitsandbytes has a version attribute, which is common in Python scientific computing libraries (PyTorch, NumPy, scikit-learn, pandas, ...)

So, this PR allows the following now:

```python
>>> import bitsandbytes as bnb
>>> print(bnb.__version__)
"0.41.0"
```


poedator:
@rasbt , it is not critical, but a very convenient feature. I'd like to see it merged. Could you check the reasons for the merge conflict, please?

poedator:
@Titus-von-Koeller , would it be possible to merge this PR somehow? -- it would be helpful in situations when we test multiple bnb versions like 0.41.3.post1.

rasbt:
> @rasbt , it is not critical, but a very convenient feature. I'd like to see it merged. Could you check the reasons for the merge conflict, please?

It was a minor issue due to to the latest BnB update overwriting a line in this PR. Should be fixed.

bitsandbytes/__init__.py
```diff
@@ -24,4 +24,6 @@
     "optim.optimizer.MockArgs": False,
 }
 
+__version__ = "0.41.1"
```
Titus-von-Koeller, line 27: please update the version to the most recent one :)

Titus-von-Koeller (changes requested):
LGTM. Please fix minor comment. Will merge in the morning.

This is a good idea. Thanks a lot for your contribution.

bitsandbytes/__init__.py
rasbt: good call!

poedator: ... which changed yet again today )

rasbt: `0.41.3.post1` should be the current one though, correct? That's currently what we have in the PR here.

TimDettmers:
Thank you, this is a good idea! Sorry for taking so long.

PR #734, opened 2023-08-26
Fixed wget link for installing cuda

reese3222:
Small fix - the wget command was downloading the full HTML page instead of the raw file from github


TimDettmers:
Nice catch!

PR #747, opened 2023-09-05
Feature : Add Intel GPU/SYCL runtime  support

abhilash1910:
In the plan to support an extensive ecosystem spanning from Huggingface, Lora/LLama families, this addition would enable this great FW to run on our GPU cards seemlessly. The discussion started from this merged PR (https://github.com/artidoro/qlora/pull/219)  and currently this PR would eventually enable low precision quantization. (PR is initial draft state).
Tagging @TimDettmers . Thanks for creating this repository.

### DIFF
--- a/csrc/kernels.dp.hpp
+++ b/csrc/kernels.dp.hpp
@@ -1,0 +1,212 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include <sycl/sycl.hpp>
+#include <dpct/dpct.hpp>
+#include <float.h>
+#include "ops.dp.hpp"
+
+#ifndef kernels
+#define kernels
+
+#pragma once
+
+//template <int QUANT_TYPE, typename INP_TYPE, typename COMP_TYPE, typename OUT_TYPE>__global__ void kMatmul_inference_4bit(INP_TYPE *A, unsigned char *B, OUT_TYPE *out, int lda, int ldb, int rowsA, int colsA, int colsB);
+
+template<typename T> extern SYCL_EXTERNAL void kEstimateQuantiles(T *__restrict__ const A, float *code, const float offset, const T max_val, const int n,
+                                            const sycl::nd_item<3> &item_ct1,
+                                            uint8_t *temp_storage_ct1);
+
+extern SYCL_EXTERNAL void kQuantize(float * code, float * __restrict__ const A, unsigned char *out, const int n,
+               const sycl::nd_item<3> &item_ct1, float *smem_code);
+extern SYCL_EXTERNAL void kDequantize(float *code, unsigned char *A, float *out, const int n,
+                 const sycl::nd_item<3> &item_ct1, float *smem_code);
+
+template<typename T, int BLOCK_SIZE, int NUM_PER_TH, int STOCHASTIC, int DATA_TYPE> extern SYCL_EXTERNAL void kQuantizeBlockwise(float * code, T * __restrict__ const A, float *absmax, unsigned char *out, float * __restrict__ const rand, const int rand_offset, const int n,
+                                                                                                            const sycl::nd_item<3> &item_ct1,
+                                                                                                            typename LoadT::TempStorage &loadt,
+                                                                                                            typename LoadFloat::TempStorage &loadf,
+                                                                                                            typename StoreChar::TempStorage &storec,
+                                                                                                            typename BlockReduce::TempStorage &reduce,
+                                                                                                            float *smem_code,
+                                                                                                            float *smem_absmax_value);
+template<typename T, int BLOCK_SIZE, int THREADS, int NUM_PER_TH, int DATA_TYPE> extern SYCL_EXTERNAL void kDequantizeBlockwise(float *code, unsigned char * A, float * absmax, T *out, const int blocksize, const int n,
+                                                                                                           const sycl::nd_item<3> &item_ct1,
+                                                                                                           typename LoadChar::TempStorage &loadchar,
+                                                                                                           typename StoreT::TempStorage &storet);
+
+template<typename T, int OPTIMIZER, int BLOCK_SIZE, int NUM_VALS>
+extern SYCL_EXTERNAL void kPreconditionOptimizer32bit2State(T* g, T* p,
+                float* state1, float* state2, float *unorm,
+                const float beta1, const float beta2, const float eps, const float weight_decay,
+                const int step, const float lr, const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void kOptimizer32bit2State(T* g, T* p,
+                float* state1, float* state2, float *unorm, const float max_unorm, const float param_norm,
+                const float beta1, const float beta2, const float eps, const float weight_decay,
+                const int step, const float lr, const float gnorm_scale, const bool skip_zeros, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER, int BLOCK_SIZE, int NUM_VALS>
+extern SYCL_EXTERNAL void kPreconditionOptimizer32bit1State(T* g, T* p,
+                float* state1, float *unorm,
+                const float beta1, const float beta2, const float eps, const float weight_decay,
+                const int step, const float lr, const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void kOptimizer32bit1State(T* g, T* p,
+                float* state1,  float *unorm, const float max_unorm, const float param_norm,
+                const float beta1, const float beta2, const float eps, const float weight_decay,
+                const int step, const float lr, const float gnorm_scale, const bool skip_zeros, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void
+kPreconditionOptimizerStatic8bit1State(T* p, T* __restrict__ const g, unsigned char*__restrict__  const state1,
+                float *unorm,
+                const float beta1, const float beta2,
+                const float eps, const int step,
+                float* __restrict__ const quantiles1,
+                float* max1, float* new_max1,
+                const float weight_decay,
+                const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1,
+                float *smem_quantiles1);
+
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void
+kOptimizerStatic8bit1State(T* p, T* const g, unsigned char* state1,
+                const float *unorm, const float max_unorm, const float param_norm,
+                const float beta1, const float beta2,
+                const float eps, const int step, const float lr,
+                float* __restrict__ const quantiles1,
+                float* max1, float* new_max1,
+                float weight_decay, const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, float *smem_quantiles1,
+                uint8_t *temp_storage_ct1);
+
+
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void
+kPreconditionOptimizerStatic8bit2State(T* p, T* __restrict__ const g, unsigned char*__restrict__  const state1, unsigned char* __restrict__ const state2,
+                float *unorm,
+                const float beta1, const float beta2,
+                const float eps, const int step,
+                float* __restrict__ const quantiles1, float* __restrict__ const quantiles2,
+                float* max1, float* max2, float* new_max1, float* new_max2,
+                const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, uint8_t *temp_storage_ct1,
+                float *smem_quantiles1, float *smem_quantiles2);
+
+
+template<typename T, int OPTIMIZER>
+extern SYCL_EXTERNAL void
+kOptimizerStatic8bit2State(T* p, T* const g, unsigned char* state1, unsigned char* state2,
+                const float *unorm, const float max_unorm, const float param_norm,
+                const float beta1, const float beta2,
+                const float eps, const int step, const float lr,
+                float* __restrict__ const quantiles1, float* __restrict__ const quantiles2,
+                float* max1, float* max2, float* new_max1, float* new_max2,
+                float weight_decay, const float gnorm_scale, const int n,
+                const sycl::nd_item<3> &item_ct1, float *smem_quantiles1,
+                float *smem_quantiles2, uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER, int BLOCK_SIZE, int N_PER_TH> extern SYCL_EXTERNAL void kOptimizerStatic8bit2StateBlockwise(
+		T* p, T* __restrict__ const g, unsigned char* state1, unsigned char* state2,
+                const float beta1, const float beta2, const float eps, const int step, const float lr,
+                float* __restrict__ const quantiles1, float* __restrict__ const quantiles2,
+                float* absmax1, float* absmax2, float weight_decay, const float gnorm_scale, const bool skip_zeros, const int n,
+                const sycl::nd_item<3> &item_ct1,
+                sycl::local_accessor<float, 2> smem_quantiles1,
+                sycl::local_accessor<float, 2> smem_quantiles2,
+                typename BlockReduce1::TempStorage &reduce1,
+                typename BlockReduce2::TempStorage &reduce2,
+                float *smem_exchange1, float *smem_exchange2,
+                uint8_t *temp_storage_ct1);
+
+template<typename T, int OPTIMIZER, int BLOCK_SIZE, int N_PER_TH> extern SYCL_EXTERNAL void kOptimizerStatic8bit1StateBlockwise(
+		T* p, T* __restrict__ const g, unsigned char* state1,
+                const float beta1, const float beta2,
+                const float eps, const int step, const float lr,
+                float* __restrict__ const quantiles1,
+                float* absmax1,
+                float weight_decay,
+                const float gnorm_scale, const bool skip_zeros, const int n,
+                const sycl::nd_item<3> &item_ct1,
+                sycl::local_accessor<float, 2> smem_quantiles1,
+                typename BlockReduce1::TempStorage &reduce1,
+                float *smem_exchange1, uint8_t *temp_storage_ct1);
+
+
+template<typename T, int BLOCK_SIZE, int NUM_VALS> extern SYCL_EXTERNAL void kPercentileClipping(T * __restrict__ g, float *gnorm_vec, int step, const int n,
+                                                                            const sycl::nd_item<3> &item_ct1,
+                                                                            typename BlockReduce::TempStorage &reduce,
+                                                                            typename LoadT::TempStorage &loadT);
+
+
+void kHistogramScatterAdd2D(float* histogram, int *index1, int *index2, float *src, const int maxidx1, const int n,
+                            const sycl::nd_item<3> &item_ct1);
+
+template <typename T, int SPMM_ITEMS, int BITS>
+extern SYCL_EXTERNAL void kspmm_coo_very_sparse_naive(int *max_count, int *max_idx,
+                                 int *offset_rowidx, int *rowidx, int *colidx,
+                                 sycl::half *values, T *B, sycl::half *out,
+                                 float *__restrict__ const dequant_stats,
+                                 int nnz, int rowsA, int rowsB, int colsB,
+                                 const sycl::nd_item<3> &item_ct1,
+                                 sycl::half *smem_dequant_stats);
+
+template <int ITEMS_PER_THREAD, int SUBTILE_ROWS, int THREADS>
+extern SYCL_EXTERNAL void kdequant_mm_int32_fp16(
+    int *__restrict__ const A, float *__restrict__ const rowStats,
+    float *__restrict__ const colStats, sycl::half *out, float *newRowStats,
+    float *newcolStats, sycl::half *__restrict__ const bias, const int numRows,
+    const int numCols, const int tileCols, const int n,
+    const sycl::nd_item<3> &item_ct1, float *smem_rowStats,
+    typename LoadInt32::TempStorage &loadint32,
+    typename ExchangeInt32::TempStorage &exchangeint32);
+
+template<typename T, int THREADS, int ITEMS_PER_THREAD, int TILE_ROWS, int TILE_COLS, int SPARSE_DECOMP> extern SYCL_EXTERNAL void kgetColRowStats(T * __restrict__ A, float *rowStats, float *colStats, int * nnz_count_row, float nnz_threshold, int rows, int cols, int tiledRows, int tiledCols,
+                                                                                                                              const sycl::nd_item<3> &item_ct1,
+                                                                                                                              uint8_t *temp_storage_ct1,
+                                                                                                                              float *smem_row_absmax_values,
+                                                                                                                              int *smem_row_nnz_values);
+template <int THREADS, int ITEMS_PER_THREAD, int TILE_ROWS, int TILE_COLS,
+          int SPARSE_DECOMP>
+extern SYCL_EXTERNAL void kDoubleRowColQuant(sycl::half *__restrict__ const A,
+                        float *__restrict__ const rowStats,
+                        float *__restrict__ const colStats,
+                        char *out_col_normed, char *out_row_normed, int *rowidx,
+                        int *colidx, sycl::half *val,
+                        int *__restrict__ nnz_block_ptr, float threshold,
+                        int rows, int cols, int tiledCols,
+                        const sycl::nd_item<3> &item_ct1,
+                        typename LoadHalf::TempStorage &loadhalf,
+                        typename StoreInt8::TempStorage &storeint8,
+                        float *smem_row_stats, unsigned int *smem_nnz_row_idx);
+
+
+template <int THREADS, int ITEMS_PER_THREAD, int TILE_ROWS, int TILE_COLS, int TRANSPOSE, int FORMAT> extern SYCL_EXTERNAL void kTransformRowToFormat(char *__restrict__ const A, char *out, int rows, int cols, int tiledCols, int outRows, int outCols,
+                                                                                                                                 const sycl::nd_item<3> &item_ct1,
+                                                                                                                                 char *smem_data);
+
+template <int FORMAT> extern SYCL_EXTERNAL  void kExtractOutliers(char *A, int *idx, char *out, int idx_size, int rowsA, int colsA, int tiledRowsA, int tiledColsA,
+                                            const sycl::nd_item<3> &item_ct1);
+
+template <typename T, int BITS, int THREADS> extern SYCL_EXTERNAL void gemm_device(int M, int N, int K, T * __restrict__ const A,  T* B,  T * out,  int lda, int ldb, int ldc);
+template <typename T, int THREADS>  extern SYCL_EXTERNAL void kgemm_4bit_inference(int M, int N, int K, T * __restrict__ const A, unsigned char *B,  float *absmax, T * out,  int lda, int ldb, int ldc, int blocksize);
+template <typename T, int THREADS, int BITS> extern SYCL_EXTERNAL  void kgemm_4bit_inference_naive(int M, int N, int K, T * __restrict__ const A, unsigned char *B,  float *absmax, const float *datatype, T * out,  int lda, int ldb, int ldc, int blocksize,
+                                                                             const sycl::nd_item<3> &item_ct1,
+                                                                             T *quant_map);
+
+template <typename T, int FUNC> extern SYCL_EXTERNAL void kfunc(T *A, T *B, T value, long n,
+                                           const sycl::nd_item<3> &item_ct1);
+
+#endif
--- a/csrc/ops.dp.hpp
+++ b/csrc/ops.dp.hpp
@@ -1,0 +1,215 @@
+// Copyright (c) Facebook, Inc. and its affiliates.
+//
+// This source code is licensed under the MIT license found in the
+// LICENSE file in the root directory of this source tree.
+
+
+#ifndef ops_H
+#define ops_H
+
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/algorithm>
+#include <sycl/sycl.hpp>
+#include <dpct/dpct.hpp>
+#include <stdio.h>
+#include <iostream>
+#include <unistd.h>
+#include <assert.h>
+#include <dpct/blas_utils.hpp>
+
+#include <dpct/sparse_utils.hpp>
+
+#include <vector>
+#include <functional>
+#include <dpct/dpl_utils.hpp>
+
+
+#define THREADS_PER_BLOCKS (512)
+
+inline void checkCudaStatus(int status) {
+    /*
+    DPCT1000:93: Error handling if-stmt was detected but could not be rewritten.
+    */
+    if (status != 0) {
+        /*
+        DPCT1009:94: SYCL uses exceptions to report errors and does not use the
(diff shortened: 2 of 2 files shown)


abhilash1910:
@Titus-von-Koeller  based on issue #894  , there are some abstractions which are specific to certain intel devices which are to be selectively implemented in Ipex but for a majority of the ops/kernels to be supported across all SYCL devices this integration is needed.  This is a WIP as it requires some dependencies to be resolved on Intel LLVM  to make it runnable across SYCL devices. In general this is a broader approach  catering to SYCL runtime. @jianan-gu

Titus-von-Koeller:
Dear @abhilash1910,

Thanks for submitting this PR and flagging the connection with [#894](https://github.com/TimDettmers/bitsandbytes/issues/894), really appreciated.

Would you be so kind to give a bit more background and describe in detail that you're PR is aiming to solve (once it's complete) and what roadblocks you see before getting there?

Beware, that contrary to Tim, my systems programming knowledge is not so good, yet, but once I understand the above better, I'm happy to coordinate + help along this contribution.

Could you please elaborate specifically where you see overlap + hurdles to resolve relative to [#894](https://github.com/TimDettmers/bitsandbytes/issues/894)?

abhilash1910:
Hi @Titus-von-Koeller 
Thank you.
Yes sure , the main aim of this PR is to provide a common runtime for SYCL kernels - this includes device agnostic kernels which can be made to run on SYCL runtime.  I had started this initially for our Intel GPU devices , which would be now partly covered by the IPEX module (specified in 894) . For users and developers who are not using IPEX runtime entirely , and want the bare metal performance on (nearly all) SYCL runtime support devices - this PR is needed. 
Although the IPEX pathway would provide some intrinsic optimizations through IPEX kernels ( and some specific ISAs which will benefit certain sets of our GPU lineups) , this is a more generic standalone implementation which is decoupled from IPEX and can be compiled and run separately .  The entirety of this port relies on [Intel LLVM's clang](https://github.com/intel/llvm/tree/sycl/clang) which is also being modified to make comparable care metal APIs . SYCL is a runtime built on top of OpenCL (which is the defacto runtime supported in majority of CPUs) and can be aligned as a middleware layer compatible with Nvidia/AMD and other devices supporting Sycl runtime (from Khronos).  
The initial motivation was to provide support to our devices specifically (GPU sets) but with the growing adoption of oneAPI and SYCL on all device backends this would provide another way to users who want to rely on pure compiler runtimes for performance.  
As of yet, there are not much compat hurdles when compared to my colleagues PR (894) as much of IPEX work would be on the IPEX framework (including quant/dequant) .But when it comes to certain functionality issues, our compiler /SYCL backend and LLVM is now getting updated to more robust APIs for translation/and analysis which is why the PR is in draft mode. I am working on the compiler support as well and since most of the translations are in place some part is left which will be done soon. 
Let me know your thoughts and appreciate your help and responses. :)

TimDettmers:
This is great, thank you so much for your contribution!

Can you run me through how a user would use/compile this for their Intel device? A step-by-step procedure would be great to get a easy overview for users.

abhilash1910:
Hi @TimDettmers , @Titus-von-Koeller  as my fellow colleagues are working on the common device abstraction properties , this is an alternative addition to support SYCL backend across vendors .  Since here we are only abstracting the sycl kernels, there is a need for a dedicated cmake for running only the kernels. The compilation of this would require a standalone cmake with the following options:
- intel basekit (oneapi latest release)
- IntelSYCL and IntelLLVM (this would be present within the basekit)
- c++17 standards as SYCL has a deep linkage
- icpx and icx compiler 
- fsycl targets for spirv (Intel) ,ptx(Nv) and amdgcn(AMD)
- Certain compiler level optimizations (narrowing/lowering etc).
This work will be a parallel separate effort  to meet customer requirements who are interested in bare sycl performance ,from the existing effort of my colleagues, and gradually more SYCL organizations would be contributing to this PR for kernels.  I will be sending out a detailed description of our sycl backend support.  Also to get a better idea regarding this, @Titus-von-Koeller  @TimDettmers  if you would be available to discuss via a meeting regarding the strategies - firstly the device common strategy which my colleagues are working on and specially the SYCL kernel path which is the current PR.

matthewdouglas:
@abhilash1910 I have a few questions.

1. What would be the advantage of using the PTX (Nvidia) target on the SYCL implementation vs. existing CUDA backend? I can understand the appeal to have a single abstraction across hardware vendors, but I just can't see shifting away from the CUDA reference implementation. Further, as far as I understand, there's no way to target Apple Silicon.
3. Does this require the plugin from Codeplay to target AMD when using the Intel compiler? I notice that the Codeplay AMD plugin is still labeled as being "beta."
4. Is it possible to build with AdaptiveCpp (aka hipSYCL / Open SYCL)?

abhilash1910:
> @abhilash1910 I have a few questions.
> 
> 1. What would be the advantage of using the PTX (Nvidia) target on the SYCL implementation vs. existing CUDA backend? I can understand the appeal to have a single abstraction across hardware vendors, but I just can't see shifting away from the CUDA reference implementation. Further, as far as I understand, there's no way to target Apple Silicon.
> 2. Does this require the plugin from Codeplay to target AMD when using the Intel compiler? I notice that the Codeplay AMD plugin is still labeled as being "beta."
> 3. Is it possible to build with AdaptiveCpp (aka hipSYCL / Open SYCL)?

@matthewdouglas  Thanks for the questions. I will try to answer them in best way possible:
1. Currently we are ramping up SYCL in terms of performance on different clang compilation targets and the intent is to benchmark kernel performance with respect to CUDA. It is not only for a common abstraction at compiler level, but the findings here allow us to optimize llvm to a great extent and inturn perlocate optimizations to sycl and "xpu" logic as a whole(Since xpu is built on top of sycl abstraction).  Regarding the adoption, there is an overlap regarding running bare kernels (SYCL) in different Intel GPU variations (probably will come up for b&b post the pythonic development is completed in the other PR) and customer demands. In terms of community , we have been getting a lot of requests - for CUDA SYCL it is for internal and external benchmarking purposes . As of yet SYCL does not target Apple .
2. I am working with Codeplay on the internal adoption to support amdgcns , and currently the PR should help optimize the backend for amd sycl as well.  We have been targetting sycl across gpu vendors to see the stability of the common kernel code.
3.AdaptiveCPP is something which we are discussing internally for amd specifically, but until we have established a SYCL interface , we will eventually  go ahead with the discussion & strategy. 

Also on other note, I would like to add that SYCL support specifically for kernels would eventually be one of the ways for Intel GPU upstream effort that is ongoing.  And having SYCL backend opens a lot of Intel (not only) but also other devices for competitive benchmarking.

abhilash1910:
Seeing the development update in #898 on a separate branch, I do agree branch separation of backends will be helpful . While the common device abstraction is the probable goal to merge, I think this PR would benefit from a standalone SYCL branch . Since we plan to integrate full sycl support on all kernel adaptations , this would be helpful to maintain as a separate branch.
Would appreciate your thoughts, @TimDettmers @Titus-von-Koeller @matthewdouglas .
PS: The PR would still take some amount of time to complete and would be eventually assisted by Codeplay.

matthewdouglas:
@abhilash1910 That sounds reasonable to me and should result in more iterative, digestible PRs until it's ready for prime time.

csrc/sycl/ops.cpp
```diff
@@ -1998,36 +2018,68 @@ template <typename T> void gemm_4bit_inference(int m, int n, int k, T * A,  unsi
...
+  sycl::queue &q_ct1 = dev_ct1.in_order_queue();
+  sycl::context ctx = q_ct1.get_context();
+    
+  int size = NUM_BLOCK;
+  *((T **)&buff_A) = sycl::malloc_device(size, A, ctx);
+  q_ct1.memcpy((T*)(buff_A), (T*)(A), size);
+  *(( unsigned char**)&buff_B) = sycl::malloc_device(size, B, ctx);
+  q_ct1.memcpy((unsigned char*)(buff_B), (unsigned char*)(B), size);
+  *((T **)&buff_out) = sycl::malloc_device(size, out, ctx);
+  q_ct1.memcpy((T*)(buff_out), (T*)(out), size);
```
matthewdouglas, line 2082: I'm just curious: is there a way to avoid all of this copying?

abhilash1910: Since each kernel call is initiated through ops , we have to dedicate some usm buffers for device processing. This is although a primary iteration , and i am thinking of ways to reduce the initiation duplication of device buffers.

csrc/sycl/ops.cpp
```diff
@@ -0,0 +1,2248 @@
...
+void dequantize(float *code, unsigned char *A, float *out, int n)
+{
+  int num_blocks = n/1024;
+  num_blocks = n % 1024 == 0 ? num_blocks : num_blocks + 1;
+  dpct::device_ext &dev_ct1 = dpct::get_current_device();
+  sycl::queue &q_ct1 = dev_ct1.in_order_queue();
+  sycl::context ctx = q_ct1.get_context();
+  int size = NUM_BLOCK;
+  
+  *((unsigned char **)&buff_A) = sycl::malloc_device(size, A, ctx);
```
JackAKirk, line 158: Is there a good reason to expose the sycl context here, instead of just passing the queue and keeping it hidden?

csrc/sycl/kernels.cpp
```diff
@@ -0,0 +1,5567 @@
...
+      valid_items_load = (n+1)/2 - i > TILE_SIZE ? TILE_SIZE : (n+1)/2 - i;
+      valid_items_store = n - i*2 > TILE_SIZE*2 ? TILE_SIZE*2 : n - i*2;
+    }
+    else
+    {
+      valid_items_load = n - i > TILE_SIZE ? TILE_SIZE : n - i;
+      valid_items_store = n - i > TILE_SIZE ? TILE_SIZE : n - i;
+    }
+    /*
+    DPCT1098:92: The '*' expression is used instead of the __ldg call. These two expressions do not provide the exact same functionality. Check the generated code for potential precision and/or performance issues.
```
JackAKirk, line 1018: For `__ldg` you should use `sycl::ext::oneapi::experimental::cuda::ldg`

See : https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_cuda_tex_cache_read.asciidoc

JackAKirk: For some background to __ldg, check out https://codeplay.com/portal/blogs/2023/09/11/nvidia-texture-cache-in-sycl

csrc/sycl/ops.h
```diff
@@ -0,0 +1,215 @@
...
+template<typename T> void percentileClipping(T * g, float *gnorm_vec, int step, const int n);
+
+extern SYCL_EXTERNAL void histogramScatterAdd2D(float* histogram, int *index1, int *index2, float *src, int maxidx1, int n);
+
+void gemmex(Context * context, bool transposeA, bool transposeB, int m, int n, int k, void *A, void *B, void *C, int lda, int ldb, int ldc);
+void strided_gemmex(Context *context, bool transposeA, bool transposeB, int m, int n, int k, void *A, void *B, void *C, int lda, int ldb, int ldc,
+                    long long int strideA, long long int strideB, long long int strideC, int batchCount);
+
+
+template <int FORMATB, int DTYPE_OUT, int SCALE_ROWS> int igemmlt(dpct::queue_ptr ltHandle, int m, int n, int k, const int8_t *A, const int8_t *B, void *C, float *row_scale, int lda, int ldb, int ldc);
```
JackAKirk, line 174: Essentially the whole thing results on the key question of whether there are any existing libraries like oneDNN/oneMKL blas etc that support a matrix "order" enum that we can assign appropriate values to to coincide with cublasLtOrder_t, such that e.g. oneDNN cuda backend can then call the appropriate cublaslt interface with the appropriate native enum value.

Without this it will not be possible to call the appropriate cublaslt function. If you are lucky and such an interface already exists in oneDNN, then eventually there will be applications that require more of the logical structure exposed in https://docs.nvidia.com/cuda/cublas/#using-the-cublaslt-api

Essentially you need a library that supports things like matrix format arch generation specific enums for arch generation specific specialization of kernels etc.

Without this we won't be able to do much with this. Until this question has been addressed I don't really see much point in us reviewing this further.

matthewdouglas: @JackAKirk For a SYCL+CUDA target, wouldn't we be able to just specialize here and use cuBLASLt directly?
For other targets, does this matter? We could restrict to the equivalent of `CUBLASLT_ORDER_COL` and `CUBLASLT_ORDER_ROW` if we had to, I would think?

JackAKirk: > @JackAKirk For a SYCL+CUDA target, wouldn't we be able to just specialize here and use cuBLASLt directly?
>
I'm not sure if you could currently use it directly even if you don't care about specializing for a particular nvidia gpu generation: this may be possible since bitsandbytes doesn't seem to use many of the other enums exposed in cublaslt that are e.g. used for problem specific implementation decisions: an example here is the epilogue type.
But even if you can call cublaslt assuming e.g. AMPERE (sm_80), then you won't be able to also make it fast on TURING (where the optimized impl is completely different), unless whatever sycl wrapper library you use exposes the arch specific enum values that cublaslt uses.

> For other targets, does this matter? We could restrict to the equivalent of `CUBLASLT_ORDER_COL` and `CUBLASLT_ORDER_ROW` if we had to, I would think?

For other targets you may not be able to reuse much of the code (I haven't thought about this deeply, and I may be wrong here, but I'd be surprised if you could reuse much) since this is all specialized to nvidia hardware (and indeed specific arch generations of nvidia hardware). The matrix multiplication hardware on intel gpus is completely different. They support completely different sizes of low level gemm operations.

JackAKirk: > @JackAKirk For a SYCL+CUDA target, wouldn't we be able to just specialize here and use cuBLASLt directly? For other targets, does this matter? We could restrict to the equivalent of `CUBLASLT_ORDER_COL` and `CUBLASLT_ORDER_ROW` if we had to, I would think?

I think you have to compare the cublaslt specification (https://docs.nvidia.com/cuda/cublas/#using-the-cublaslt-api) with the oneDNN spec carefully. This is is the only way to answer such questions. The point I made is that until the answers to these questions are known, I don't see how much progress can be made.

matthewdouglas: On the python side, bitsandbytes already does specialize:

```python
def get_special_format_str():
    if not torch.cuda.is_available():
        return "col_turing"
    major, _minor = torch.cuda.get_device_capability()
    if major <= 7:
        return "col_turing"
    if major == 8:
        return "col_ampere"
    return "col_turing"
```
The `bitsandbytes.functional.transform()` function is used (which calls the C++ library `ctransform_row2turing, ctransform_row2ampere, etc). Then we have specific implementations in the C++ library: `cigemmlt_turing_32, cigemmlt_turing_8, cigemmlt_ampere_32, cigemmlt_ampere_8`

JackAKirk: > @JackAKirk For a SYCL+CUDA target, wouldn't we be able to just specialize here and use cuBLASLt directly? For other targets, does this matter? We could restrict to the equivalent of `CUBLASLT_ORDER_COL` and `CUBLASLT_ORDER_ROW` if we had to, I would think?

Oh sorry I think I misunderstood what you meant at first: previously it had been stated to me that cublaslt calls would be ported to some sycl library irrespective of the target.
I think that the option of calling native cuda libraries in the sycl (cuda) backend would also be an option that could be investigated. I think this is what you mean.

matthewdouglas: I do tend to agree though, outside of NVIDIA hardware there will need to be new implementations for this that I wouldn't expect e.g. SYCLomatic / DPC++ migration tool to be able to handle. I'm not really sure what it looks like for HIP either.

There's lots to do, so one step at a time.

matthewdouglas: That's essentially what I mean.

Point of further clarification: I don't expect we would really depart from the native CUDA implementation we have today as a default for NVIDIA hardware. The primary use case for SYCL should be around non-NVIDIA hardware. That said I don't see why we couldn't have a mix of SYCL+CUDA code as an option, maybe experimentally, but IMHO it shouldn't be a priority.

JackAKirk: > That's essentially what I mean.
> 
> Point of further clarification: I don't expect we would really depart from the native CUDA implementation we have today as a default for NVIDIA hardware. The primary use case for SYCL should be around non-NVIDIA hardware. 

Totally makes sense to me. I'm just reviewing in the capacity my team has been asked to, which is regards to enabling a port that supports the cuda backend and allows good performance.

abhilash1910: Yes on this topic, there are a couple of approaches (since gemms are not ported yet). onemkl blas porting mirroring the cublaslt apis , which would involve segregation of descriptors as it would not be a 1:1 mapping. An alternate route is oneDNN . Now since cublasLT is closed source, it is difficult to ascertain the improvements met from using cublasLt over cudnn calls for gemm.  In terms of SYCL support of cublasLt, it is a WIP to integrate similar interface . For ptx backends, I think that the SYCL api (when made available) for cublas would be compatible with nv devices, whereas an option for keeping cublas calls as it is when building sycl code sounds good as well. 
But since primarily we plan to integrate intel sycl support, I think that a translation is must needed in this case.

csrc/sycl/kernels.cpp
```diff
@@ -1020,7 +1020,7 @@ SYCL_EXTERNAL void kDequantizeBlockwise(float *code, unsigned char * buff_A, flo
     /*
     DPCT1064:96: Migrated __ldg call is used in a macro/template definition and may not be valid for all macro/template uses. Adjust the code.
     */
-    local_abs_max = absmax[(i+item_ct1.get_local_id(2)*NUM_PER_TH)/(blocksize)];
+    local_abs_max =  sycl::ext::oneapi::experimental::cuda::ldg(&absmax[(i+item_ct1.get_local_id(2)*NUM_PER_TH)/(blocksize)]);
```
matthewdouglas, line 1023: Is this implemented for backend other than CUDA? I can't really find anything in the docs about it other than how this basically translates to the [ld.global.nc PTX instruction](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#data-movement-and-conversion-instructions-ld-global-nc).

JackAKirk: Yes it is portable across all backends as described in the spec document: https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_cuda_tex_cache_read.asciidoc#backend-support-status
However it is a no-op on other backends: it simply returns the value, and does not use any special hardware. For Nvidia it forces usage of the read-only texture cache that can considerably improve performance for relevant use cases.

For information on the purpose/performance of it see https://codeplay.com/portal/blogs/2023/09/11/nvidia-texture-cache-in-sycl

csrc/sycl/kernels.cpp
```diff
@@ -0,0 +1,5528 @@
...
+      int idx = (index1[i]*maxidx1) + index2[i];
+      dpct::atomic_fetch_add<sycl::access::address_space::generic_space>(&histogram[idx], src[i]);
+  }
+}
+
+void warpreduceKernelMax(int* data, const sycl::nd_item<3> &item_ct1) {
+  int threadid = item_ct1.get_local_id(2);
+  int input = data[threadid];
+  int output = 0;
+  output = sycl::reduce_over_group(item_ct1.get_sub_group(), input, sycl::maximum<>());
```
JackAKirk, line 523: This doesn't look like it has been written in a portable way: e.g. see https://developer.codeplay.com/products/oneapi/amd/2024.1.0/guides/troubleshooting#sub-group-size-issues-in-codes-ported-across-platforms-architectures

csrc/sycl/kernels.cpp
```diff
@@ -0,0 +1,5528 @@
...
+  }
+  
+  float warp_max;
+  sycl::host_accessor hacc_values{buff_values};
+  sycl::host_accessor hacc_indices{buff_indices};
+  for(int i = 0; i < 8; i++)
+  {
+    // 3. do warp reduction + broadcast back
+    
+    output = sycl::reduce_over_group(item_ct1.get_sub_group(), max1, sycl::maximum<>());
```
JackAKirk, line 604: Same here re portability, if you ported this directly from cuda code you can only expect it be functionally identical on devices of subgroup size 32.

csrc/sycl/kernels.cpp
```diff
@@ -0,0 +1,5528 @@
...
+   /*
+   DPCT1082:16: Migration of nvcuda::wmma::fragment<wmma::matrix_a, 8, 32, 16, half, wmma::row_major> type is not supported.
+   */
+   /*
+   DPCT1082:17: Migration of nvcuda::wmma::matrix_a type is not supported.
+   */
+   /*
+   DPCT1082:18: Migration of nvcuda::wmma::row_major type is not supported.
+   */
+   sycl::ext::oneapi::experimental::matrix::joint_matrix<sycl::sub_group, sycl::half, sycl::ext::oneapi::experimental::matrix::use::a, 8, 16, sycl::ext::oneapi::experimental::matrix::layout::row_major> a_frag;
```
JackAKirk, line 4356: This code is only portable across different Nvidia devices: see https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_matrix/sycl_ext_oneapi_matrix.asciidoc#nvidia-tensor-cores-supported-combinations

These parameter combinations (the matrix sizes) are only supported in the cuda backend. Unfortunately different matrix core vendors do not support the same range of sizes on matrix tiles. Hence, whilst the interfaces are portable, the parameters are not. So if you want to support more than cuda you need to write a new version for each additional backend (intel AMX,XMX or amd Matrix Cores) if you want to still use joint_matrix.

csrc/sycl/kernels.cpp
matthewdouglas: I think we can just remove `kCompressMax` as it does not appear to be used at all. There's a few other areas where there might be some dead code, at least on the CUDA side where this is sourced from.

csrc/sycl/kernels.cpp
abhilash1910: For AMX or XMX, and other cores, does a runtime macro design pattern look ok?  I am trying to figure out how to link separate instruction calls for different vendors. Suggestions appreciated.

csrc/sycl/kernels.cpp
abhilash1910: Yes kcompressmax (+ some other cuda code) is pretty much unused. I will clean up where-ever dead code resides.

csrc/sycl/kernels.cpp
JackAKirk: The only other option I know of is to pass a backend query to the kernel that can be queried at runtime. MACRO might make more sense I guess for something like this. 
If you don't want to rely on a bitsandbytes macro then these backend macros get defined: depending on the build: https://github.com/codeplaysoftware/SYCL-samples/blob/main/src/MPI_with_SYCL/send_recv_buff.cpp#L16

Now in theory two could be defined at the same time.. but this might not be a concern.

Maybe there is some other option. But I don't know it.

Titus-von-Koeller:
Dear @abhilash1910,

thanks so much for your ongoing work on this topic and @matthewdouglas , our esteemed community member, for contributing as well.

> Since we plan to integrate full sycl support on all kernel adaptations , this would be helpful to maintain as a separate branch.

Yes, sure, we can do that. Sorry for taking a while to get back to you. I also have to admit that this whole SYCL topic is still a bit outside of my comfort zone, so it's difficult for me to give meaningful feedback and evaluate your work.

Thanks @matthewdouglas for mentioning to me in private a bit more about SYCL and how it could aid in the BNB mission.

For the time being, I still have to focus largely on other topics, but I appreciate this effort.

What should I base the `sycl-refactor` branch off of? We would do the same as with the `multi-backend-refactor` branch targets PRs onto it instead of `main`, am I understanding you correct?

I'll also try to set up nightly releases for those two branches in the next weeks. Then it will be easier for testing and integration with other software in the eco system.

abhilash1910:
> Dear @abhilash1910,
> 
> thanks so much for your ongoing work on this topic and @matthewdouglas , our esteemed community member, for contributing as well.
> 
> > Since we plan to integrate full sycl support on all kernel adaptations , this would be helpful to maintain as a separate branch.
> 
> Yes, sure, we can do that. Sorry for taking a while to get back to you. I also have to admit that this whole SYCL topic is still a bit outside of my comfort zone, so it's difficult for me to give meaningful feedback and evaluate your work.
> 
> Thanks @matthewdouglas for mentioning to me in private a bit more about SYCL and how it could aid in the BNB mission.
> 
> For the time being, I still have to focus largely on other topics, but I appreciate this effort.
> 
> What should I base the `sycl-refactor` branch off of? We would do the same as with the `multi-backend-refactor` branch targets PRs onto it instead of `main`, am I understanding you correct?
> 
> I'll also try to set up nightly releases for those two branches in the next weeks. Then it will be easier for testing and integration with other software in the eco system.

Thanks @Titus-von-Koeller . The intention of the SYCL code is to make it runnable across vendors as a pure C++ runtime implementation. I think the best solution would be to base ```sycl-refactor``` branch out of main as my colleagues are working on the ```multi-backend-refactor```.  The reason is that this code base adapts from the centralised cuda cpp code mainly , and it being mainly another cpp abstraction for multiple vendors it would make sense to base it out of main.  The complexity of this PR being very high as regular refining of the apis and compiler backend standards  are done/being done to ensure compatibility across vendors and hence the slow progress (one step at a time).    Currently there is a large section of users waiting on this feature for pure SYCL performance (both in Intel & beyond) . I am also open to suggestions and thanks to @matthewdouglas  for support .  Nightly releases are definitely going to be perfect in the coming times after the initial functional PR is ready.

Titus-von-Koeller:
Ok, created the new `sycl-refactor` branch off of `main` and switched the PR to point that way, as prev agreed.

Unfortunately, I was off sick 2 days this week, which cost me valuable work time. I gotta focus first on refactoring the device abstraction based on some new insights how to better integrate with PyTorch on that subject. One of the things I'll focus on next will be setting up nightly releases of the `multi-backend-refactor` and `sycl` branches, but can't promise a timeline yet.

@abhilash1910 Regarding adding an Intel specific Github runner to `bitsandbytes`. I'm in contact with Ke from Intel about that.

Does the SYCL stuff need any special considerations regarding continuous integration? Do I assume correctly that the Intel side of running the SYCL code can be tested + maintained with the same hardware as the other PRs from Intel?

abhilash1910:
Thanks @Titus-von-Koeller , hope you are feeling well . 
No worries as the PR is progressing slowly to adhere to latest standards. 
Yes for SYCL since the main idea is to not only run on Intel hardware but also others, I think  other Nv  /Intel or AMD systems can be used for CI. Currently I am testing the changes on Intel hardware as well as Nv a100. 
The Intel specific PRs which are specific to CPU/XPU  which use the IPEX pathway ( would require a hosted runner for CI ) . Great that you are in contact with my colleague Ke regarding this.  For CICD on the SYCL side the same hosted runner should work just fine, and I think to best test regressions even Nv hardware can be used (which I believe is present in the existing CI stack ). 
Since SYCL eventually boils to SPIRV for Intel and PTX codes for NV, I think any running CI can be used to test full convergence and functionally .  Thanks again for  all the work.

abhilash1910:
Yes these are I think the prospective answers:
1. This is something which really requires a lot of collaborative thought. SYCL being a language to support hardware semantics can be leveraged (through different device specific cmake args) to build for separate runtimes & devices. When the torch extensions arising from these separate cmakes are used to build and make the .so files, we can plug in the torch dispatcher as we normally would in any application.  The only catch here is that there would be separate binaries for separate devices which need wrapping with torch . I would also ping @JackAKirk @npmiller @ruyk @AerialMantis for their thoughts on this. 

2. Yes this is something which is very interesting as to see the performance , specifically of gemm kernels  and how much is the gap in perf when it comes to the pure cublas implementation. It would be SYCL runtime compiled for ptx targets vs native Cuda runtime performance. This would give a good idea of further improvement of SYCL standards in general. 

3. SYCL being a language in its entirety does need extensions to be pluggable with torch. The extensions would come after we compiled the binaries (from SYCL) and link device specific binaries with torch extensions. For instance we have DPCPPExtension for SYCL for SPIRV (Intel targets) and standard CUDA_extensions for (Cuda or in some cases HIP). The later case requires the binaries to be compatible with ptx /cuda or amdgcn targets.

4. There are a couple of considerations on this since we have a standing approachable PR to torch for Intel GPU Support (which contains sycl kernels).  It is actually a separate non interfering branch which requires some thoughts.  (I am keeping this open for now until I  think of a best way if intersection is possible).

abhilash1910:
@Titus-von-Koeller  A detail on this PR:  I am nearing completion of the kernels . Since the LOC is quite big for this PR , I think the pytorch aspect of linking the binaries can be extended in a separate PR.   I am open to suggestions and can add here as well but it would make the review process very hectic. 
Currently the kernels are almost finalised (barring some small changes), it would be great if the review process is also started .
I have a collection of kernel wise test scripts separately ( in native c++ without invoking pt backend) which I can add as a benchmark test suite for sycl . Appreciate thoughts from other reviewers as well. Thanks

matthewdouglas:
One point to note is that bitsandbytes isn't built as a PyTorch extension currently for any of the backends. Instead the shared lib exports a C API that is invoked with Python's `ctypes` module.

This SYCL backend also does not need to link to libtorch or even python, and since it exposes the same API (see pythonInterface.cpp), it should be possible to just build multiple targets (e.g. spirv, ptx, amdgcn) and load the correct .so at runtime; no CUDAExtension or DPCPPExtension needed.

abhilash1910:
> One point to note is that bitsandbytes isn't built as a PyTorch extension currently for any of the backends. Instead the shared lib exports a C API that is invoked with Python's `ctypes` module.
> 
> This SYCL backend also does not need to link to libtorch or even python, and since it exposes the same API (see pythonInterface.cpp), it should be possible to just build multiple targets (e.g. spirv, ptx, amdgcn) and load the correct .so at runtime; no CUDAExtension or DPCPPExtension needed.

Yes this is something which I noticed, and in that case the kernels can be built for multiple targets.  
At runtime the kernels can be offloaded without linking to the respective pytorch bindings for the devices. 
Thanks for the details @matthewdouglas

abhilash1910:
Hi @Titus-von-Koeller , pinging for a review in this case (albeit some cosmetic changes are WIP) . After some discussion internally , this PR would only consist of the SYCL kernels  only so that these can be abstracted across vendors.  Since the kernels are mostly completed I would request start of the review process ; also there is a test suite which has been created  for SYCL and wanted to discuss how to add that.
 FYI: There  is another PR from my collegaue with xpu support https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1257 which is using ipex for xpu (Intel GPU) support only. We are discussing better ways to formulate it on the multibackend refactor branch.

abhilash1910:
Pinging @Titus-von-Koeller @matthewdouglas with latest updated code. 
cc @JackAKirk , @npmiller

Titus-von-Koeller:
Dear @abhilash1910,

We (the bitsandbytes team) are under very tight bandwidth at the moment so we won't have time to deep dive into it just yet. Currently we're both trying to get out of a maintenance impasse and preparing for the Intel CPU+XPU, AMD ROCm alpha release. Additionally, we're working on infra, pkging and testing improvements. Also, the multi-backend stuff still needs a major refactor with everyone waiting for us to finalize the backend abstraction with a move to the `torch.library` API for registering custom ops and dispatching through the PyTorch dispatcher.

These things currently take priority.

We'll however try to take a look as soon as we can and with @matthewdouglas having just been added to the team full-time, we're in a better position going forward.

Titus-von-Koeller:
> Also there is a test suite which has been created for SYCL and wanted to discuss how to add that

@abhilash1910 [are these the tests you were referring to](https://github.com/abhilash1910/bitsandbytes_sycl_samples/tree/main/tests_pvc)?

abhilash1910:
Hi @Titus-von-Koeller 
Thanks for the details. Yes it is infact just the public version of it - there is an internal testing suite which is present. PLan to add that to my personal repository - until we see how to add to sycl-refactor. 
For the XPU logic addition, I believe some of these kernels would get eventually used up for 1257.  I am aware of the major refactoring going on for the multi-factor backend and for XPU specifically we are currently prototyping for the full integration. 
Some context on this work, currently we are examining kernel-kernel performance on primarily a100 / iGPU (datacenter) for public competitive benchmarking purposes.  I believe that down the line this branch would evolve to be a sycl runtime only version of bitsandbytes for compiler level performance & analysis. 
Appreciate the work you guys are doing , and I am taking a look into the XPU upsteaming logic as well  internally for multi-backend refactor.

matthewdouglas:
Hi everyone,
This PR is quite stale and we've since merged in some SYCL kernels for XPU. If there are parts of this that may make sense to reuse for CPU or XPU, then we could be open to that in an isolated PR. However, we're not going to be able to invest in the use of SYCL as a generic backend.

Thank you for understanding.

PR #761, opened 2023-09-11
patch `requires_grad` error for ints

gszep:
this patch bypasses the error https://github.com/pytorch/pytorch/issues/83913. In some cases we cannot assume that `requires_grad=False` in the modified block


TimDettmers:
The Int4 layers do not have weight gradients. So I am hesitant to merge this since it could introduce silent unexpected behavior in many cases. Maybe I am missing the point here: why do you want to set `requires_grads=True`?

gszep:
Sorry I was being a dum dum. You're right. Closing the PR 😁

PR #753, opened 2023-09-07
Save and load  in NF4 / FP4 formats

poedator:
Purpose: enable saving and loading transformers models in 4bit formats.
Enables this PR in transformers: https://github.com/huggingface/transformers/pull/26037

addresses feature request #603 and other similar ones elsewhere.

tested with Bloom-560 and Llama-2-7b. 
tested quantization, saving, loading,  -- matched tensors and quant_states -- matched inference results.
tested both nf4 and fp4
test added

PR #763, opened 2023-09-11
Refactoring of quant_state

poedator:
Converted `quant_state` from nested list into nested class `QuantState`.

Rationale: to have a stable and pretty format before enabling saving of 4-bit quantized models.

this PR enables #753


TimDettmers:
It seems changes for the quant state are also in #753. Is this correct? If so, lets close this and move our discussion there.

poedator:
> It seems changes for the quant state are also in #753. Is this correct? If so, lets close this and move our discussion there.

actually #763 is not repeating #753, but is a prerequisite. I started from just saving but then went for the Q-state refactor. So you may want to look at the diff at this one.  #753 will only show diff over #763, and #763 should be merged first.

I will address some of the comments from #753 in this PR. Soon.

JaclynCodes (commented):
:)

PR #776, opened 2023-09-17
WIP: dequantize nf4 to int8

sshleifer:


TimDettmers:
This is a great addition to the library -- thank you, Sam!

If you try to create an int8 template overload, does it compile, or would their other changes be needed? I would want to avoid an explicit specialization of the template here.

TimDettmers:
This is a great addition to the library -- thank you, Sam!

If you try to create an int8 template overload, does it compile, or would their other changes be needed? I would want to avoid an explicit specialization of the template here.

PR #788, opened 2023-09-24
CMake and Windows Compilation

Jamezo97:
I've attempted to setup a general CMakeLists.txt that will work cross platform, including on Windows.

To get things building on Windows some code changes were required.
1) Replace pthread with std thread. I've tried to keep the code consistent and fast here.
2) Explicitly declare all combinations of templated functions. I'm surprised this compiled before, I must be missing something here?
3) Don't use the `IsComplete` template meta, the MSVC compiler gets pretty sad. I've switched it out with something that's technically equivalent given the current usage of the class.

Hopefully this is useful.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,124 @@
+# This CMake config hopefully makes it easier to compile.
+# Ensure the CUDA Toolkit is available on your path. Then run:
+#   For  GCC: `cmake -B build . && cmake --build build`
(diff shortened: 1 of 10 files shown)


TimDettmers:
This is a very important contribution. Thank you for being so patient. We will work on this soon.

Jamezo97:
Closing due to changes merged in #908

wkpark:
thank you for your work!

PR #792, opened 2023-09-27
Use /usr/bin/env to find the correct path to g++

framp:
Building the project fails on nixos (and everywhere else g++ is not in a standard path)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 MKFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
 ROOT_DIR := $(patsubst %/,%,$(dir $(MKFILE_PATH)))
 
-GPP:= /usr/bin/g++
+GPP:= /usr/bin/env g++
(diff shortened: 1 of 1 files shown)


TimDettmers:
Thank you, this is an important issue. However, for me g++ is not in that path (it is `/usr/bin/g++` for me). I think a proper fix would check multiple paths and select the one that works.

framp:
using env should get the path to g++ as long as it's in the $PATH 

I've built my fork on nixos and arch linux with success

rickardp:
I think this change is superseded by cmake.

@framp Can you verify if you can build now using the updated instructions in compiling.md?

framp:
I can confirm this is fixed! Thanks for taking care of this issue ^^

I'm not running nixos anymore (I got tired of ML related packages not being up to date or not working and I switched to Arch Linux) but I ran it inside a VM and g++ is detected correctly

Screenshot 2024-02-09 at 18 49 00

Unfortunately I can't test the full details as I don't have cuda available

Issue #675, opened 2023-08-04
Permission denied error in cuda_setup when any env var contains /root path


osma:
# My setup

Linux, Python 3.10.8, CUDA 12, `pip install bitsandbytes==0.41.0`

This is a Slurm cluster and there is an environment variable like this:

    SLURM_SUBMIT_DIR=/root

# The problem

Running `python -m bitsandbytes` fails with a permisison error. It appears that `bitsandbytes` decides to check if the path `/root` contains `libcudart.so` because this path appears in an environment variable.

Note that my `LD_LIBRARY_PATH` contains the real path to the CUDA libraries. cuda_setup does find them, but it doesn't stop there; instead it goes on to check all kinds of other random environment variables, including `SLURM_SUBMIT_DIR` in my situation, in case they also happen to contain CUDA libraries, and then stumbles on the permission error when it can't read files under `/root`. FWIW, I think this is also a logic error (it should stop when it finds the CUDA libraries using `LD_LIBRARY_PATH` and look no further) but that's beyond the scope of this report.

```
$ python -m bitsandbytes
Traceback (most recent call last):
  File "/appl/easybuild/opt/Python/3.10.8-GCCcore-12.2.0/lib/python3.10/runpy.py", line 187, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/appl/easybuild/opt/Python/3.10.8-GCCcore-12.2.0/lib/python3.10/runpy.py", line 146, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/appl/easybuild/opt/Python/3.10.8-GCCcore-12.2.0/lib/python3.10/runpy.py", line 110, in _get_module_details
    __import__(pkg_name)
  File "/wrk-vakka/users/xxx/llama2-finetune/venv/lib/python3.10/site-packages/bitsandbytes/__init__.py", line 6, in <module>
    from . import cuda_setup, utils, research
  File "/wrk-vakka/users/xxx/llama2-finetune/venv/lib/python3.10/site-packages/bitsandbytes/research/__init__.py", line 1, in <module>
    from . import nn
  File "/wrk-vakka/users/xxx/llama2-finetune/venv/lib/python3.10/site-packages/bitsandbytes/research/nn/__init__.py", line 1, in <module>
    from .modules import LinearFP8Mixed, LinearFP8Global
[...]

PR #812, opened 2023-10-09
[Bug Fix] Fix Permission denied error in cuda_setup

KaiserWhoLearns:
Current path checking seems to introduce Permission Error when checking paths where the user do not have access to ([Issue 810](https://github.com/TimDettmers/bitsandbytes/issues/810) and [Issue 675](https://github.com/TimDettmers/bitsandbytes/issues/675)).

Combined the fix in [Pull Request644](https://github.com/TimDettmers/bitsandbytes/pull/644), along with another place that calls `(path / libname).is_file()` without checking user access.

Note: This problem will likely only happen when a user without root access tries to use the package, while the root user has some installations of CUDA.


TomasAndersonFang:
Thanks, it solves my problem!!!

manishiitg:
this solved my problem as well!!

swumagic:
Bitsandbytes was not supported windows before, but my method can support windows.（yuhuang）
1  open folder J:\StableDiffusion\sdwebui，Click the address bar of the folder and enter CMD
or WIN+R, CMD  。enter，cd /d J:\StableDiffusion\sdwebui
2  J:\StableDiffusion\sdwebui\py310\python.exe -m pip uninstall bitsandbytes

3 J:\StableDiffusion\sdwebui\py310\python.exe -m pip uninstall bitsandbytes-windows

4 J:\StableDiffusion\sdwebui\py310\python.exe -m pip install https://github.com/jllllll/bitsandbytes-windows-webui/releases/download/wheels/bitsandbytes-0.41.1-py3-none-win_amd64.whl

Replace your SD venv directory file（python.exe Folder） here（J:\StableDiffusion\sdwebui\py310）

TimDettmers:
I merged a different fix which changes the order of the permission try-catch. Would this solution have an additional advantage over the other?

KaiserWhoLearns:
> I merged a different fix which changes the order of the permission try-catch. Would this solution have an additional advantage over the other?


Were you referring to the first try-catch that was changed? For the original order, the `PermissionError` will directly go into the `OSError` as it's a subclass of OSError, so the later `except` will never be reached.

Titus-von-Koeller:
Ok, thanks for raising this issue and taking the initiative to solve it. Looking forward to merge your next contribution!

PR #832, opened 2023-10-20
Fix parameter name in error message

michaelmior:
In some error messages, a parameter was referred to as `bnb_4bit_compute_type` instead of `bnb_4bit_compute_dtype` (note the missing `d`).


michaelmior:
This typo seems to still exist.

TimDettmers:
Thank you!

PR #847, opened 2023-10-27
Applying the jetson fixes

ivansmith7795:


ivansmith7795:
Merging changes from jetson branch

TimDettmers:
Did this work out for you? It seems a straightforward fix and a good contribution if this would make the library jetson compatible.

Titus-von-Koeller:
@rickardp @younesbelkada 

Do you have opinions on this PR? Could one of you two do the review?

Makefile
```diff
@@ -41,14 +41,15 @@ CC_KEPLER += -gencode arch=compute_37,code=sm_37 # Kepler
 CC_CUDA11x := -gencode arch=compute_75,code=sm_75
 CC_CUDA11x += -gencode arch=compute_80,code=sm_80
 CC_CUDA11x += -gencode arch=compute_86,code=sm_86
-
+CC_CUDA11x += -gencode arch=compute_87,code=sm_87
```
rickardp, line 44: Can we confirm that the cmake file works with the Jetson devices? It compiles, but I do not have a device to test with.

Wheels can be taken from the latest build from here
https://github.com/TimDettmers/bitsandbytes/actions/workflows/python-package.yml

csrc/kernels.cu
```diff
@@ -2409,7 +2409,7 @@ template <int ITEMS_PER_THREAD, int SUBTILE_ROWS, int THREADS>__global__ void kd
 }
 
 
-template <int THREADS, int ITEMS_PER_THREAD, int TILE_ROWS, int TILE_COLS, int SPARSE_DECOMP> __global__ void kDoubleRowColQuant(half *__restrict__ const A, float *__restrict__ const rowStats, float * __restrict__ const colStats, char *out_col_normed, char *out_row_normed, int *rowidx, int *colidx, half *val, int * __restrict__ nnz_block_ptr, float threshold, int rows, int cols, int tiledCols)
+template <int THREADS, int ITEMS_PER_THREAD, int TILE_ROWS, int TILE_COLS, int SPARSE_DECOMP> __global__ void kDoubleRowColQuant(half *__restrict__ const A, float *__restrict__ const rowStats, float * __restrict__ const colStats, int8_t *out_col_normed, int8_t *out_row_normed, int *rowidx, int *colidx, half *val, int * __restrict__ nnz_block_ptr, float threshold, int rows, int cols, int tiledCols)
```
rickardp, line 2412: Not sure why this is needed, but as long as it compiles on all platforms (looking at you, MSVC :) ), I don't see a problem with the change either .IIRC, int8_t is exactly 8 bits, while char is *at least* 8 bits

include/SIMD.h
```diff
@@ -28,6 +28,9 @@ FORCE_INLINE int popcnt32(int x32)
 
 #if defined(USE_AVX) || defined(USE_AVX2)
 #include <immintrin.h>
+#elif defined __aarch64__
+#warning "--- THIS IS AARCH64"
+#include <sse2neon.h>
```
rickardp, line 33: We are going to need to support Neon one way or the other. I am pondering if this is the right approach though, or if we should implement the Neon intrinsics directly? If it saves us time in the short run, maybe a viable option?

setup.py
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/python3
```
rickardp, line 1: If we want this, `/usr/bin/env python3` is more portable.

Also, the file is not executable. Need to chmod 755 and commit if this is to make sense

matthewdouglas:
Hi all,
I'm going to close this as it's quite outdated and should no longer be relevant. Most of the changes here (e.g. `char` => `int8_t`) have been applied via other PRs like #1401. We're building aarch64-sbsa wheels now to be included in the next release.

As far as Jetson support is concerned, we're not yet distributing wheels for that, but it can be built. E.g. there is a distribution being published that has been mentioned here: https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1437#issuecomment-2806126680.

PR #859, opened 2023-11-08
Fix typehint error

younesbelkada:
The recent 0.41.2 version causes some errors when one tries to install bitsandbytes 

```bash
    class QuantState:
  File "/usr/local/lib/python3.8/site-packages/bitsandbytes/functional.py", line 603, in QuantState
    def from_dict(cls, qs_dict: dict[str, Any], device: torch.device) -> 'QuantState':
TypeError: 'type' object is not subscriptable
```

This PR fixes it - would be great to add this PR in a patch release if possible 🙏 

cc @TimDettmers @poedator @Titus-von-Koeller

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -13,7 +13,7 @@ from scipy.stats import norm
 import numpy as np
 
 from functools import reduce  # Required in Python 3
-from typing import Tuple, Any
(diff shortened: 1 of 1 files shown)


TimDettmers:
Thank you for being so quick on this!

PR #867, opened 2023-11-09
Avoid double-quantizing when calling `cuda()`

jph00:
Currently calling `cuda()` twice quantizes the quantized data, breaking the module. However it's necessary to call it a 2nd time after using the meta device, in order to place buffers on the device.


Titus-von-Koeller:
Thanks for your contribution, @jph00!

PR #868, opened 2023-11-09
Fix for 4bit without compress_statistics.

poedator:
addressing issues from https://github.com/huggingface/transformers/pull/26037#pullrequestreview-1722036649

the problem came from[ this line](https://github.com/TimDettmers/bitsandbytes/pull/753/files#diff-f8dbfa8ee0162e506a70aeb3be4f1c29ff19e71d331f184f2b01267de2232dc5R659). Somehow shape was not saved for non-nested setups.

The tests had fault in not passing the `compress_statistics` argument properly [here](https://github.com/TimDettmers/bitsandbytes/pull/753/files#diff-208eb304b17a40a9705d566f5ba0abf0b317266205cbcf7d146526e4d8f7a58aR37). So the issue did not come up before.

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -651,7 +651,7 @@ class QuantState:
     def as_dict(self, packed=False):
         """
         returns dict of tensors and strings to use in serialization via _save_to_state_dict()
-        param: packed -- returns dict[str, torch.Tensor] for state_dict
+        param: packed -- returns dict[str, torch.Tensor] for state_dict fit for safetensors saving
         """
         qs_dict = {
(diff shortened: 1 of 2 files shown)


poedator:
additionally I  added unsharing for `nested_quant_map` quant_state component to avoid the need to deal with restoring shared tensors after loading from safetensors format. Since these components have 256 fp16 elements each, it is not affecting the size of saved model materially.

Code is tested with bnb tests and with fresh 4bit serialization tests at transformers.

@TimDettmers, please review and consider merging.

TimDettmers:
Thank you, @poedator! This looks good and all tests are green for me.

PR #873, opened 2023-11-15
fix library detection

wkpark:
fix library detection

manually cherry-picked from PR #788

See also #876

### DIFF
--- a/bitsandbytes/__main__.py
+++ b/bitsandbytes/__main__.py
@@ -11,34 +11,18 @@ import torch
 
 HEADER_WIDTH = 60
 
-def execute_and_return(command_string: str) -> Tuple[str, str]:
-    def _decode(subprocess_err_out_tuple):
-        return tuple(
-            to_decode.decode("UTF-8").strip()
-            for to_decode in subprocess_err_out_tuple
-        )
-
-    def execute_and_return_decoded_std_streams(command_string):
-        return _decode(
-            subprocess.Popen(
-                shlex.split(command_string),
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE,
-            ).communicate()
-        )
-
-    std_out, std_err = execute_and_return_decoded_std_streams(command_string)
-    return std_out, std_err
 
 def find_file_recursive(folder, filename):
-    folder = shlex.quote(folder)
-    filename = shlex.quote(filename)
-    cmd = f'find {folder} -name {filename}'
-    out, err = execute_and_return(cmd)
-    if len(err) > 0:
-        raise RuntimeError('Something when wrong when trying to find file. Maybe you do not have a linux system?')
+    import glob
+    outs = []
+    try:
+        for ext in ["so", "dll", "dylib"]:
+            out = glob.glob(os.path.join(folder, "**", filename + ext))
+            outs.extend(out)
+    except Exception as e:
+        raise RuntimeError('Error: Something when wrong when trying to find file. {e}')
 
-    return out
+    return outs
 
 
 def generate_bug_report_information():
@@ -48,18 +32,23 @@ def generate_bug_report_information():
     print('')
 
     if 'CONDA_PREFIX' in os.environ:
-        paths = find_file_recursive(os.environ['CONDA_PREFIX'], '*cuda*so')
+        paths = find_file_recursive(os.environ['CONDA_PREFIX'], '*cuda*')
         print_header("ANACONDA CUDA PATHS")
         print(paths)
         print('')
     if isdir('/usr/local/'):
-        paths = find_file_recursive('/usr/local', '*cuda*so')
+        paths = find_file_recursive('/usr/local', '*cuda*')
         print_header("/usr/local CUDA PATHS")
+        print(paths)
+        print('')
+    if 'CUDA_PATH' in os.environ and isdir(os.environ['CUDA_PATH']):
+        paths = find_file_recursive(os.environ['CUDA_PATH'], '*cuda*')
+        print_header("CUDA PATHS")
         print(paths)
         print('')
 
     if isdir(os.getcwd()):
-        paths = find_file_recursive(os.getcwd(), '*cuda*so')
+        paths = find_file_recursive(os.getcwd(), '*cuda*')
         print_header("WORKING DIRECTORY CUDA PATHS")
         print(paths)
         print('')
--- a/bitsandbytes/__main__.py
+++ b/bitsandbytes/__main__.py
@@ -67,7 +67,7 @@ def generate_bug_report_information():
     print_header("LD_LIBRARY CUDA PATHS")
     if 'LD_LIBRARY_PATH' in os.environ:
         lib_path = os.environ['LD_LIBRARY_PATH'].strip()
-        for path in set(lib_path.split(':')):
+        for path in set(lib_path.split(os.pathsep)):
             try:
                 if isdir(path):
                     print_header(f"{path} CUDA PATHS")
--- a/bitsandbytes/__main__.py
+++ b/bitsandbytes/__main__.py
@@ -60,7 +60,7 @@ def generate_bug_report_information():
             try:
                 if isdir(path):
                     print_header(f"{path} CUDA PATHS")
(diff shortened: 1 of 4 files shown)


wkpark:
rebased. fix some mistakes. partial squashed.

bitsandbytes/__main__.py
```diff
@@ -46,26 +32,31 @@ def generate_bug_report_information():
     print('')
 
     if 'CONDA_PREFIX' in os.environ:
-        paths = find_file_recursive(os.environ['CONDA_PREFIX'], '*cuda*so')
+        paths = find_file_recursive(os.environ['CONDA_PREFIX'], '*cuda*')
```
TimDettmers, line 35: We need to make sure that this does not match non-binary files. How can we make sure that this is correct in most environments?

bitsandbytes/__main__.py
```diff
@@ -11,32 +11,18 @@
...
+    outs = []
+    try:
+        for ext in ["so", "dll", "dylib"]:
+            out = glob.glob(os.path.join(folder, "**", filename + ext))
+            outs.extend(out)
+    except Exception as e:
+        raise RuntimeError('Error: Something when wrong when trying to find file. {e}')
 
-    return out
+    return outs
```
TimDettmers, line 25: Does the list instead of path cause errors in some edge cases? As I understand it, we either have `dll` or `dylib`, is that correct? Or can we have both in some cases?

bitsandbytes/__main__.py
```diff
@@ -46,26 +32,31 @@ def generate_bug_report_information():
...
         print(paths)
         print('')
     if isdir('/usr/local/'):
-        paths = find_file_recursive('/usr/local', '*cuda*so')
+        paths = find_file_recursive('/usr/local', '*cuda*')
         print_header("/usr/local CUDA PATHS")
         print(paths)
         print('')
+    if 'CUDA_PATH' in os.environ and isdir(os.environ['CUDA_PATH']):
+        paths = find_file_recursive(os.environ['CUDA_PATH'], '*cuda*')
```
TimDettmers, line 45: Same as above. Can we ensure correctness in weird envs?

bitsandbytes/__main__.py
```diff
@@ -46,26 +32,31 @@ def generate_bug_report_information():
...
+        paths = find_file_recursive(os.getcwd(), '*cuda*')
         print_header("WORKING DIRECTORY CUDA PATHS")
         print(paths)
         print('')
 
     print_header("LD_LIBRARY CUDA PATHS")
     if 'LD_LIBRARY_PATH' in os.environ:
         lib_path = os.environ['LD_LIBRARY_PATH'].strip()
-        for path in set(lib_path.split(':')):
+        for path in set(lib_path.split(':' if not os.sep == "\\" else ";")):
```
TimDettmers, line 59: I think this is might be incorrect. What is the `os.sep` for linux? I think `:` is a custom separator that is just for linux env variables. Can you please check this?

bitsandbytes/cuda_setup/env_vars.py
```diff
@@ -26,7 +26,7 @@ def to_be_ignored(env_var: str, value: str) -> bool:
 
 
 def might_contain_a_path(candidate: str) -> bool:
-    return "/" in candidate
+    return os.sep in candidate
```
TimDettmers, line 29: @Titus-von-Koeller you worked with universal paths before. Do you see a problem with this? I think once we separate out the `:` (or other divisors) we could use the PosixPath instead of this.

bitsandbytes/cuda_setup/main.py
```diff
@@ -31,7 +32,13 @@
 # libcudart.so is missing by default for a conda install with PyTorch 2.0 and instead
 # we have libcudart.so.11.0 which causes a lot of errors before
 # not sure if libcudart.so.12.0 exists in pytorch installs, but it does not hurt
-CUDA_RUNTIME_LIBS: list = ["libcudart.so", 'libcudart.so.11.0', 'libcudart.so.12.0']
+system = platform.system()
+if system == 'Darwin':
+    CUDA_RUNTIME_LIBS: list = ["libcuda.dylib", '/usr/local/cuda/lib/libcuda.dylib']
```
TimDettmers, line 37: Can you explain to me a bit where this very fixed path on Drawin comes from?

bitsandbytes/cuda_setup/main.py
```diff
@@ -111,7 +118,9 @@ def manual_override(self):
                           'For example by adding the following to your .bashrc: export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:<path_to_cuda_dir/lib64\n'
                           f'Loading CUDA version: BNB_CUDA_VERSION={os.environ["BNB_CUDA_VERSION"]}'
                           f'\n{"="*80}\n\n'))
-                    self.binary_name = self.binary_name[:-6] + f'{os.environ["BNB_CUDA_VERSION"]}.so'
+                    binary_name = self.binary_name.rsplit(".", 1)[0]
```
TimDettmers, line 121: This is a bit of a mess (based on my messy previous code). I have not looked at the details. Is there a way where we can clean this up slightly?

TimDettmers (commented):
Thank you for this PR! Overall it looks very good. There are some issues that I would like more information on but I think we would be able to merge this without any major issues.

bitsandbytes/__main__.py
wkpark: `dll`, `so`, or `dylib`  ext part appended later in the `find_file_recursive()`

bitsandbytes/__main__.py
wkpark: > Same as above. Can we ensure correctness in weird envs?

currently, `find_file_recursive()` is used for debugging purposes only (used at `generate_bug_report_information()`).
so I think it is enough.

bitsandbytes/__main__.py
wkpark: now fixed to use `os.pathsep` https://docs.python.org/3/library/os.html#os.pathsep

bitsandbytes/cuda_setup/main.py
wkpark: oops. I guess this line is an outdated hunk from some older cuda detection routine and not cleaned up before posting this PR.

wkpark:
rebased with some fixes.

wkpark:
rebased, partial squashed, conflict resolved

bitsandbytes/cuda_setup/env_vars.py
Titus-von-Koeller: this looks good to me

bitsandbytes/__main__.py
Titus-von-Koeller: answer makes sense to me

Titus-von-Koeller:
@wkpark Thanks a lot for the nice PR and your continuing great work around bnb, despite frustrations! The code looks good from a preliminary review. Unfortunately, this part of the code doesn't have an easy way to test and I currently don't have too much time to look very deeply into this.

This code has been kind of flaky in the past. I'm a bit worried about releasing it without thorough testing. How risky do you see this merge and how could we make sure that things work out smoothly?

Did you validate on Linux as well? Another target to check is Collab. I wish we had a better strategy for testing this module: Any ideas? If we were less worried about breaking things, we could follow up with further refactorings, [which would be needed](https://github.com/TimDettmers/bitsandbytes/issues/894).

bitsandbytes/__main__.py
```diff
@@ -11,34 +11,18 @@
...
-    if len(err) > 0:
-        raise RuntimeError('Something when wrong when trying to find file. Maybe you do not have a linux system?')
+    import glob
+    outs = []
+    try:
+        for ext in ["so", "dll", "dylib"]:
+            out = glob.glob(os.path.join(folder, "**", filename + ext))
+            outs.extend(out)
+    except Exception as e:
+        raise RuntimeError('Error: Something when wrong when trying to find file. {e}')
```
akx, line 23: This is missing an `f` so the exception is never printed...

PR #895, opened 2023-12-03
Update cuda_install.sh

theamdara:
Add option to download and install Cuda 12.3


Titus-von-Koeller:
Thanks for your contribution!

PR #896, opened 2023-12-03
chore: update dev setup

Titus-von-Koeller:


akx:
This doesn't address the 132 Ruff errors found, though? 😄 

Also, `fix = true` is somewhat iffy as a default, what's the reasoning there?

PR #913, opened 2023-12-13
Add sm_90a to enable use of accelerated wgmma and setmaxnreg instructions

ConsceIeratus:
sm_90a adds support for accelerated wgmma and setmaxnreg instructions.

https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#ptx-module-directives-target

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ CC_cublasLt111 += -gencode arch=compute_80,code=sm_80
 CC_cublasLt111 += -gencode arch=compute_86,code=sm_86
 
 CC_ADA_HOPPER := -gencode arch=compute_89,code=sm_89
-CC_ADA_HOPPER += -gencode arch=compute_90,code=sm_90
+CC_ADA_HOPPER += -gencode arch=compute_90a,code=sm_90a
 
 
 all: $(BUILD_DIR) env


TimDettmers:
Thank you, this is good to know that wgmma is now added. I think Hopper supports both sm_90 and sm_90a. Since we do not make use of wgmma or setmaxnreg for now we would not need sm_90a. I would want to not add it at the moment to keep the binary a bit smaller. I am currently having troubles with the binary size since all binaries must be smaller than 100MB for PyPi uploads.

Titus-von-Koeller:
Thanks for raising this. We'll keep this in mind and implement it once we have figured out the current cross-platform + build + distribution topics.

akx:
For PyPI, you can request a quota increase for the project via https://github.com/pypi/support :)

younesbelkada:
Nice - @akx would you mind sharing that in https://github.com/TimDettmers/bitsandbytes/discussions/990 as well 🙏

Titus-von-Koeller:
> For PyPI, you can request a quota increase for the project via https://github.com/pypi/support :)

Nice, that would unblock us on this before we've figured out the cross-platform compilation and distribution stuff.

I'll look into that today. Valuable input, thanks!

matthewdouglas:
Hi all,

I am going to close this as it has been stale for a long time now. We've moved on to use CMake over the `Makefile`. We're also not making direct use of any `sm_90a` specific features. 

In the future we may consider shipping builds with Hopper/Blackwell architecture-specific (or the new CUDA 12.9 family-specific) targets as needed. Until then, we target `sm_90` and `compute_90` on CUDA < 12.8. For CUDA 12.8 we target `sm_90`, `sm_100`, `sm_120`, and `compute_120`.

PR #922, opened 2023-12-17
Move to CPU before attempting to convert to numpy.

iantbutler01:
I ran into this when loading from 4bit in Transformers while everything was loaded in a `with torch.device("cuda"):` context. I know that functionality isn't fully merged yet in Transformers but I had been using it last night in a separate fork after rebasing and installing the work from https://github.com/huggingface/transformers/pull/26037.

I ran the test suite and there are a few (4) failures that I'm not certain would be caused by my code such as an issue with the triton jit.

### DIFF
--- a/bitsandbytes/utils.py
+++ b/bitsandbytes/utils.py
@@ -187,7 +187,7 @@ def unpack_tensor_to_dict(tensor_data):
     Returns:
     A Python dictionary containing the unpacked data.
     """
-    json_bytes = bytes(tensor_data.numpy())
+    json_bytes = bytes(tensor_data.cpu().numpy())
     json_str = json_bytes.decode('utf-8')
     unpacked_dict = json.loads(json_str)
 


TimDettmers:
Thank you for the PR! @poedator can you confirm with these changes that the resulting tensor is on the correct device? @younesbelkada for the tests to trigger, currently we would need to merge this, correct?

(@younesbelkada can you add me to the slack channel for the test runner)

poedator:
the proposed change looks safe to me. however I am unable to test it in the next few days.

poedator:
In my tests there is no need for such additional transformation - when non-tensor data is packed, it is saved and loaded as cpu tensor. 

However if there are cases when people use `with torch.device("cuda"):` context, this may indeed create issues. In that case the suggested change from @iantbutler01  should help.
@TimDettmers

younesbelkada:
Hi !
@TimDettmers I just added you on that slack channel, I propose we merge this PR as it is quite tiny, and we let it live on main for a day and see if all test pass! wdyt @Titus-von-Koeller ?

Titus-von-Koeller:
@younesbelkada yeah, I think that's a good approach. I ll merge to main and then we see if all HuggingFace integration tests pass.

Thanks @iantbutler01 for your contribution!

PR #925, opened 2023-12-18
[`Core` / `GitHub`] Add stale bot

younesbelkada:
# What does this PR do?

As per title, this PR adds a stale bot to the library in order to filter out old issues

The scripts and workflow are copied and adapted from : https://github.com/huggingface/trl/blob/main/scripts/stale.py

Read more about the `GITHUB_TOKEN` secret here: https://docs.github.com/en/actions/security-guides/automatic-token-authentication

cc @Titus-von-Koeller @TimDettmers

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,27 @@
+name: Stale Bot
+
+on:
+  schedule:
+    - cron: "0 15 * * *"
+
+jobs:
+  close_stale_issues:
+    name: Close Stale Issues
+    if: github.repository == 'TimDettmers/bitsandbytes'
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Setup Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.8
+
+    - name: Install requirements
+      run: |
+        pip install PyGithub
+    - name: Close stale issues
+      run: |
+        python scripts/stale.py
--- a/scripts/stale.py
+++ b/scripts/stale.py
@@ -1,0 +1,60 @@
+# Copyright 2023 The HuggingFace Team, the AllenNLP library authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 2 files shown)


Titus-von-Koeller:
LGTM, thanks @younesbelkada, great input 🤗🤘🏼

tomaarsen:
Heya you two!

I recognize that this stale bot has been removed now, but I am curious in it for my own purposes - this implementation feels a bit "manual", i.e. a normal cron script that runs a manually included Python script to mark as stale, close issues, etc.

On the other hand, there is also the [stale](https://github.com/actions/stale) GitHub action. Do you know why this manual version was chosen rather than the GitHub action? I was planning on using the GitHub action, but if there's something lacking/suboptimal in that one, then I'd love to know about it before I affect hundreds of issues 😄 

- Tom Aarsen

Titus-von-Koeller:
Hey @tomaarsen,

Great to hear from you! Maybe we should have a coffee sometime, as we seem to be struggling with similar things :)

I was hoping @younesbelkada would answer because he was the one that brought forward the implementation as it currently is. My understanding is that we took the best practice from Transformers and just applied it without changes here.

As to whether this is better over the default Github action, I'm not sure. But looking at the link you provided, it seems to me like it offers a bunch more options to fine-tune what's happening. Maybe we should also switch over to this for `bitsandbytes`. I would agree that for you the officially supported GitHub action is likely a better way forward.

For now it solved our immediate issue of needing to draw a line somewhere when trying to get on top of overdue maintenance + prioritize. It's still pretty hard the way it currently is, even without looking at the old stuff and hoping that someone reopens, if it's still important.

tomaarsen:
Haha that sounds great 😄 

I especially like the idea of the GitHub one as you can seemingly limit the [number of actions](https://github.com/actions/stale#operations-per-run) and you can specify how frequently it must run. That means, I should be able to configure it to stale only ~10 issues per day, only on weekdays, and then I can manually look at the staled issues myself to potentially un-stale them.

I'll have a look at this for [Sentence Transformers](https://github.com/UKPLab/sentence-transformers) in the coming days most likely, and I'll let you know how it goes :)

Titus-von-Koeller:
Yeah, that sounds much more configurable. Potentially very useful. Let us
know how it goes :)

PR #257, opened 2023-04-03
Make bitsandbytes portable (with express goal of supporting Apple Silicon/ARM64)

rickardp:
See issue #252 for discussion

- [x] C++ compiles on ARM64 in non-SIMD mode
- [ ] C++ compiles on ARM64 with SIMD (Neon) instructions
- [x] Do not fail unit tests if no CUDA device present
- [ ] Make Python library portable to non-CUDA
- [x] Unit tests run on devices present
- [x] Make unit tests work regardless of CUDA present
- [x] GitHub workflow to build wheels for all supported platforms (I've done this in https://github.com/rickardp/splitstream/blob/master/.github/workflows/python-package.yml, some inspiration will be taken from there although this is using a separately built shared lib and will be done differently)
- [x] Replace use of pthread with canonical C++ std::thread (while also avoiding arcane C `malloc` and using type-safe C++ primitives) 
- [x] Switch to CMake to reliably build on all platforms.

image

PR #947, opened 2024-01-02
[do not merge] Old portability collection PR, for reference

rickardp:
This is essentially the same as PR #257. As the old PR was force pushed beyond repair, I'm reposting this for the purpose of referencing the diffs and possibly cherry picking useful parts into other PRs.

The PR sets up a pipeline that builds the .so files separately using a build matrix, then builds binary wheels for the Python packages. The MPS stuff here is just a stub and should not be merged.

AFAIK, the CMake files in this branch contain all the releevant patches to make the CPU code buildable/testable on Windows x64, Mac x64/arm64, Linux x64/arm64.

To summarize this PR should provide

- A CMake file that compiles "eveywhere" and builds one target at a time
- A GitHub workflow that builds first the native libs using a build matrix, then puts everything together in a bdist_wheel
- Source level changes to make native code compile on Windows, Mac and arm64
- MPS stub code for future reference
- Python code refactoring. Probably not worth keeping


Titus-von-Koeller:
Ok, we'll keep it here for reference. Please close it once not needed anymore. We'll approach you soon about cross-platform support. Thanks for all the good work so far. Let's work together to get this figured out. We're still a bit behind on getting on top of maintenance since we took over, so please excuse the delays. Thanks for your understanding!

Titus-von-Koeller:
Closing this as it seems dead and not needed anymore.

Let me know if it needs reopening.

PR #948, opened 2024-01-02
Don't require scipy for regular use

akx:
Refs #488.

The only use for `scipy.stats.norm` is `create_normal_map`, so it probably does not need to be a hard requirement for any and all users of the library.

Many users will of course have it installed anyway.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,11 +26,13 @@ setup(
     license="MIT",
     keywords="gpu optimizers optimization 8-bit quantization compression",
     url="https://github.com/TimDettmers/bitsandbytes",
-    install_requires=['scipy'],
     packages=find_packages(),
     package_data={"": libs},
-    install_requires=['torch', 'numpy', 'scipy'],
-    extras_require={'benchmark': ['pandas', 'matplotlib']},
+    install_requires=['torch', 'numpy'],
+    extras_require={
+        'benchmark': ['pandas', 'matplotlib'],
+        'test': ['scipy'],
+    },
     long_description=read("README.md"),
     long_description_content_type="text/markdown",
     classifiers=[
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -233,8 +233,15 @@ def create_linear_map(signed=True, total_bits=8, add_zero=True):
         l = values.numel()//2
         return torch.Tensor(values[:l].tolist() + [0]*gap + values[l:].tolist())
 
+
 def create_normal_map(offset=0.9677083, use_extra_value=True):
-    from scipy.stats import norm
+    try:
+        from scipy.stats import norm
+    except ImportError as ie:
+        raise ImportError(
+            "Scipy is required for `create_normal_map`. "
+            "Install `bitsandbytes` with the `[test]` extra."
+        ) from ie
 
     if use_extra_value:
         # one more positive value, this is an asymmetric type


Titus-von-Koeller:
Thanks for raising this, makes sense! I need to discuss this with the other maintainers first. I put it on my list. Will get back to you.

akx:
Rebased post #488.

setup.py
```diff
@@ -26,11 +26,13 @@ def read(fname):
...
     url="https://github.com/TimDettmers/bitsandbytes",
-    install_requires=['scipy'],
     packages=find_packages(),
     package_data={"": libs},
-    install_requires=['torch', 'numpy', 'scipy'],
-    extras_require={'benchmark': ['pandas', 'matplotlib']},
+    install_requires=['torch', 'numpy'],
+    extras_require={
+        'benchmark': ['pandas', 'matplotlib'],
+        'test': ['scipy'],
```
rickardp, line 34: Is `create_normal_map` only used for testing? Otherwise the name might be confusing?

setup.py
```diff
@@ -26,11 +26,13 @@ def read(fname):
...
     keywords="gpu optimizers optimization 8-bit quantization compression",
     url="https://github.com/TimDettmers/bitsandbytes",
-    install_requires=['scipy'],
     packages=find_packages(),
     package_data={"": libs},
-    install_requires=['torch', 'numpy', 'scipy'],
-    extras_require={'benchmark': ['pandas', 'matplotlib']},
+    install_requires=['torch', 'numpy'],
+    extras_require={
+        'benchmark': ['pandas', 'matplotlib'],
```
rickardp, line 33: Are these runtime dependencies at all? The seem to be used only in scripts and are not part of the wheel? Maybe the `benchmark` extra is not needed at all?

setup.py
akx: The only call to the function is in a test (and it's mentioned in a couple copy-pasted comments too):

```
(bitsandbytes) ~/b/bitsandbytes (main) $ ag create_normal_map
tests/test_functional.py
2356:    code = F.create_normal_map()

bitsandbytes/nn/modules.py
306:        Implementation of the NF4 data type in bitsandbytes can be found in the `create_normal_map` function in

bitsandbytes/functional.py
236:def create_normal_map(offset=0.9677083, use_extra_value=True):
847:            Implementation of the NF4 data type in bitsandbytes can be found in the `create_normal_map` function ```

I can't find other relevant references to it [across all of GitHub either](https://github.com/search?q=create_normal_map+language%3APython&type=code&l=Python) – and anyway, the function is still there for downstream users who might need it, they'll just need `scipy` installed.

File: setup.py (hunk 2 above)
Comment (akx) [reply setup.py:33]: The idea is that when running the benchmark scripts, you'd do 

```
pip install -e .[benchmark]
```
to get the regular requirements plus the things required for benchmarking, or if you want that and the test deps,
```
pip install -e .[benchmark,test]
```

rickardp: Doesn’t this also make it part of the public metadata, e.g “pip install bitsandbytes[benchmark]” ?

akx: Yes, it does.

Titus-von-Koeller:
Checked with Tim. He agrees that this makes sense and sees no issue with it.

Thanks for raising this and taking the initiative to provide a fix!

PR #908, opened 2023-12-11
Cmake + workflows

wkpark:
CmakeList.txt + github workflows action

* CMake by @Jamezo97
* github workflows actions - support ubuntu and windows (windows disabled for now)
* cu11.8, cu12.1 + py310, py311 build matrix

Please see ~~https://github.com/wkpark/bitsandbytes/actions/runs/7166014267~~
~~https://github.com/wkpark/bitsandbytes/actions/runs/7167362163~~
https://github.com/wkpark/bitsandbytes/actions/runs/7648577944

PR #949, opened 2024-01-02
Make native code portable and add GitHub workflow for building

rickardp:
This PR fixes

- Adds the CMake support based on the original contribution by @niclimcy ( + Apple build targets)
- Adds GitHub workflow to build first native libs using a build matrix. Builds CUDA for Linux arm64/x64 + Windows x64, CPU for Windows, Linux and Mac. Python builds bdist_wheels for matrix of Python versions (packages native libs built once).
- Adds all known source code patches for making the code build on Windows, Mac and Linux/arm64

It does not add #847, which should be considered

See RFC #1020 for discussion


niclimcy:
I would split github workflow into one commit, cmakelist into another, and platform portability changes into one more commit

rickardp:
> I would split github workflow into one commit, cmakelist into another, and platform portability changes into one more commit

I guess the source code changes could be broken out, but there would be no way of verifying them without a working build system.

Same goes for GitHub actions. Without actually being able to build for multiple arch's/OSs it would only fail. And conversely - without GH actions there is no way to verify the code working unless you happen to own all hardware.

Titus-von-Koeller:
Hey both!

Thanks for your contribution. You're right CMake might definitely make sense to move to, especially given our current effort to support cross-platform. This needs a bit more thought and input from Tim, which we should get in the coming weeks. Unfortunately, Tim's availability is very limited right now. He mentioned concerns about migrating away from Make: I'll have to find out more and will get back to you!

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+      uses: microsoft/setup-msbuild@v1.1
+      if: ${{ startsWith(matrix.os, 'windows') }}
+      # Compile C++ code
+    - name: Build C++
+      shell: bash
+      run: |
+        set -ex
+        build_os=${{ matrix.os }}
+        build_arch=${{ matrix.arch }}
+        ( git clone https://github.com/NVlabs/cub ./dependencies/cub; cd dependencies/cub; git checkout 1.11.0 )
```
akx, line 110: Ditto here re release tarball. Maybe this build script should be in e.g. `.github/scripts/build.sh`?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+        path: output/*
+        retention-days: 7
+  build-wheels:
+    needs:
+    - build-shared-libs
+    - build-shared-libs-cuda
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
akx, line 138: 3.13 isn't released yet, I think this would require [`allow-prereleases`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases)?

That said, building the wheels only needs to be done on the latest published Python, so

```suggestion
        python-version: ["3.x"]
```

ought to be enough 😄

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+        pip install -r requirements.txt
+    # TODO: How to run CUDA tests on GitHub actions?
+    #- name: Run unit tests
+    #  if: ${{ matrix.arch == 'x86_64' }} # Tests are too slow to run in emulation. Wait for real aarch64 agents
+    #  run: |
+    #    PYTHONPATH=. pytest --log-cli-level=DEBUG tests
+    - name: Build wheel
+      shell: bash
+      run: |
+        python setup.py bdist_wheel
```
akx, line 175: One shouldn't run `setup.py` directly anymore, but use [`build`](https://pypi.org/project/build):
```suggestion
    - name: Build wheel
      shell: bash
      run: |
        pip install build
        python -m build .
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+        path: dist/bitsandbytes-*.whl
+        retention-days: 7
+  publish:
+    needs: build-wheels
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build dist
+      run: |
+        python setup.py sdist
```
akx, line 189: Ditto re `setup.py`.
```suggestion
    - name: Build dist
      run: |
        pip install build
        python -m build -s .
```
Whether or not an sdist is actually required at all is another question :)

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        arch: [x86_64, aarch64]
+        exclude:
+          - os: windows-latest # This probably requres arm64 Windows agents
+            arch: aarch64
+    runs-on: ${{ matrix.os }} # One day, we could run them on native agents. Azure supports this now but it's planned only for Q3 2023 for hosted agents
+    steps:
+      # Check out code
+    - uses: actions/checkout@v3
```
akx, line 27:
```suggestion
    - uses: actions/checkout@v4
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+        os: [ubuntu-latest, windows-latest]
+        arch: [x86_64, aarch64]
+        cuda_version: ['12.1.0']
+        exclude:
+          - os: windows-latest # This probably requres arm64 Windows agents
+            arch: aarch64
+    runs-on: ${{ matrix.os }} # One day, we could run them on native agents. Azure supports this now but it's planned only for Q3 2023 for hosted agents
+    steps:
+      # Check out code
+    - uses: actions/checkout@v3
```
akx, line 81:
```suggestion
    - uses: actions/checkout@v4
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+    - name: Set up Docker multiarch
+      if: startsWith(matrix.os, 'ubuntu')
+      uses: docker/setup-qemu-action@v2
+      # On Linux we use CMake within Docker
+    - name: Setup cmake
+      if: ${{ !startsWith(matrix.os, 'linux') }}
+      uses: jwlawson/actions-setup-cmake@v1.13
+      with:
+        cmake-version: '3.26.x'
+      # Windows: We install Cuda on the agent (slow)
```
akx, line 92: If this is slow, can it be cached using e.g. `actions/cache`?

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+        arch: [x86_64, aarch64]
+        exclude:
+          - os: windows-latest # This probably requres arm64 Windows agents
+            arch: aarch64
+    runs-on: ${{ matrix.os }}
+    steps:
+      # Check out code
+    - uses: actions/checkout@v3
```
akx, line 146:
```suggestion
    - uses: actions/checkout@v4
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+      with:
+        name: shared_library
+        path: output/
+    - name: Copy correct platform shared library
+      shell: bash
+      run: |
+        cp output/${{ matrix.os }}/${{ matrix.arch }}/* bitsandbytes/
+      # Compile C++ code
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
```
akx, line 159:
```suggestion
      uses: actions/setup-python@v5
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: bdist_wheel
+        path: dist/bitsandbytes-*.whl
+        retention-days: 7
+  publish:
+    needs: build-wheels
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
akx, line 186:
```suggestion
    - uses: actions/checkout@v4
```

setup.py
```diff
@@ -6,16 +6,21 @@
...
 print("libs:", libs)
 
 
 def read(fname):
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
+# Tested with wheel v0.29.0
+class BinaryDistribution(Distribution):
+    def has_ext_modules(foo):
+        return True
```
akx, line 23: Should probably look into using [Hatchling](https://hatch.pypa.io/latest/) instead of any more `setup.py` shenanigans, but that's for another PR (and I'll gladly help out with that!)

requirements.txt
```diff
@@ -1,3 +1,5 @@
-lion-pytorch
-pytest
-scipy
+pytest==7.2.2
+einops==0.6.0
+wheel==0.40.0
+lion-pytorch==0.0.6
+scipy==1.11.4
```
akx, line 5: Please loosen these up some – chances are an exact version will not necessarily be available, especially not as a wheel, for all platforms + Python versions. [The `~=` operator](https://stackoverflow.com/a/39590286/51685) should be good here. (And let's sort them for good measure...)

```suggestion
einops~=0.6.0
lion-pytorch~=0.0.6
pytest~=7.2.2
scipy~=1.11.4
wheel~=0.40.0
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+      uses: microsoft/setup-msbuild@v1.1
+      if: ${{ startsWith(matrix.os, 'windows') }}
+      # Compile C++ code
+    - name: Build C++
+      shell: bash
+      run: |
+        set -ex
+        build_os=${{ matrix.os }}
+        build_arch=${{ matrix.arch }}
+        ( git clone https://github.com/NVlabs/cub ./dependencies/cub; cd dependencies/cub; git checkout 1.11.0 )
```
akx, line 43: Might be worth it to use `actions/checkout` for this too? Or at least, have this do a shallow clone, or better yet, unless a real git checkout is needed, just use the release tag tarball https://github.com/NVIDIA/cub/archive/refs/tags/1.11.0.tar.gz

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
+name: Python package
+
+on:
+  push:
+    branches: [ "*" ]
```
akx, line 5: Unnecessary filter 😁 
```suggestion
```

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,201 @@
...
+  pull_request:
+    branches: [ master ]
```
akx, line 7: The main branch is `main` here:
```suggestion
    branches: [ main ]
```

.github/workflows/python-package.yml
rickardp: My reasoning was that we wanted to test on all supported versions. But maybe this is overkill?

akx: Testing the built wheels could occur on all supported versions (in a separate step), I think? 😄

rickardp: If that is what is preferred, that would work. Just be aware that the actual build time is probably negligible compared to testing and spinning up an agent

akx: Just for the record, I'm not a maintainer here, just a contributor, so I don't know what the upstream preferences are :)

Maybe the steps that compile the native libraries should also do the wheel-wrapping?

.github/workflows/python-package.yml
rickardp: yes you are right. Old habits 😂

.github/workflows/python-package.yml
rickardp: Yes, got that. I think @Titus-von-Koeller assugigned himself as a reviewer so hopefully he will pitch in with any opinions :)

.github/workflows/python-package.yml
rickardp: I think the action actually uses caching. Problem is, downloading multiple GBs even from the cache is slow, then it has to be installed. Now it was some time since I played around with this, but IIRC cache did help slightly but not by orders of magnitude

setup.py
rickardp: I must admit I have not changed the way I do Python wheels much in many years. This is probably very old fashioned. I would suggest following up with a modernization PR once it has been agreed if we go with cmake and what the output artifacts are

requirements.txt
rickardp: Note that these are build time dependencies, not run time dependencies. I would actually argue that pinning dependencies hard is a good thing while building. I view the requirements txt file as the lock file other package managers use. This is the only way to get reproducible builds AFAIK. Or is there a better way?

akx: These are really test-time dependencies, so pinning them to an exact version may cause any developer's `pip install -r requirements.txt` to fail in creative ways.

Dependencies required to build a [PEP517](https://peps.python.org/pep-0517/) package should be declared according to [PEP518](https://peps.python.org/pep-0518/), i.e. in `pyproject.toml`, and this package already does so :) IOW, `pip install -r requirements.txt` should not be necessary to build this package if it's built with `python -m build .` (as recommended elsewhere in my comments).

We're not guaranteed bit-for-bit reproducible builds anyway since the GHA build infrastructure isn't immutable...

.github/workflows/python-package.yml
akx: Ah, yep, seems so: https://github.com/Jimver/cuda-toolkit/blob/dc0ca7bb29c5a92f7a963d3d5c93f8d59765136a/action.yml#L25-L28

.github/workflows/python-package.yml
```diff
@@ -0,0 +1,203 @@
...
+      # Compile C++ code
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+      # 
+    - name: Install Python dependencies
+      shell: bash
+      run: |
+        pip install -r requirements.txt
```
akx, line 166: As discussed in the `requirements.txt` comment, this shouldn't be necessary at all for building :)
```suggestion
```

.github/workflows/python-package.yml
rickardp: Agreed, but missing a {} in this case. I can fix later

requirements.txt
rickardp: Not bit for bit, but isn't it considered good practice when I make a PR to build/test on the exact same version of dependencies as the target branch? I don't know if Python is any different here, but the normal workflow I've grown accustomed to is that a CI system should not arbitrarily upgrade packages from what the developer uses.

And I don't understand why the pip install would fail if we test on all supported platforms? Isn't it the other way around ,that an arbitrary patch upgrade might for some obscure reason break tests in an unrelated PR?

Correct me if I am wrong but I think PEP518 is concerned with the build system itself, not runtime dependencies for running the tests.

Is it considered bad practice in Python to pin versions in requirements.txt? I always learned to pip freeze > requirement.txt when working with CI.

.github/workflows/python-package.yml
rickardp: Fixed

.github/workflows/python-package.yml
rickardp: For now, I removed Python 3.13 until it is released. I think that makes sense in any case

requirements.txt
akx: >  isn't it considered good practice when I make a PR / is it considered bad practice in Python to pin versions in requirements.txt

IMO, the answer is "it depends". The libraries mentioned here should be prominent and well-behaved enough that they follow semver and minor/patch version increases should not cause issues.

> And I don't understand why the pip install would fail if we test on all supported platforms? 

The issue is that `requirements.txt` here is not only for CI – running tests locally requires these dependencies to be installed, and as time goes by and developers e.g. change platforms, upgrade Pythons, etc., a too strictly specified (then old) version will not have wheels/compatibility with that platform, leading to another kind of cryptic installation failure.

> Correct me if I am wrong but I think PEP518 is concerned with the build system itself

Yep! If the workflow is refactored (as discussed in the other comments, I think) to build a wheel once, these wouldn't be required for the build wheel step. For tests, I'd probably prefer `pip install -e .[test]` or similar, i.e. that the test dependencies are declared in an `extra`.

> I always learned to `pip freeze > requirement.txt` when working with CI.

That's not quite foolproof either. If a package depends on e.g. `pywin32` when running on Windows, this would create a requirements file that will absolutely always fail to install on any non-Windows platform.

rickardp: > should be prominent and well-behaved enough that they follow semver

IMHO this is not enough. Upgrading even patch versions should be a manual step. Hopefully automated, with Dependabot, but still something that will result in a code change. This is what lockfiles are for, and I understand that the concept in Pip currently is not perfect.

> The issue is that requirements.txt here is not only for CI – running tests locally requires these dependencies to be installed, and as time goes by and developers e.g. change platforms

Why not just update the dependencies in this case? Wouldn't you want this to be the new baseline for tests in this case?

I'm happy to change in whichever way the maintainers see suitable but I do want an argument for pinning versions in CI builds.

rickardp: Yes, that does sound like a good idea. This would be more like the way Yarn or Cargo does it.

rickardp: @akx I tried to read up some around the best practices on how to build Python packages today, and made an attempt to move in that direction. You should now have commit rights to this branch, so feel free to fix anything you
think can be done better.

Key points:
* Use the `build` module to build
* Move dev time dependencies to requirements-dev.txt and CI dependencies to requirements-ci.txt (the difference is that the latter have pinned versions). I verified that Dependabot handles these two files as expected. This can be automated better by using something like `poetry` but I think that needs to be discussed a bit first.
* I did not yet move the wheel building, but I agree in principle with your suggestion to build the wheels in one step, I just couldn’t figure out quickly how to do the platform specific wheels in a standard Linux runner. Feel free to fix if you want :)

younesbelkada (commented):
This is great - thanks a lot @rickardp 
I just have a question about how this PR interacts with https://github.com/TimDettmers/bitsandbytes/pull/908 do you know if there is any duplicate with that PR?

akx:
@younesbelkada Yes, this PR conflicts with #908; both do similar things (but this is one is somewhat broader in scope).

younesbelkada:
Perfect, would it make sense to first merge #908 then rebase this branch with main to add Apple build targets as well ?

akx:
Well, both orders-of-merging will mean some extra work to rebase the other PR anyway...

wkpark:
I've tested this PR under M1 with the latest main branch (cherry-pick minimal fixes + misc osx fix)

how to test mps device?

this is my try
~~~bash
% python3 -m pytest tests/test_functional.py
==================================================================================== test session starts ====================================================================================
platform darwin -- Python 3.11.5, pytest-8.0.0, pluggy-1.4.0
rootdir: /Volumes/External/src/bitsandbytes
configfile: pytest.ini
collected 632 items                                                                                                                                                                         

tests/test_functional.py::test_estimate_quantiles[float] SKIPPED (Torch not compiled with CUDA enabled)                                                                               [  0%]
tests/test_functional.py::test_estimate_quantiles[half] SKIPPED (Torch not compiled with CUDA enabled)                                                                                [  0%]
tests/test_functional.py::test_quantile_quantization SKIPPED (Torch not compiled with CUDA enabled)                                                                                   [  0%]
tests/test_functional.py::test_dynamic_quantization SKIPPED (Torch not compiled with CUDA enabled

(snip)..
===================================================================================== warnings summary ======================================================================================
bitsandbytes/cextension.py:34
  /Volumes/External/src/bitsandbytes/bitsandbytes/cextension.py:34: UserWarning: The installed version of bitsandbytes was compiled without GPU support. 8-bit optimizers, 8-bit multiplication, and GPU quantization are unavailable.
    warn("The installed version of bitsandbytes was compiled without GPU support. "

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================================================================== PASSES ===========================================================================================
_________________________________________________________________________________ test_blockwise_cpu_large __________________________________________________________________________________
----------------------------------------------------------------------------------- Captured stdout call ------------------------------------------------------------------------------------
0.013782978057861328
0.008883953094482422
0.0051801204681396484
0.0054590702056884766
0.008730173110961914
0.008594989776611328
0.004836082458496094
0.004832029342651367
======================================================================== 75 passed, 557 skipped, 1 warning in 8.08s =========================================================================
~~~

rickardp:
> This is great - thanks a lot @rickardp I just have a question about how this PR interacts with #908 do you know if there is any duplicate with that PR?

I don't see a problem rebasing the CMake parts. It is not super hard to adapt one to the other. The main thing that this PR adds to CMakeLists is the MacOS stuff. Otherwise it's mainly naming of options and some details, like how MSVC is configured. I don't think that those changes are significant.

Though as we add platforms I'd problably suggest going with a GtiHub pipeline more similar to what is in this PR. Basically I'd like to see that bulding the native code and the Python code happens separately. This will be even more important as we add more backends and CUDA versions. If anything, I'd split it up even more by testing separately like @akx  suggested in the comments to this PR.

rickardp:
> I've tested this PR under M1 with the latest main branch (cherry-pick minimal fixes + misc osx fix)
> 
> how to test mps device?
> ```

Note that the actual running of MPS tests is not part of this PR.  I ripped apart the original PR into what I believe are the mergeable "atoms". This one solves portability and pipelines by adding all device targets and compilation + GitHub pipelines. The actual running of MPS code is still in the old "reference" PR. My hope is that we can solve these bits first to make it easier to parallelize the efforts. So, in short, there will be a macos wheel but it will run the same code as the linux CPU-only wheel at the moment

.github/workflows/python-package.yml
rickardp: Replaced with requirements-ci.txt as discussed

requirements.txt
rickardp: I settled with a "manually" managed requirements-ci.txt and one for local dev with floating versions. These will be handled correctly by dependabot, so as long as they aren't incorrectly updated this will work and they will be kept in sync. For the future, I believe a more advanced build system like Poetry would be useful since it can manage all of this with just one source of truth and one lockfile. But I think this can come in a  subsequent PR

Titus-von-Koeller: I could just skim your conversation, but I generally really like working with lockfiles and explicitly specified dev dependencies.

Conda-lock lockfiles btw are multi-platform now. Anyways, the issue with conda is that the pip only dependencies don't mix well, right? Or am I wrong in that assumption?

I'm all in for implementing modern packaging practices in this repo. Looking forward to that.

younesbelkada:
Thanks very much @rickardp for your great work! 
I think we'll first merge #908 so that we'll be able to test everything step by step - would you be happy to rebase on main once #908 gets merged and update the main readme + the docs with instructions on how to build bnb with MacOS ? 
You can start filling the instructions here: https://github.com/TimDettmers/bitsandbytes/blob/main/docs/source/installation.mdx
Thanks again !

wkpark:
based on your fantastic work, I can manage to work with recent changes with minimal required fixes.
https://github.com/wkpark/bitsandbytes/actions/runs/7736718847

- `std::thread` fix not included.
- `quantize_block_args` fix not included.
- cleanup trailing white spaces. `csrc/mps_kernels.metal`

See also:
- https://github.com/wkpark/bitsandbytes/commit/e7b5b3f30b1d1292a94762a5133cfb705b6670ae (cmake fix)
- https://github.com/wkpark/bitsandbytes/commit/6801d851e059c810819003e5fc6ea007ce3dbcb2 (workflows fix)
- https://github.com/wkpark/bitsandbytes/commits/cmake-osx-rebase-minimal/

If you don't mind you can use it as you wish.
Thank you for your hard work!

requirements.txt
akx: I think we shouldn't mix anything conda-related in here, simply because we don't need to and it's an entirely different ecosystem. But for a library, the dependencies declared in the package (i.e. not `requirements.txt`) **must** be loosely pinned to avoid conflicts. But we'll get there when we get to looking at packaging...

akx:
@rickardp Let me know if you need a hand with the rebase now that #908 was merged 😄

younesbelkada (commented):
Hi ! 
I just realised that now on the build cmake workflow we upload some artifacts: https://github.com/TimDettmers/bitsandbytes/blob/3a630c589dd4ab779d28067b17c87712c9f1a674/.github/workflows/cmake.yml#L168 did they get uploaded on the main repo? (I guess I'll just have to wait to see, but just double checking)
If that's the case, it would make the repo size quite heavy no? Or does the artifacts gets uploaded outside the main repository?

younesbelkada:
Ok false alarm, they seem to be uploaded outside the repo - e.g.: https://github.com/TimDettmers/bitsandbytes/actions/runs/7737622243/job/21096916533#step:19:80

younesbelkada:
Awesome, thanks very much for explaining !

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+    if(APPLE)
+        message(FATAL_ERROR "CUDA is not supported on macOS" )
+    endif()
+    option(NO_CUBLASLT "Don't use CUBLAST" OFF)
+    if(NO_CUBLASLT)
+        set(CMAKE_CUDA_ARCHITECTURES 50 52 60 61 70 72)
+    else()
+        set(CMAKE_CUDA_ARCHITECTURES 75 80 86 89 90)
+    endif()
+endif()
```
Neumann-A, line 16: `CMAKE_CUDA_ARCHITECTURES` is a user facing setting. Do not simply override it!

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+else()
+    message("Building with CPU only")
+    set(LIBSUFFIX "cpu")
+
+    project(bitsandbytes LANGUAGES CXX)
+    set(GPU_SOURCES)
+endif()
+
+if(APPLE)
+  set(CMAKE_OSX_DEPLOYMENT_TARGET 13.1)
```
Neumann-A, line 60: this is also user facing

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+endif()
+set(CMAKE_CXX_STANDARD 14)
+set(CXX_STANDARD_REQUIRED C++14)
```
Neumann-A, line 63: probably want to use `target_compile_features` instead.

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+
+if(WIN32)
+    # Mute warnings
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -diag-suppress=177")
+
+    # Enable fast math on VC++
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast")
```
Neumann-A, line 70: don't ! (unless you exactly know why you are activating it and don't have checks like std::is_inf or is_nan as checks.)

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+
+    # Export all symbols
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
+# Weird MSVC hacks
+if(MSVC)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}  /NODEFAULTLIB:msvcprtd /NODEFAULTLIB:MSVCRTD /NODEFAULTLIB:LIBCMT")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
+    set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /arch:AVX2")
```
Neumann-A, line 80: are you using AVX2 intrinsic? Otherwise this is also a user decision.

CMakeLists.txt
akx: `--use_fast_math` is already in the makefile this will eventually replace.

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+    # Enable fast math on VC++
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast")
+
+    # Export all symbols
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
+# Weird MSVC hacks
+if(MSVC)
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}  /NODEFAULTLIB:msvcprtd /NODEFAULTLIB:MSVCRTD /NODEFAULTLIB:LIBCMT")
```
Neumann-A, line 78: This means somebody has inconsistent CRT linkage. (Shouldn't really be required)

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+
+target_include_directories(bitsandbytes PUBLIC
+    ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}
+    ${CMAKE_CURRENT_SOURCE_DIR}/csrc
+    ${CMAKE_CURRENT_SOURCE_DIR}/include)
+
+if(ENABLE_CUDA)
+    target_include_directories(bitsandbytes PUBLIC ${CUDA_TOOLKIT_ROOT_DIR}/include)
+
+    set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --use_fast_math")
```
Neumann-A, line 98: I prefer `string(APPEND CMAKE_CUDA_FLAGS  " added_flag")` however is this a build requirement? I think no. If you want it for your builds add it via `-DCMAKE_CUDA_FLAGS` on the command line instead.

CMakeLists.txt
wkpark: I can tell without this hack MSVC build works fine.

CMakeLists.txt
wkpark: Celeron does not support AVX2 for example

Neumann-A: This would also override user settings of `/arch:AVX512` since MSVC takes the last flag

CMakeLists.txt
rickardp: Sounds reasonable, I'll change this

CMakeLists.txt
rickardp: The code fails to compile on a lower target. For my understanding, why would you want to change this without chaning source code?

CMakeLists.txt
rickardp: I'm keeping @wkpark's solution when merging

CMakeLists.txt
rickardp: I'll remove it if it compiles

CMakeLists.txt
rickardp: > This would also override user settings of `/arch:AVX512` since MSVC takes the last flag

The source code uses AVX2 intrinsics. "user" here is developer, i.e. you need to change source code to make use of AVX512.

Neumann-A: > The source code uses AVX2 intrinsics. "user" here is developer, i.e. you need to change source code to make use of AVX512.

That is incorrect. MSVC can compile AVX2 intrinsic without the `/ARCH:AVX2` since most intrinsics are visible from `immintrin.h`. The /ARCH flags just allows the compiler to use these intrinsics itself while optimizing the code.

rickardp:
This PR should now be integrated with the latest changes

CMakeLists.txt
rickardp: Yes, sorry if I was unclear. I just meant that since the AVX2 intrinsics is part of the source code, does it make sense to expose this to the end user compiling from source?

rickardp: To clarify, I don't mind removing this or making it overridable. I just want to understand how to design the replacement :)

requirements.txt
rickardp: I agree with @akx here, conda is more geared towards setting up dev machines and not so much for CI building of PyPi packages. I would argue we should look at something like [Poetry](https://python-poetry.org/docs/dependency-specification/) if we want something more advanced than what is in this PR. Admittedly I've never worked with Poetry but I skimmed through their documentation and examples and it looks really nice to work with. Anyone with real world experience of Poetry or any high level pip-compatible build backend naturally overrides what I just wrote :)

I would argue we stick with just simple, non-opinonated Python build tools first, then look to migrate in subsequent PRs. Using the `build` package like it does here should serve as a good baseline IMHO.

akx: I would recommend Hatch over Poetry (but that's for a future PR).

CMakeLists.txt
rickardp: The hack could be safely removed, which has now been done. I am not sure why this was needed in the past

requirements.txt
rickardp: > I would recommend Hatch over Poetry (but that's for a future PR).

Interesting, yes that sounds good too. I just went by the GitHub start count to be honest, but practical experience wins as usual :)

younesbelkada:
Amazing work @rickardp ! we'll review this asap with @Titus-von-Koeller !

akx:
@rickardp You should be able to run `pip install pre-commit` and `pre-commit run --all-files` to autofix instead of saving things manually 😄

rickardp:
> @rickardp You should be able to run `pip install pre-commit` and `pre-commit run --all-files` to autofix instead of saving things manually 😄

Got it! But now I wanted to confirm that saving the files in VS code made the linter happy.

akx:
A `cmake -DCOMPUTE_BACKEND=mps .` compilation builds fine on my Macbook :tada:

rickardp:
> A `cmake -DCOMPUTE_BACKEND=mps .` compilation builds fine on my Macbook :tada:
> 
> 
> 
> Some additional patches in `cuda_setup/` (which really isn't `cuda_setup` at this point at all) are required to find the dylib on Macs though, but that can be done in a follow-up 😄 

Yes, I purposefully left that part out, as well as the unit test fixes. Those are still left in the draft PR I created (#947) if anyone wants to have a look. I think my solution to those predates the later discussions and work by other contributors so I think this is more of historical value right now. Happy to collaborate on a follow up PR once the CUDA setup refactoring has been settled.

akx:
@Titus-von-Koeller @younesbelkada I think we should prioritize getting this merged to enable other work :)

wkpark:
This PR tried to address a lot of issues, and I've drawn/referenced heavily from it for my PRs, which I've adapted. so thank you very much!!
and I've tested already but this PR is not completed yet.
- CPU module not working. (setup.py is not correctly fixed.)
- MPS module not working yet.

so I've extracted some working stuff with minimal fixes to reduce possible conflicts into #1018, #1019

I suspect @rickardp had this in mind, too, he already has extracted some PRs.
I think it would be easier to leave this PR as a reference and review each extracted PR individually.

Titus-von-Koeller:
Yes, I agree. I've been trying to merge it from my phone yesterday but it's not allowing me from this device somehow. I'm on vacation until next Monday. I'll figure sth out.

Thanks for the wonderful work everyone once again!

Titus-von-Koeller:
Ok, well, seems I can merge through the mobile browser but not the app.

Please resolve merge conflicts and @younesbelkada and I will merge right away.

akx:
> I'm on vacation until next Monday. I'll figure sth out.

Hey, no worries, enjoy your vacation! 🏖️  There's no hurry here.

rickardp:
This PR is now integrated with the latest `main`.

> CPU module not working. (setup.py is not correctly fixed.) MPS module not working yet.

I commented this earlier. It is left in the "old reference" PR. I think it conflicts with the ongoing refactoring of cuda_setup which is why I think that needs to settle before attempting to integrate this. With this PR, macOS builds can run the CPU tests but does not use MPS.

> I suspect @rickardp had this in mind, too, he already has extracted some PRs. I think it would be easier to leave this PR as a reference and review each extracted PR individually.

My idea was to break out some of the stuff to make it easier to merge this one. Note that I do think the remainder of this PR still needs to be merged (either through this or through a new one) as this PR still does this

- Sets up the pip-based workflow with the xplat targets for macOS
- Adds the cmake changes to support macOS
- Builds CPU only for macOS

If so desired, I think the easiest way if we don't want the dead code of the .metal file in this PR, is to simply remove them from this PR. They don't do anything, but it is currently dead code until the MPS parts are added back

younesbelkada:
Thanks a lot @rickardp @akx @wkpark for your hardwork and for the entire refactor !
Our previous Dockerfiles uses `make cuda11x` to build bnb from source in order to run our tests, I am trying to adapt the docker files to reflect the correct way of building bnb from source here: https://github.com/huggingface/peft/pull/1437 but with no success at the moment, could you help out there ? 🙏

PR #951, opened 2024-01-03
Remove redundant key in env_vars.py

kpister:
Removes redundant key from the set. Also reformats the comment to match other styles.


Titus-von-Koeller:
Makes sense, thanks for catching this and happy for your contribution 🤗

Issue #937, opened 2023-12-21
Bug when moving 4bit quantized model betwee CPU and GPU


michaelnny:
Hi,

I've found new implementation of Params4bit.cuda() will cause issue when moving a quantized model to CPU and GPU. The issue seems to be caused by the removal of the old condition check `if self.quant_state is not None`.

https://github.com/TimDettmers/bitsandbytes/blob/f63abb5a0d0bc971d28972ba890a9e59596caac4/bitsandbytes/nn/modules.py#L167-L173

For example, if I temporarly move the quantized model to CPU, and then try to move it back to GPU, then the already quantized weights will be quantized again. 

```
class QuantizedModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.ln = bnb.nn.Linear4bit(16, 16, bias=False, quant_type='nf4')

    def forward(self, x):
        return self.ln(x)

# step 1 - initialize model
bnb_model = QuantizedModel()
bnb_model.to('cuda')

# This works fine
bnb_output = bnb_model(input.to('cuda'))
print(bnb_output)

# step 2 - move model to CPU
bnb_model.to('cpu')

# step 3 - move model back to GPU     <==== this re-quantize cause shape mismatch during forward pass
bnb_model.to('cuda')
bnb_output = bnb_model(input.to('cuda'))
print(bnb_output)

```

Full error message
```
RuntimeError                              Traceback (most recent call last)

[<ipython-input-12-7e73cf2ab8eb>](https://localhost:8080/#) in <cell line: 2>()
      1 bnb_model.to('cuda')
----> 2 bnb_output = bnb_model(input.to('cuda'))
      3 print(bnb_output)

8 frames

[/usr/local/lib/python3.10/dist-packages/bitsandbytes/autograd/_functions.py](https://localhost:8080/#) in forward(ctx, A, B, out, bias, quant_state)
    514         # 1. Dequantize
    515         # 2. MatmulnN
--> 516         output = torch.nn.functional.linear(A, F.dequantize_4bit(B, quant_state).to(A.dtype).t(), bias)
    517 
    518         # 3. Save state

RuntimeError: mat1 and mat2 shapes cannot be multiplied (4x16 and 1x128)
```

UPDATE: The full test code to reproduce the issue can be found in the following Colab notebook:
https://colab.research.google.com/drive/1rBWjA5VsWdTE6GiATJWHFKHbdt-HrHE4?usp=sharing

PR #953, opened 2024-01-04
Fix bug: when moving quantized model between CPU and GPU, the 4bit weights gets re-quantized

michaelnny:
when moving quantized model between CPU and GPU, the 4bit weights gets re-quantized, this issue is reported #937 

The purposed code add a condition check and only apply the quantization if no quant_state exists.

The full test code can be found in this Colab notebook:
https://colab.research.google.com/drive/1rBWjA5VsWdTE6GiATJWHFKHbdt-HrHE4?usp=sharing

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -165,6 +165,11 @@ class Params4bit(torch.nn.Parameter):
         return self
 
     def cuda(self, device):
+        if self.quant_state is not None:
+            if self.data.device != device:
+                self.data = self.data.to(device)
+                self.quant_state.to(device)
+            return self
         w = self.data.contiguous().half().cuda(device)
         w_4bit, quant_state = bnb.functional.quantize_4bit(w, blocksize=self.blocksize, compress_statistics=self.compress_statistics, quant_type=self.quant_type)
         self.data = w_4bit


Titus-von-Koeller:
Hi @michaelnny,

sorry for not reacting earlier. I was off sick and we're generally still struggling to get on top of the maintenance load, as @younesbelkada and I are taking over maintenance from Tim.

Thanks for raising this issue and taking the initiative to contribute code. In this specific case, we've already had [a related PR that we had to revert](https://github.com/TimDettmers/bitsandbytes/commit/a403c0ed060981a77529d263f1fe6a5102418388), due to failing PEFT integration tests with an error thrown due to something off about stuff being on the wrong device. So good catch, I think your code would have fixed that with the code

```
            if self.data.device != device:
                self.data = self.data.to(device)
                self.quant_state.to(device)
```

We actually discussed this now in [this new PR for FSDP where this issue now also got fixed](https://github.com/TimDettmers/bitsandbytes/pull/970#issuecomment-1898010581). This will be part of the next release.

Sorry for not being able to react earlier and thanks again for your valuable contribution, even if your solution didn't make it over the finishing line this time! Would be happy to see other future contributions by you :)

Seems now this is fixed, due to the [new implementation here](https://github.com/TimDettmers/bitsandbytes/pull/970/files#diff-4d235c7e595546c6656c229dfa139298ce6602b356c2d0bafcb2352eb2cfae79R192-R194), as part of the PR for the FSDP functionality.

michaelnny:
@Titus-von-Koeller Hi, thanks for the update. Yes, earlier today I saw the new PR for FSDP has already fixed this issue, so I'm closing this particular pull request now.

PR #970, opened 2024-01-17
Initial FSDP Support for QLoRA Finetuning

warner-benjamin:
This PR adds initial FSDP support for training QLoRA models. It enables basic FSDP and CPU Offload support, with low memory training via `FSDP.sync_module_states` option unsupported.

This PR builds off of #840 commit 8278fca and [BNB FSDP](https://github.com/Titus-von-Koeller/bitsandbytes/tree/fsdp) by @TimDettmers and @Titus-von-Koeller.

An example of using this PR to finetune QLoRA models with FSDP can be found in our demo script: [fsdp_qlora](https://github.com/AnswerDotAI/fsdp_qlora).

## Rational
The primary blocker for FSDP QLoRA finetuning is the quantized storage type of uint8. FSDP can only shard float data types. Additionally, every time FSDP moves a Linear4Bit from CPU to GPU when using CPU offloading will result in a quantization of existing data. Even if the data is already quantized.

### Selectable Quantization Storage
This PR adds a selectable quantization storage option `quant_storage` to `Linear4Bit` and `Params4Bit`. The quantization storage dtype defaults to `torch.uint8` for backward compatibility with existing code.

While selecting any floating-point storage type will allow FSDP to shard `Linear4Bit` layers, setting the quantization storage dtype to match the rest of the non-LoRA layers' dtype allows `Linear4Bit` layers to be wrapped identically to `Linear` layers in a LoRA wrapping policy, such as the [`fsdp_auto_wrap_policy` from llama-recipes](https://github.com/facebookresearch/llama-recipes/blob/main/src/llama_recipes/policies/wrapping.py).

If the quantization storage dtype does not match the rest of the layers's dtype, then the `Linear4Bit` layers will have to be wrapped individually.

### Prevent Multiple Quantization
The PR adds a quantization flag to prevent `Params4Bit` from quantizing already quantized params when transferring from CPU to GPU. For example, when training with FSDP CPU offloading.

### Set Quant State during FSDP Forward
FSDP does not copy the `Params4Bit` QuantState dictionary in when moving sharded layers. This PR sets the QuantState as a component of `Linear4Bit` and copies it to `Params4Bit` if it no longer exists in `Params4Bit`.

## Testing
This PR adds `quant_storage` testing to the Linear4Bit tests and fixes an issue with the current tests where NF4 wasn't tested.

We also tested these changes against PEFT's QLoRA tests and did not find any regressions from the current bitsandbytes behavior.

We have also tested FSDP Mixed Precision in fp32 and bf16 and noticed no changes in training behavior when setting the `Linear4Bit` and `Params4Bit`'s `quant_storage` dtype to match the FSDP `MixedPrecision.param` dtype.

We have successfully finetuned Llama-2, Mistral, and TinyLlama models with FSDP & QLoRA using our [demo script](https://github.com/AnswerDotAI/fsdp_qlora).

## Downstream Implications
Existing implementations may require some modification to work with this, for example:

- Model loading via Transformers `load_in_4bit` will need a way to set `quant_storage` (the [demo script](https://github.com/AnswerDotAI/fsdp_qlora) uses custom model loading)
- PEFT's `prepare_model_for_kbit_training` upcasts all non-uint8 params to float32 under the assumption that the base (quantized) weights are stored in uint8, which is now no longer guaranteed
- PEFT's `get_nb_trainable_parameters` multiplies the number of parameters from `Params4Bit` by two which is only valid if `quant_storage` is uint8

## Future Work
Currently QLoRA finetuning using FSDP's low memory via the `sync_module_states` option doesn't work. Enabling this will require a future PR.

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -607,7 +607,7 @@ class QuantState:
 
         qs_dict: based on state_dict, with only relevant keys, striped of prefixes.
 
-        item with key `quant_state.bitsandbytes__[nf4/fp4]` may contain minor and non-tensor quant state items.        
+        item with key `quant_state.bitsandbytes__[nf4/fp4]` may contain minor and non-tensor quant state items.
         """
 
         # unpacking tensor with non-tensor components
@@ -802,7 +802,7 @@ def dequantize_blockwise(
 
     if quant_state is None:
        quant_state = QuantState(absmax=absmax, code=code, blocksize=blocksize, dtype=torch.float32)
-    
+
     absmax = quant_state.absmax
     if quant_state.nested:
         absmax = dequantize_blockwise(quant_state.absmax, quant_state.state2)
@@ -931,7 +931,7 @@ def quantize_4bit(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksiz
 
 
     if out is None:
-        out = torch.zeros(((n+1)//2, 1), dtype=torch.uint8, device=A.device)
+        out = torch.zeros(((n+1)//8, 1), dtype=torch.float32, device=A.device)
 
     assert blocksize in [4096, 2048, 1024, 512, 256, 128, 64]
 
@@ -1626,7 +1626,7 @@ def gemv_4bit(
     ldb = ct.c_int32(ldb)
     ldc = ct.c_int32(ldc)
 
-    if B.dtype == torch.uint8:
+    if B.dtype == torch.float32:
         if A.dtype == torch.float16:
             lib.cgemm_4bit_inference_naive_fp16(m, n, k, get_ptr(A), get_ptr(B), get_ptr(absmax), get_ptr(state.code), get_ptr(out), lda, ldb, ldc, ct.c_int32(state.blocksize))
         elif A.dtype == torch.bfloat16:
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -884,13 +884,13 @@ def get_4bit_type(typename, device=None, blocksize=64):
     return data.to(device)
 
 
-def quantize_fp4(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False):
-    return quantize_4bit(A, absmax, out, blocksize, compress_statistics, 'fp4')
+def quantize_fp4(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False, quant_storage=torch.uint8):
+    return quantize_4bit(A, absmax, out, blocksize, compress_statistics, 'fp4', quant_storage)
 
-def quantize_nf4(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False):
-    return quantize_4bit(A, absmax, out, blocksize, compress_statistics, 'nf4')
+def quantize_nf4(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False, quant_storage=torch.uint8):
+    return quantize_4bit(A, absmax, out, blocksize, compress_statistics, 'nf4', quant_storage)
 
-def quantize_4bit(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False, quant_type='fp4') -> Tensor:
+def quantize_4bit(A: Tensor, absmax: Tensor = None, out: Tensor = None, blocksize=64, compress_statistics=False, quant_type='fp4', quant_storage=torch.uint8) -> Tensor:
     """
     Quantize tensor A in blocks of 4-bit values.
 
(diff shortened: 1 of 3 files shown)


Titus-von-Koeller:
Thank you all for your fine work and describing it thoroughly. We're very happy with the collaboration and after our talks in the last days and having given the code an initial review, the next step is to merge onto main as that triggers our daily pre-release CI pipeline on the HF side, running all the HF integration tests, making sure that nothing breaks in Transformers, PEFT + Accelerate. (This is our workaround for not having our own GPU runners, so we're using a pipeline on the HF side, which doesn't yet include the BnB tests itself.)

Have you run the BnB test suite itself? Everything looking good there so far? We have a few flaky tests that we still need to make more reproducible, but this is still important information. It would be good to paste the output here for review.

The procedure now is that we'll do a preliminary merge, get back to you tmr with the integration test results tmr and then speak about the next steps in our video call, including any potential improvements that we still might want to add, as well as Transformers integration, etc.

Thanks again for your contribution and good work! Really happy to move forward with this.

warner-benjamin:
Further testing shows no issues using FSDP Mixed Precision with `autocast` when `Linear4Bit.quant_storage=torch.float32` to match the FSDP Mixed Precision `param_dtype`: `MixedPrecision(param_dtype=torch.float32, reduce_dtype=torch.float32, buffer_dtype=torch.float32)`

Titus-von-Koeller:
```
Results of the single_gpu_huggingface/peft-gpu-bnb-source:latest PEFT scheduled tests.
core_single_gpu.log: 2 failed tests
+----+-----------------+--------------+----------------------------------+
|    | Test Location   | Test Case    | Test Name                        |
+====+=================+==============+==================================+
|  0 | tests/test_c    | PeftGPUCommo | test_4bit_merge_and_disable_lora |
|    | ommon_gpu.py    | nTests       |                                  |
+----+-----------------+--------------+----------------------------------+
|  1 | tests/test_c    | PeftGPUCommo | test_4bit_merge_lora             |
|    | ommon_gpu.py    | nTests       |                                  |
+----+-----------------+--------------+----------------------------------+

Results of the single_gpu_huggingface/peft-gpu-bnb-latest:latest PEFT scheduled tests.
core_single_gpu.log: 2 failed tests
+----+-----------------+--------------+----------------------------------+
|    | Test Location   | Test Case    | Test Name                        |
+====+=================+==============+==================================+
|  0 | tests/test_c    | PeftGPUCommo | test_4bit_merge_and_disable_lora |
|    | ommon_gpu.py    | nTests       |                                  |
+----+-----------------+--------------+----------------------------------+
|  1 | tests/test_c    | PeftGPUCommo | test_4bit_merge_lora             |
|    | ommon_gpu.py    | nTests       |                                  |
+----+-----------------+--------------+----------------------------------+
```

Both times (HF libs from source vs pip, with BNB always from source) the same tests are failing in the same way:

[core_single_gpu.log](https://github.com/TimDettmers/bitsandbytes/files/13973726/core_single_gpu.log)

This is the same failing test case - but with a different error! - that we got after merging [@jph00 's commit](https://github.com/TimDettmers/bitsandbytes/commit/a403c0ed060981a77529d263f1fe6a5102418388), which we had to then revert. See the error log of [core_single_gpu-Jeremy's_fix.log](https://github.com/TimDettmers/bitsandbytes/files/13973918/core_single_gpu-Jeremy.s_fix.log) for reference, as I think that this might hint that both are related to the prevention of "Params4Bit from quantizing already quantized params when transferring from CPU to GPU".

I'll try to look more into this during the day, but just wanted to share the necessary info with everyone asap.

cc @sourab @younesbelkada @TimDettmers for visibility / collab in fixing this

Titus-von-Koeller:
x-posting @pacman100 's message on Discord here for visibility:

> sourab_mangrulkar — Today at 1:38 PM

> Hello everyone! 😄
> 
> Thank you for adding me to this collaboration and I am looking forward to the QLoRA+FSDP support which would be huge!
> 
> I have gone through the following:
> bitsandbytes  implementation of following classes/functions  Linear4bit, Params4bit, quantize_4bit, matmul_4bit, MatMul4Bit and dequantize_4bit .
> PR https://github.com/TimDettmers/bitsandbytes/pull/970 and demo script https://github.com/AnswerDotAI/fsdp_qlora

> wrt to the CI failures, I have following hypothesis:
> In merge method, the following things happen wherein the lora delta weights are added to the dequantized base layer weights. This is followed by setting the weight of the base layer to a new instance of  Params4bit with the new merged weights:
> 
> w_data = bnb.functional.dequantize_4bit(weight.data, weight.quant_state) + lora_data
> self.get_base_layer().weight = bnb.nn.Params4bit(w_data.to("cpu"), requires_grad=False,**kwargs).to(weight.device)            
> 
> 
> Now, kwargs copies the ones currently there which has bnb_quantized=True and the old quant_state so the new merged weights aren't quantized and the old quant state remains.  After this, during Linear4bit forward call, the old quant state is used while the weights are unquantized.
> 
> Trying out this PR https://github.com/huggingface/peft/pull/1370 to see it gets fixed. 
> I can confirm the above PR fixes the 2 failing tests, yay!

weifengpy:
amazing! will take a look from FSDP side to see what can be improved

Issue #982, opened 2024-01-24
keyword argument repeated: install_requires in setup.py


animeshk08:
### System Info

NA

### Reproduction

run setup.py in any environment. Below runtime exception is thrown
`keyword argument repeated: install_requires`

### Expected behavior

Clean run.

Titus-von-Koeller:
Thanks for raising this!

PR #983, opened 2024-01-24
Fix Remove redundant dependency keyword parameter

animeshk08:
Fix for: https://github.com/TimDettmers/bitsandbytes/issues/982


Titus-von-Koeller:
Good catch. It's great to have so many active contributors on this project. Really happy about it :) Thanks!

PR #984, opened 2024-01-24
Ruff fixes

akx:
This PR applies a bunch of fixes suggested by the Ruff linter (which has been configured since 2c605d03dd4352aa1e3b5f08c8f0bcdec8b48334 (#896), just never really run).

Some of these are mechanical auto-fixes (`--fix --select=I,F401` style), but most of them are manual.

The commits themselves are atomic and small enough to be able to be reviewed one by one.

It would probably be a good idea to enforce linting with Ruff in CI; if the maintainers like, I can contribute a the requisite GitHub Actions configuration too :)

PR #985, opened 2024-01-24
Tests: improve CUDA support detection 

akx:
Most of the tests require CUDA capabilities; only some of them were previously marked as such and skipped when no CUDA was available, the rest of them were left to fail.

This PR adds a [pytest hook](https://docs.pytest.org/en/7.1.x/reference/reference.html?highlight=hooks#pytest.hookspec.pytest_runtest_call) that turns the Torch `Torch not compiled with CUDA enabled` assertion error into `pytest.skip`, and a Pytest fixture that a test can use to assert that CUDA support is available too.

On a machine where `make cpuonly` passes (namely in a `linux/amd64` Docker box), running tests on `main` (53f8af8c90f0f52c17e33d3d9a1a6a1dce398443) results in

> 706 failed, 75 passed, 2339 skipped, 1 warning, 16 errors in 42.86s

With this PR, the result is a more sensible

> 75 passed, 3061 skipped, 1 warning in 9.08s

(though whether `test_nvidia_transform` should (partially) succeed on a machine with no CUDA is anybody's guess).

Related to discussion in #984.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,19 @@
+import pytest
+import torch
+
+
+def pytest_runtest_call(item):
+    try:
+        item.runtest()
+    except AssertionError as ae:
+        if str(ae) == "Torch not compiled with CUDA enabled":
+            pytest.skip("Torch not compiled with CUDA enabled")
+        raise
+
+
+@pytest.fixture(scope="session")
+def requires_cuda() -> bool:
+    cuda_available = torch.cuda.is_available()
+    if not cuda_available:
+        pytest.skip("CUDA is required")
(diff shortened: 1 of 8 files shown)


Titus-von-Koeller:
Looks good to me! Thanks for another useful contribution.

Just out of curiosity, of course you don't have to answer, what led you to work with BNB, if you're not running Linux with a GPU? Are you involved with the Apple Silicon cross platform work?

akx:
@Titus-von-Koeller Let's just say I have my contributing fingers here and there in the GenAI ecosystem 😁

PR #987, opened 2024-01-25
Add .git-blame-ignore-revs file

akx:
This PR adds a [.git-blame-ignore-revs](https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta) to make it easier to find relevant information in `git blame` views, as commits that only refactor code will be ignored there.


Titus-von-Koeller:
Very good addition! This was definitely needed.

Thanks for your continued engagement in the repo. I really value that :D

PR #992, opened 2024-01-27
Update functional.py

sxndqc:
Update type indicator


Titus-von-Koeller:
Good catch! Thanks for your contribution, really appreciated.

akx:
This isn't a valid type annotation, though 😅 

```
bitsandbytes/functional.py:893: error: Syntax error in type annotation  [syntax]
bitsandbytes/functional.py:893: note: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)
```

PR #995, opened 2024-01-29
Fix some issues found by Mypy

akx:
This PR fixes a variety of issues spotted by Mypy.

Most of these revolve around functions being inaccurately typed.

There's also an actual bug fix (the missed call to `is_triton_available`).

The library is by no means yet "mypy clean", though, and this isn't an exhaustive PR :)


Titus-von-Koeller:
Thanks, I reviewed everything, lgtm! We need more of the thoroughness you apply in this repo :) Great fixes and inputs that you are providing. Feels really motivating to have such active contributors 🙌🏻

PR #1000, opened 2024-01-31
test_nvidia_transform: fix variable reference

akx:
`out_order` is the global parametrization list, not the test fixture argument.

Simple typo bug I noticed while working on #1001.


younesbelkada (approved):
Thanks for fixing the test!

Titus-von-Koeller:
very good catch!

PR #998, opened 2024-01-30
Don't crash Python interpreter via `assert(false)` when missing cublaslt

akx:
As discussed in https://github.com/TimDettmers/bitsandbytes/pull/984#issuecomment-1917763792.

Before:
```
tests/test_functional.py::test_dequant_mm[dim1_121_dim4_201_dims_2_formatB_col_turing_has_bias_True] Fatal Python error: Aborted

Thread 0x00007fdb0d0e7640 (most recent call first):
  <no Python frame>

Current thread 0x00007fdc47e18000 (most recent call first):
  File "/home/akx/bitsandbytes/bitsandbytes/functional.py", line 1930 in igemmlt
  File "/home/akx/bitsandbytes/tests/test_functional.py", line 990 in test_dequant_mm
  [...]

Aborted
```

After:

```
tests/test_functional.py::test_dequant_mm[dim1_121_dim4_201_dims_2_formatB_col_turing_has_bias_True] FAILED
[...]
        if has_error == 100:  # `ERR_NOT_IMPLEMENTED` is defined as 100 in `ops.cu`
>           raise NotImplementedError("igemmlt not available (probably built with NO_CUBLASLT)")
E           NotImplementedError: igemmlt not available (probably built with NO_CUBLASLT)
```

PR #1001, opened 2024-01-31
Test improvements

akx:
This PR improves the technical state of the tests some – it doesn't fix flakiness or coverage any.

* The main thing is to use [Pytest's `parametrize`](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) in the way it's intended to be used; stacking multiple `parametrize`s results in a Cartesian product of all of the parameters. Where the original code did this by hand and precariously constructed test ID strings, we'll now let Pytest do it for us. It's about half shorter too. The exact same number of tests is generated as before (3,136) – the ids are different (more readable) in the new version though.
* The original code would generate random dimension numbers for tests on every invocation, making things like `--lf` (run last failed) etc. not exactly work as they should. This is somewhat alleviated, but not perfectly, by using a seeded generator. The number of files visited (i.e. the number of `randint` calls to that seeded generator) affects the result still.
* The `NO_CUBLASLT` `NotImplementedError` implemented in #998 is now translated into a test skip, not a test failure.
* Two Pytest [markers](https://docs.pytest.org/en/7.1.x/example/markers.html) were added, to allow e.g. `-k "not (slow or benchmark)"`.
  * There are still a number of non-benchmark tests that contain internal looping that could be replaced with e.g. [pytest-repeat](https://pypi.org/project/pytest-repeat/) to make it possible to run a shorter or longer test suite. That's out of scope for this PR though.

I tested that tests run just as before on my hardware, and many if not all succeed (`310 failed, 1991 passed, 563 skipped, 18 deselected, 704 warnings in 488.88s (0:08:08)`) – but that's probably the flakiness of the tests in general to blame.

This PR stacks on #1000.

These transformations were done partly with https://ast-grep.github.io/ (h/t @herringtondarkholme!)

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -8,3 +8,6 @@ addopts = -rP
 log_cli = True
 log_cli_level = INFO
 log_file = logs/pytest.log
+markers =
+    benchmark: mark test as benchmark
+    slow: mark test as slow
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,6 +5,10 @@ import torch
 def pytest_runtest_call(item):
     try:
         item.runtest()
+    except NotImplementedError as nie:
+        if "NO_CUBLASLT" in str(nie):
+            pytest.skip("CUBLASLT not available")
+        raise
     except AssertionError as ae:
         if str(ae) == "Torch not compiled with CUDA enabled":
             pytest.skip("Torch not compiled with CUDA enabled")
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,7 +1,14 @@
 from itertools import product
+import random
 from typing import Any
 
 import torch
+
+test_dims_rng = random.Random(42)
+
+
+def get_test_dims(min: int, max: int, *, n: int) -> list[int]:
+    return [test_dims_rng.randint(min, max) for _ in range(n)]
 
 
 def format_with_label(label: str, value: Any) -> str:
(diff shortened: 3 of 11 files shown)


Titus-von-Koeller:
This is really amazing, you're such a pro!

I've really been struggling with the idiosyncratic style of a lot of the code base, which is often very hard to follow and I find that it could profit so much from the kind of Pythonic refactoring that you showcased in this PR. The code in general would profit a lot from being more pythonic: That would make the whole library much more accessible, easy to maintain and contribute to. Code that is written well in the sense that it makes things as understandable as they can be is the best documentation.

A significant step in the right direction. I reviewed every commit and everything looks right to me. Of course, with the test suite in it's current state, it's a bit hard to not be sure we didn't accidentally introduce some error. Anyways, you definitely also at fixed at least one bug and reducing the flakiness is an absolute must. Also, we have to take some calculated risks in improving the code quality in this repo and I'm fully on board with that. You also seem to know what you're doing (: But no worry if sth breaks, then we fix forward..

Please keep it coming, if you feel like it. Super happy to have you on board in this effort!

Titus-von-Koeller:
P.S. Thanks for mentioning https://ast-grep.github.io/

Looks really promising and didn't hear of it yet. I've been looking for good tools to help with refactoring. Will look into it. Let me know if you can recommend other stuff :)

akx:
@BenjaminBossan Absolutely – though in this case it wouldn't help as much, since the original code generated random tests :)

But as soon as we have a test suite that's not as flaky, `pytest-cov` all the way!

PR #1006, opened 2024-02-01
Enable line-ending and other hygiene lints

akx:
Makes sure all files end with a newline, as is good and proper, and that empty lines don't have phantom whitespace... And other small relevant lints.


Titus-von-Koeller:
Thank you 🙌🏻

Issue #1010, opened 2024-02-01
CI: don't build separate wheels for different Python versions


akx:
We don't need to build Python wheels with both `cp310` and `cp311` tags; Python 3.11 (and 3.12) is fine with `cp310` wheels. Since we don't actually use Python's extension ABI, that should be fine no matter what.

We just need to figure out a stable way to force always emitting a `cp310` wheel. https://github.com/karellen/wheel-axle/ seems to do that but I'm a bit hesitant to add a whole new dep for that.

younesbelkada:
Thanks @akx ! If I understand correctly this is only for python == 3.10 & 3.11 ? For older python versions we need to build separate wheels right?

akx:
No, a. cp38 wheel should work just as fine (though it needs a bit of experimentation).

matthewdouglas:
Closing this as we've been building one set of wheels tagged as just `py3` for all Python versions.

PR #1011, opened 2024-02-01
Tweak/optimize Cmake build slightly

akx:
This follows up on #908 to make it do less work and simplify things some.

Since we don't actually link against the Python ABI, we only need to build on Python 3.10. CPython 3.10 _and newer_ should happily accept a `cp310` wheel:

```
$ docker run -it --platform=linux/amd64 python:3.12 sh -c 'pip debug --verbose | grep cp310'
[...]
  cp310-abi3-manylinux1_x86_64
  cp310-abi3-linux_x86_64
```

(#1010 discusses being able to force the wheel tag, so we could build on a newer, faster Python, but still get a cp310 wheel.)

### DIFF
--- a/.github/workflows/build_pr_documentation.yml
+++ b/.github/workflows/build_pr_documentation.yml
@@ -9,6 +9,7 @@ concurrency:
 
 jobs:
   build:
+    if: github.repository == 'TimDettmers/bitsandbytes'
     uses: huggingface/doc-builder/.github/workflows/build_pr_documentation.yml@main
     with:
       commit_sha: ${{ github.event.pull_request.head.sha }}
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [ "main" ]
 
+concurrency:
+  group: cmake-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -148,12 +148,6 @@ jobs:
     - name: Build CPU
       run: cmake --build ${{ steps.strings.outputs.build-output-dir }} --config ${{ matrix.build_type }}
 
-    - name: Test
-      working-directory: ${{ steps.strings.outputs.build-output-dir }}
-      # Execute tests defined by the CMake configuration. Note that --build-config is needed because the default Windows generator is a multi-config generator (Visual Studio generator).
-      # See https://cmake.org/cmake/help/latest/manual/ctest.1.html for more detail
-      run: ctest --build-config ${{ matrix.build_type }}
-
     - name: Build dist
       shell: bash -el {0}
       run: |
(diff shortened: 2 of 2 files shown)


akx:
cc @wkpark

wkpark:
what about python3.9 or 3.8? pytorch nightly build still support python3.8/3.9 (https://download.pytorch.org/whl/nightly/torch/  Of course, I don't expect anyone to still be using the old python3.8 for normal use.)

akx:
Should be fine to lower the wheel tag to `cp38` eventually!

.github/workflows/build_pr_documentation.yml
```diff
@@ -9,6 +9,7 @@ concurrency:
 
 jobs:
   build:
+    if: github.repository == 'TimDettmers/bitsandbytes'
```
younesbelkada, line 12: can you elaborate on this change? Would this affect PRs from forks?

.github/workflows/cmake.yml
```diff
@@ -61,7 +47,7 @@ jobs:
         environment-file: environment-bnb.yml
         use-only-tar-bz2: false
         auto-activate-base: true
-        python-version: ${{ matrix.python-version }}
+        python-version: "3.10"
```
younesbelkada, line 50: Hmm thinking about it, ideally I think we should add also another "old enough" python version such as 3.8 just to be sure 
In the past we had some issues due to a wrong typehint (!) that broke transformers for python 3.8: https://github.com/TimDettmers/bitsandbytes/pull/859 so I think adding it would circumvents such an issue being on main - what do you think?

younesbelkada: I just saw https://github.com/TimDettmers/bitsandbytes/issues/1010 I think what you said makes sense, we could do the python=3.8 build in another PR - what do you think?

.github/workflows/build_pr_documentation.yml
akx: Yes. I was running exploratory versions of this PR in my own fork, and the docs workflow wouldn't be necessary there.

.github/workflows/cmake.yml
akx: We would add a follow-up step that tests the built libraries on all supported versions, but building the binaries (and wheel) only needs to occur on one version.

.github/workflows/build_pr_documentation.yml
younesbelkada: ok makes sense

younesbelkada (approved):
Thanks again !

younesbelkada:
cc @wkpark all good if we merge this?

wkpark:
> all good if we merge this?

Yes. looks good to me.

Titus-von-Koeller:
Hey, I am not completely up to speed with everything, but I noticed that all the builds are kicked off even when I just add a documentation commit. That seems pretty wasteful.

Wouldn't it make sense to only trigger those workflows in cases where there was a change detected in the csrc directory and otherwise use the last cached one form a parent commit?

Not sure how easy/hard that is. Just sth that came to mind. It also doesn't need to be addressed right away, even if it's something we want to do.

younesbelkada:
@Titus-von-Koeller great point, actually we can do that in a straightforward manner with GHA, let me open a quick PR and ping you

younesbelkada:
@Titus-von-Koeller : https://github.com/TimDettmers/bitsandbytes/pull/1015

PR #1015, opened 2024-02-02
ENH: Run the CMake CI only when relevant files are modified

younesbelkada:
As per title

cc @Titus-von-Koeller @akx @wkpark 

Similar PR as: https://github.com/huggingface/trl/pull/1309

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -5,6 +5,14 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+    paths:
+      # Run only when relevant files are modified
+      - "csrc/**"
+      - "include/**"
+      - "bitsandbytes/**"
+      - ".github/**.yml"
+      - "tests/**"
+
 
 concurrency:
   group: cmake-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


younesbelkada:
Turns out this simply filters out PRs that touches certain files, but it will still trigger the CI on each commit in case that condition is not met. For changing the behaviour to inspect each commit, we need a more elaborated workflow and I am not sure it is something we 100% want for now

akx:
I'm not sure this is worth it right now. I foresee it'll bite us when CI is unexpectedly not run when it should because someone reorganizes the repo and doesn't know this filter list is there.

Titus-von-Koeller:
I would really like to have this, but it's not a top prio. @wkpark and @rickardp do you see any issues with this approach or something similar?

It just seems so wasteful to run on every commit, unless it's really needed.

rickardp:
I added this too, because the docs were updated a lot. But maybe it would be better to turn it around and exclude patterns instead?

Titus-von-Koeller:
ah, I see, you mean #1051

I hadn't had time to review all the open PRs, due to needing to concentrate on other ongoing work.

rickardp:
What is in this PR is already merged. Note that this PR references a different file.

I think it is worth considering reversing the pattern to avoid that a file structure change causes workflows not to run. But there are downsides to that approach too.

younesbelkada:
Thanks everyone for the clarification ! I just closed the PR :D

PR #996, opened 2024-01-30
CUDA setup cleanup

akx:
This follows up on #873 (and #703/#527) to slightly clean up the CUDA setup and diagnostics code.

The last commit is the most interesting code-wise; #703/#527 added the sensible-enough behavior to sort CUDA version strings as sequences of integers, but since the retval was again re-sorted in `evaluate_cuda_setup()`, the sort was a no-op.
This removes the "dumber" sort from `evaluate_cuda_setup()`.

PR #1017, opened 2024-02-02
Fix usage of BNB_CUDA_VERSION for manual override

matthewdouglas:
I'm aware of related PRs like #996, but for now I needed to make a couple smaller changes in order to do some testing when building against CUDA versions that don't match the PyTorch distributions.


akx (approved):

matthewdouglas:
Closing in favor of #996.

PR #1019, opened 2024-02-03
CI: split workflows

wkpark:
split cmake workflows into several jobs, based on #949, #947 @rickardp and @akx's workflows work.

 * [x] split build job into shared-libs, shared-cuda, wheels jobs.
 * [x] add ubuntu aarch64 build.
 * [x] reduce `environment-bnb.yml` package dependency to reduce build time.


Titus-von-Koeller:
Thanks very much @wkpark! yes, makes sense to split it out.

PR #1025, opened 2024-02-03
[Portability] Add build artifacts to .gitignore

rickardp:
Code change broken out from PR https://github.com/TimDettmers/bitsandbytes/pull/949 to keep this change separate from the portability/pipelines in that PR.

Currently running cmake produces changes to version controlled files. This PR adds build output and cmake cache files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ bitsandbytes.dir/
 Debug/
 Release/
 
+# IDE local files
+.vs/
+
 # Distribution / packaging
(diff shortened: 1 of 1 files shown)


matthewdouglas:
While we're at it can we add `.vs/` as well for files created in Visual Studio too? I was going to open a similar PR to this but you beat me to it :)

younesbelkada (approved):
Makes sense, thanks !

PR #1028, opened 2024-02-04
Fixed VS Code format on save

rickardp:
There is a pipeline linter now that checks that the files are correctly formatted.

This PR will make VS Code format the files correctly on save.

The .editorconfig should work with most IDEs.


rickardp:
There is now two code formatters configured for this project: Yapf and Ruff. Would it make sense to align on one of them?

akx:
This repo isn't using `ruff format` (or `black`) for Python files; the Yapf configuration file was just copied over here but it too has never been actually run.
There [was discussion about code formatting in another PR, but no concrete decision was made](https://github.com/TimDettmers/bitsandbytes/pull/994#issuecomment-1915968196). (I'd prefer `ruff format` with a line-length of, say, 120.)

This change is fine, though, running the Ruff _lint_ autofixes via VSCode and having a compliant `editorconfig` is a great DX improvement.

akx (approved):
👍

rickardp:
> This repo isn't using `ruff format` (or `black`) for Python files; the Yapf configuration file was just copied over here but it too has never been actually run. There [was discussion about code formatting in another PR, but no concrete decision was made](https://github.com/TimDettmers/bitsandbytes/pull/994#issuecomment-1915968196). (I'd prefer `ruff format` with a line-length of, say, 120.)
> 
> This change is fine, though, running the Ruff _lint_ autofixes via VSCode and having a compliant `editorconfig` is a great DX improvement.

Yeah, I just learned of Ruff from your PR actually. I've always been using black, but Ruff is lightning fast and seems to share the same no-bikeshedding design goal, which I really like. I actually thought Ruff was running formatting via the pre-commit toolchain as it wanted to delete duplicate linefeeds etc, but I did not look into it further. I agree that Ruff seems like a good choice, and I do think that running a complete Python format-on-save would be a quality of life improvement

akx:
Yeah, Ruff has two personas, as it were: the linter, which has most of flake8's rules implemented, but notably not the whitespace rules just yet (there's a PR pending for that though: https://github.com/astral-sh/ruff/pull/9266) _and_ a Black-compatible formatter, which _also_ cleans up whitespace to adhere to the same rules, but there's no way to apply just that part of the formatter (for technical raisins). There's a clear usecase for only wanting (part of) the whitespace rules instead of the wholesale reformatting that the formatter would apply too :)

Titus-von-Koeller:
Hey @rickardp,

thanks for the PR. This useful.

Titus-von-Koeller:
@akx

Actually, I had posted an update with a decision to go forward with the black / ruff format approach and deprecate yapf.

See https://github.com/TimDettmers/bitsandbytes/pull/994#issuecomment-1922125255

We should also set up a c-format formatter for C/C++/CUDA.


Thanks a lot both! This kind of work will make it much easier for everyone to collaborate and contribute.

I ll be back from vacation in a week.

PR #1018, opened 2024-02-03
minimal macos support

wkpark:
This is a PR to support macos based on @rickardp's work (PR #949 )
for those who want to use bitsandbytes under OSX with minimal support.

 * original work done by @niclimcy and @rickardp
 * [x] ~~macos cpu module support~~
 * [x] load `libbitsandbytes_cpu.dylib` correctly.

Issue #1031, opened 2024-02-04
[RFC] Cross-Platform Refactor: Testing and CI/CD Strategy


Titus-von-Koeller:
The goal of this issue is to:

- Combine discussions on testing frameworks, CI/CD infrastructure, and handling flaky tests into one comprehensive RFC.
- Address how to run tests across different backends, the development of a mature CI system that covers each cross-platform backend, and strategies to ensure test reproducibility and tackle flaky tests.
- Discuss the current limitations, such as the lack of GPU-accelerated GitHub runners for Linux and the manual testing process, and explore solutions for a more automated and inclusive testing environment.

**Tackling Flaky Tests and Reproducibility Issues**: Address the challenges of flaky tests and non-determinism in outputs, seeking community input on establishing more robust testing practices.

**Enhancing the Testing Framework for Multi-Backend Support**: Focus on discussing the strategies for testing different backends, addressing the core question of how to ensure comprehensive test coverage across all supported platforms without hindering contributions.

## **Status quo**
The current status quo that we don't even have GPU accelerated Github runners (Linux). So far we manually test contributions or have them submit their test results in the PR. Recently there has been the addition of an integration pipeline (thanks @younesbelkada), testing the HF library integrations nightly, on the Hugging Face side, but these results cannot directly be fed back into the PRs (correct me if I'm wrong) and have been manually posted to the PR.

## **Support needed for Github Runner infra**
Given that we're planning to add Intel, AMD, Mac, Windows support as officially supported, the current approach is no longer feasible. In order to find a way forward that works well for all the just mentioned platforms, we would like to have an open discussion in this thread to try and align everyone around an approach that preferably works for everyone.

It will be hard to change [this architecture decision](https://github.com/TimDettmers/bitsandbytes/pull/898#issuecomment-1909439116) once it's in place, that's why we want to take this step with caution, including stakeholders from each platform. **To summarize, we think a cross-platform CI is needed to make sure we're releasing without fear of breaking the non-Linux + CUDA backends. We would welcome the communities contributions in order to make this a reality.**

Unfortunately, we cannot use the HF organization's Github runner infrastructure, as BNB shall remain independent and using the runners would not comply with HFs security policies as long as BNB is not part of the org. This means in practice that BNB will need support in the form of compute grants to run CI/CD. Hugging Face already agreed to step up for a portion of that. However, Intel and AMD would need to contribute from their side both with compute and helping us build up the CI/CD system to make this run smoothly.

[...]

PR #1035, opened 2024-02-05
CI: fix workflows

wkpark:
~~fix regressions introduced by PR #949~~

- [x] ~~PR #1038 included.~~ (merged)

# misc workflows fixes
## restored
restore some accidentally and not carefully merged but abandoned stuff.
- [x] support cmake `-B build` option correctly.
- [x] use python-version=3.10 for builds
- [x] build a wheel with all available cuda versions.
- [x] ~~use conda+mamba method. disable cuda-toolkits. (the cuda-toolkits is too slow and not stable in some cases.)~~
- [x] use more flexible/customizable `cmake` options. (set `COMPUTE_CAPABILITY` for example)

----
## fixed
- [x] fix Jimver/cuda-toolkit speed issue (extracted to #1055) ~~cuda-toolkit seems too slow and even breaks sometimes. (Please see https://github.com/wkpark/bitsandbytes/actions/runs/7793281860/job/21252763742)~~
- [x] fix wheel names for aarch64
- [x] ~~fix docker using `container: image:` method.~~ use docker-run-actions
- [x] update docker image version
  - fixed deprecation warning for cuda12.1 docker image (cuda12.1.0 is deprecated but cuda12.1.1 is supported)

## misc
- [x] drop artifact `retention-days:` it could be configured Settings->Actions->General.
- [x] `concurrency:` fix restored.
- [x] `fail-fast: false` restored.
- [x] use `ilammy/msvc-dev-cmd@v1.13.0` instead of `microsoft/setup-msbuild`
   -  `microsoft/setup-msbuild` is used for msbuild. `ilammy/msvc-dev-cmd@v1.13.0`
- [x] use `cmake -G Ninja` : ninja is more faster and popular. ([triton](https://github.com/openai/triton) use Ninja for example)
- [x] differently define `COMPUTE_CAPABILITY`
   - for `pull_request`, use concise `COMPUTE_CAPABILITY=50;52;60;61;62;70;72;75;80;86;87;89;90`
   - for `refs/tags/*`, use full capabilities, `COMPUTE_CAPABILITY=61;75;86;89` (only popular GPUs included)
   - with this method, build time reduced from ~40min to ~25min
- [x] build cuda12.1 only for pull_request, full build for `refs/tags/*`

## aarch64 build issue
Build time using docker + aarch64 arch seems too slow. current total build time is about 30~40min.

- normal build time is about ~5 ~10min, aarch64 build on docker is about ~25 ~30min
- we can use cross-compiler to fix the build speed issue.
- or we can reduce `COMPUTE_CAPABILITY` for `pull_requests`

***outdated or resolved***
~~CPU builds with aarch64 gcc/g++ are fine, but for cuda builds there is no benefit as docker itself does not have arm64 libraries installed and it will fail to link correctly.~~ resolved.

~~Some of @rickardp's changes, whether intended or not, maybe better done in `python-package.yml`. However, it causes some serious problems, and it is important to quickly fix things that were fine before, I am submitting this PR.~~

~~P.S.: PR #1018 is not included.~~


akx (commented):
At first glance, this basically reverts a lot of Rickard's work - could you explain why the changes are needed? There also seems to be a bunch of changes that aren't really "fix regression"..?

CMakeLists.txt
```diff
@@ -182,10 +182,10 @@ if(WIN32)
...
 set_target_properties(bitsandbytes PROPERTIES OUTPUT_NAME ${BNB_OUTPUT_NAME})
 if(MSVC)
-    set_target_properties(bitsandbytes PROPERTIES LIBRARY_OUTPUT_DIRECTORY_RELEASE bitsandbytes)
-    set_target_properties(bitsandbytes PROPERTIES LIBRARY_OUTPUT_DIRECTORY_DEBUG bitsandbytes)
-    set_target_properties(bitsandbytes PROPERTIES RUNTIME_OUTPUT_DIRECTORY_RELEASE bitsandbytes)
-    set_target_properties(bitsandbytes PROPERTIES RUNTIME_OUTPUT_DIRECTORY_DEBUG bitsandbytes)
+    set_target_properties(bitsandbytes PROPERTIES LIBRARY_OUTPUT_DIRECTORY_RELEASE "$<1:${CMAKE_SOURCE_DIR}/bitsandbytes>")
+    set_target_properties(bitsandbytes PROPERTIES LIBRARY_OUTPUT_DIRECTORY_DEBUG "$<1:${CMAKE_SOURCE_DIR}/bitsandbytes>")
+    set_target_properties(bitsandbytes PROPERTIES RUNTIME_OUTPUT_DIRECTORY_RELEASE "$<1:${CMAKE_SOURCE_DIR}/bitsandbytes>")
+    set_target_properties(bitsandbytes PROPERTIES RUNTIME_OUTPUT_DIRECTORY_DEBUG "$<1:${CMAKE_SOURCE_DIR}/bitsandbytes>")
```
wkpark, line 188: revert fixes. without this, `cmake -B build` won't work.

CMakeLists.txt
```diff
@@ -128,15 +128,15 @@ elseif(BUILD_MPS)
...
+                COMMAND xcrun metal -c -o "bitsandbytes.air" "${CMAKE_SOURCE_DIR}/${METAL_FILES}"
+                COMMAND xcrun metallib "bitsandbytes.air" -o "${CMAKE_SOURCE_DIR}/bitsandbytes/bitsandbytes.metallib"
                 DEPENDS "${METAL_FILES}"
                 COMMENT "Compiling Metal kernels"
                 VERBATIM)
-    add_custom_target(metallib DEPENDS "bitsandbytes/bitsandbytes.metallib")
+    add_custom_target(metallib DEPENDS "${CMAKE_SOURCE_DIR}/bitsandbytes/bitsandbytes.metallib")
 else()
-    set(LIBSUFFIX "cpu")
+    string(APPEND BNB_OUTPUT_NAME "_cpu")
```
wkpark, line 139: reverted.

CMakeLists.txt
```diff
@@ -128,15 +128,15 @@ elseif(BUILD_MPS)
...
     file(MAKE_DIRECTORY "build")
-    add_custom_command(OUTPUT "bitsandbytes/bitsandbytes.metallib"
-                COMMAND xcrun metal -c -o "build/bitsandbytes.air" ${METAL_FILES}
-                COMMAND xcrun metallib "build/bitsandbytes.air" -o "bitsandbytes/bitsandbytes.metallib"
+    add_custom_command(OUTPUT "${CMAKE_SOURCE_DIR}/bitsandbytes/bitsandbytes.metallib"
+                COMMAND xcrun metal -c -o "bitsandbytes.air" "${CMAKE_SOURCE_DIR}/${METAL_FILES}"
+                COMMAND xcrun metallib "bitsandbytes.air" -o "${CMAKE_SOURCE_DIR}/bitsandbytes/bitsandbytes.metallib"
                 DEPENDS "${METAL_FILES}"
                 COMMENT "Compiling Metal kernels"
                 VERBATIM)
```
wkpark, line 136: to make `cmake -B build`  work

.github/workflows/python-package.yml
```diff
@@ -1,9 +1,9 @@
 name: Python package
 
 on:
-  push: {}
+  push:
+    branches: [ "main" ]
```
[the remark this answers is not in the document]
akx: Having no branch filter here means CI gets run on pushes that don't have a PR too, so you can test things out in your own branch.

CMakeLists.txt
akx: But why?

CMakeLists.txt
akx: Why does `cmake -B build` need to work?

.github/workflows/python-package.yml
```diff
@@ -17,153 +17,231 @@ on:
...
-        build_arch=${{ matrix.arch }}
-        for NO_CUBLASLT in ON OFF; do
-          if [ ${build_os:0:6} == ubuntu ]; then
-            image=nvidia/cuda:${{ matrix.cuda_version }}-devel-ubuntu22.04
-            echo "Using image $image"
-            docker run --platform linux/$build_arch -i -w /src -v $PWD:/src $image sh -c \
-              "apt-get update \
-              && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends cmake \
-              && cmake -DCOMPUTE_BACKEND=cuda -DNO_CUBLASLT=${NO_CUBLASLT} . \
-              && make"
```
wkpark, line 142: cmake normally used with `-B build` option. but is does not.

.github/workflows/python-package.yml
```diff
@@ -17,153 +17,231 @@ on:
...
+    - name: Copy libraries
       shell: bash
       run: |
-        set -ex
-        build_os=${{ matrix.os }}
-        build_arch=${{ matrix.arch }}
-        for NO_CUBLASLT in ON OFF; do
-          if [ ${build_os:0:6} == ubuntu ]; then
-            image=nvidia/cuda:${{ matrix.cuda_version }}-devel-ubuntu22.04
-            echo "Using image $image"
```
wkpark, line 137: using docker seems faster in this case. but this image is deprecated.

akx: That certainly isn't a regression fix?

If the image works and is faster, I think it's a good idea to use it?

wkpark:
> At first glance, this basically reverts a lot of Rickard's work - could you explain why the changes are needed? There also seems to be a bunch of changes that aren't really "fix regression"..?

as I already mentioned, this is a quick and manual revert to fix the current regression state.

akx:
The missing `_cpu` tag might well be a regression, but this PR seems to be putting the Mamba/Conda stuff back in (why?), adding a `setuptools` and `wheel` in the shared-libs step, etc.

Wouldn't it be better to just fix things for the better instead of a "quick revert" that someone will need to fix _again_ later? Since there's a lot of activity in this repo just now, and we're working in a state where we don't really have the luxury to actually properly test anything, I think we should accept the fact that `main` might be broken.

akx:
Could you maybe look into other options than `mamba`? I mean, this repo itself has `install_cuda.py` and `install_cuda.sh` scripts which apparently are meant to install the CUDA Toolkit – maybe we should use those?

.github/workflows/python-package.yml
rickardp: The image patch verison needs to be bumped, the image itself is not deprecated

.github/workflows/python-package.yml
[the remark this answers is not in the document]
rickardp: Regarding branch filters etc, feel free to suggest changes. The decisions were deliberate, but mine, maybe there are other considerations. My idea was that all PRs should be built here, but when creating forks you also get them built and validated. I was planning to add the concurrency part, I deliberately left it out due to the flux in pipelines and it was annoying to have the builds cancelled all the time, but I think it makes sense t

.github/workflows/python-package.yml
```diff
@@ -17,150 +17,365 @@ on:
...
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        arch: [x86_64, aarch64]
-        cuda_version: ['12.1.0']
-        exclude:
-          - os: windows-latest # This probably requires arm64 Windows agents
-            arch: aarch64
+        os: [windows-latest]
+        arch: [x86_64]
+        cuda-version: ['11.8.0', '12.1.1']
```
rickardp, line 124: I deliberately skipped cuda 11 from pipelines not to make them slower before the dust had settled. The main branch is very active at the moment. I don't see a problem adding it, but we must realize that we burn CPU minutes quite fast now. I did not consider this a regression since Python packages were published outside of pipelines before (and still are, we only build them).

rickardp:
Some comment from me, possibly #949 was merged too quick without aligning on the scope. Sorry for not realizing this! May I suggest discussing improvements in the RFCs, maybe in #1031?

I would argue the easiest way to manage CUDA dependencies is through Docker. What is currently lacking is dependabot-managed update for the CUDA versions. I was actually planning to fix this in a follow up PR, it involves creating Dockerfiles for the CUDA images as this is what Dependabot can work wiith. We need to continuously monitor for updates as NVIDIA deprecates CUDA versions fast, and Dependabot is the realistic way to handle this.  The problem here is Windows. While there seems to be some community-based Windows containers, AFAIK hosted runners won't run Windows containers. The community GitHub action does a decent job, but it's slow, owing to the fact that it runs the horribly slow CUDA installer. Some kind of image-based approach (just like Docker  for the Linux builds) would be the best solution

CMakeLists.txt
rickardp: What is the use case here? 

Please correct me if I’m wrong, but I believe this option was meant to support multiple configurations without clearing cache/incremental builds. For this to work the output must never overlap, which it would do here (for example x64 vs arm64).

.github/workflows/python-package.yml
```diff
@@ -17,150 +17,373 @@ on:
...
+      if: ${{ matrix.os == 'ubuntu-latest' && matrix.arch == 'aarch64' }}
+      run: |
+        # Allow cross-compile on aarch64
+        sudo apt-get install -y g++-aarch64-linux-gnu binutils-aarch64-linux-gnu
+
     - name: Setup cmake
       uses: jwlawson/actions-setup-cmake@v1.14
       with:
         cmake-version: '3.26.x'
-    - name: Add msbuild to PATH
```
rickardp, line 44: What is the reason for changing this?

.github/workflows/python-package.yml
```diff
@@ -17,150 +17,373 @@ on:
       - 'pytest.ini'
       - '**/*.md'
   release:
+    branches: [ main ]
     types: [ published ]
 
+concurrency:
+  group: cmake-${{ github.ref }}
```
rickardp, line 24:
```suggestion
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```

.github/workflows/python-package.yml
```diff
@@ -182,10 +405,10 @@ jobs:
         ls -lR output/
         cp output/${{ matrix.os }}/${{ matrix.arch }}/* bitsandbytes/
       # Set up the Python version needed
-    - name: Set up Python ${{ matrix.python-version }}
```
rickardp, line 185: Why hard code the Python version? Even if we split it into a build and test stage, I think defining the version in one place is better

.github/workflows/python-package.yml
```diff
@@ -17,150 +17,373 @@ on:
...
       # Check out dependencies code
     - uses: actions/checkout@v4
       name: Check out NVidia cub
       with:
         repository: nvidia/cub
         ref: 1.11.0
         path: dependencies/cub
-      # Compile C++ code
-    - name: Build C++
+    - name: Setup Mambaforge
```
rickardp, line 152: I am still not convinced mamba is the way to go.

Also, there's a lot of duplication here w.r.t the Docker flow. What is the reason we need to change this?

.github/workflows/python-package.yml
```diff
@@ -17,150 +17,373 @@ on:
...
       with:
         name: shared_library_${{ matrix.os }}_${{ matrix.arch }}
         path: output/*
-        retention-days: 7
+
   ##
-  # This job matrix builds the CUDA versions of the libraries for platforms that support CUDA (Linux x64/aarch64 + Windows x64)
+  # This job matrix builds the CUDA versions of the libraries for platforms that support CUDA (Windows)
   ##
   build-shared-libs-cuda:
```
rickardp, line 118: This is a windows only flow now, so it should probably be called build-shared-libs-cuda-windows. But TBH I don't see why it has to be split out from the Linux flow

.github/workflows/python-package.yml
matthewdouglas: Same, it seems overly complicated for not much obvious benefit to me. I like to be able to do something simple like `pip install -e .[dev]` and use the cache on it.

Titus-von-Koeller:
Hey @wkpark, @akx, @matthewdouglas, @rickardp,

Yes agreed, #949 was merged a bit hasty, due to an misunderstanding and too quick trigger finger on my side. Sorry for that! And thanks everyone for the collective cleanup / hotfix actions that ensued and got everything back to working order.

Based on your discussion above, I'm now not sure if this PR is still WIP or if you all agree that it's ready for review and merge? If not, what do we need to still implement or agree on in order to move forward?

.github/workflows/python-package.yml
```diff
@@ -187,16 +264,23 @@ jobs:
...
       shell: bash
-      run: python -m build .
+      run: |
+        python -m build . --wheel
+        # fix wheel name
+        if [ "${{ matrix.arch }}" = "aarch64" ]; then
+            o=$(ls dist/*.whl)
+            n=$(echo $o | sed 's@_x86_64@_aarch64@')
+            [ "$n" != "$o" ] && mv $o $n
+        fi
```
akx, line 274: This looks like a hack? Why is the wheel named x86_64... if it contains aarch64 material? IOW, are you sure it contains aarch64 material if it's named wrong?

rickardp:
> Hey @wkpark, @akx, @matthewdouglas, @rickardp,
> 
> 
> 
> Yes agreed, #949 was merged a bit hasty, due to an misunderstanding and too quick trigger finger on my side. Sorry for that! And thanks everyone for the collective cleanup / hotfix actions that ensued and got everything back to working order.
> 
> 
> 
> Based on your discussion above, I'm now not sure if this PR is still WIP or if you all agree that it's ready for review and merge? If not, what do we need to still implement or agree on in order to move forward?

TBH I thought this PR was abandoned and most of the stuff had been broken out to other PRs.

I'll have another look as I am not completely sure what issues remain that this PR solves

rickardp:
> use ilammy/msvc-dev-cmd@v1.13.0 instead of microsoft/setup-msbuild
microsoft/setup-msbuild is used for msbuild. ilammy/msvc-dev-cmd@v1.13.0

This is already on master I think

rickardp:
>  fix wheel names for aarch64

> drop artifact retention-days: it could be configured Settings->Actions->General.

> fail-fast: false restored.

I suggest we make these separate PRs. I think retention days we want to set separately per artifact type as some artifacts (wheels) are more useful than others (.so files). But agreed these need tuning

.github/workflows/python-package.yml
matthewdouglas: @akx So to me it looks like the lib is built on aarch64 via Docker, but this step of packaging the artifacts for a wheel is being executed on an x86-64 host always. 

I agree it seems hacky. Maybe using something like `wheel tags $o --platform-tag=manylinux2014_aarch64` will feel a little better.

wkpark:
> Hey @wkpark, @akx, @matthewdouglas, @rickardp,
> 
> Yes agreed, #949 was merged a bit hasty, due to an misunderstanding and too quick trigger finger on my side. Sorry for that! And thanks everyone for the collective cleanup / hotfix actions that ensued and got everything back to working order.
> 
> Based on your discussion above, I'm now not sure if this PR is still WIP or if you all agree that it's ready for review and merge? If not, what do we need to still implement or agree on in order to move forward?

several PRs have been already extracted and merged.
We can consider this PR as a reference or ideas, I do not intend to get it merged without further discussion or agree.

CMakeLists.txt
matthewdouglas: Some IDEs behave in a way where this may be needed, so I'm going to extract out into a separate PR.

Titus-von-Koeller:
Hey @wkpark,

it's been a while :) Thanks for your amazing work on BNB at the time.

I'm not sure what to do with this PR. Should we just close it or do you think there are important ideas here that we should still implement? In that case we would want to track them somehow and link to this PR, but close it as it's just for reference and can be referenced even when closed.

We would appreciate your input!

matthewdouglas:
Hi all,

I will close this PR as it is stale and many of the ideas in it have been implemented in separate PRs. We've had a stable build workflow that has been used for several releases now. If there's any remaining improvements let's open new PRs.

Thanks!

PR #1038, opened 2024-02-05
HOTFIX: Fix regression (cpu fix)

wkpark:
- [x] fix critical regression bug: add `_cpu` tag correctly.
- [x] use library extension correctly for macos. (from PR #1018)


bitsandbytes/cuda_setup/main.py
```diff
@@ -37,7 +37,7 @@
     # we have libcudart.so.11.0 which causes a lot of errors before
     # not sure if libcudart.so.12.0 exists in pytorch installs, but it does not hurt
     CUDA_RUNTIME_LIBS = ["libcudart.so", "libcudart.so.11.0", "libcudart.so.12.0", "libcudart.so.12.1", "libcudart.so.12.2"]
-    DYNAMIC_LIBRARY_SUFFIX = ".so"
+    DYNAMIC_LIBRARY_SUFFIX = { "Darwin": ".dylib", "Windows":".dll" }.get(platform.system(), ".so")
```
akx, line 40: Using a dict for this is a great idea, but since we check for `if platform.system() == 'Windows'` above, this should be simply moved out of the `else` branch (and the `DYNAMIC_LIBRARY_SUFFIX = ".dll"` above in the Windows branch then eliminated).

CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ elseif(BUILD_MPS)
                 VERBATIM)
     add_custom_target(metallib DEPENDS "bitsandbytes/bitsandbytes.metallib")
 else()
-    set(LIBSUFFIX "cpu")
+    string(APPEND BNB_OUTPUT_NAME "_cpu")
```
akx, line 139: 👍 Looks like `LIBSUFFIX` isn't used (unless it's a CMake thing I don't know about).

akx (changes requested):

CMakeLists.txt
rickardp: No, this was a merge error made by me the other day. I used LIBSUFFIX earlier, this was changed to BNB_OUTPUT_NAME, I missed this line

bitsandbytes/cuda_setup/main.py
```diff
@@ -28,16 +28,15 @@
 
 from .env_vars import get_potentially_lib_path_containing_env_vars
 
-if platform.system() == 'Windows':  # Windows
+DYNAMIC_LIBRARY_SUFFIX = { "Darwin": ".dylib", "Windows": ".dll", "Linux": ".so"}.get(platform.system(), ".so")
+if DYNAMIC_LIBRARY_SUFFIX == ".dll":  # Windows
```
akx, line 32: I think
```suggestion

if platform.system() == 'Windows':  # Windows
```
would still be clearer :)

akx (approved):
🎉

rickardp:
We should merge this as soon as possible IMHO. Currently the installation instructions does not work until this is merged

younesbelkada (approved):
Thank you !

PR #1044, opened 2024-02-06
restore BUILD_* options

wkpark:
* support both `BUILD_*` and `COMPUTE_BACKEND` options

internally `COMPUTE_BACKEND=cpu|cuda|mps` use `BUILD_*` options, so we can support old options with minor fix.

- use `COMPUTE_BACKEND` if it is available. (default value is "") => "`cpu`" build will be selected.
- the default build is "`cpu`" without any given options. (previously default is "`BUILD_CUDA=ON`")
- ~~with this fix, we can restore the old `Makefile` behavior.~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ set(METAL_FILES csrc/mps_kernels.metal)
 # C++ sources are always included
 list(APPEND SRC_FILES ${CPP_FILES})
 
-set(COMPUTE_BACKEND "cpu" CACHE STRING "The compute backend to use (cpu, cuda, mps)")
+set(COMPUTE_BACKEND "" CACHE STRING "The compute backend to use (cpu, cuda, mps)")
 set_property(CACHE COMPUTE_BACKEND PROPERTY STRINGS cpu cuda mps)
 option(PTXAS_VERBOSE "Pass through -v flag to PTX Assembler" OFF)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -23,39 +23,47 @@ set(METAL_FILES csrc/mps_kernels.metal)
...
-set(COMPUTE_BACKEND "cpu" CACHE STRING "The compute backend to use (cpu, cuda, mps)")
+set(COMPUTE_BACKEND "" CACHE STRING "The compute backend to use (cpu, cuda, mps)")
 set_property(CACHE COMPUTE_BACKEND PROPERTY STRINGS cpu cuda mps)
 option(PTXAS_VERBOSE "Pass through -v flag to PTX Assembler" OFF)
 
 if(APPLE)
   set(CMAKE_OSX_DEPLOYMENT_TARGET 13.1)
 endif()
 
+option(BUILD_CUDA "Build bitsandbytes with CUDA support" OFF)
```
rickardp, line 34: This allows specifying both MPS and CUDA at the same time, which will fail as it is now. What is the expected outcome?

CMakeLists.txt
```diff
@@ -23,39 +23,47 @@ set(METAL_FILES csrc/mps_kernels.metal)
...
-message(STATUS "Building with backend ${COMPUTE_BACKEND}")
+if("${COMPUTE_BACKEND}" STREQUAL "")
+else()
+    message(STATUS "Building with backend ${COMPUTE_BACKEND}")
+endif()
 
-if(${COMPUTE_BACKEND} STREQUAL "cuda")
+if("${COMPUTE_BACKEND}" STREQUAL "cuda")
     if(APPLE)
         message(FATAL_ERROR "CUDA is not supported on macOS" )
```
rickardp, line 47: This check is ignored when specifying -DBUILD_CUDA=ON

CMakeLists.txt
wkpark: previously implemented `CMakeLists.txt` does not do much and, `BUILD_MPS` even not work currently.

CMakeLists.txt
```diff
@@ -23,39 +23,47 @@ set(METAL_FILES csrc/mps_kernels.metal)
...
 set_property(CACHE COMPUTE_BACKEND PROPERTY STRINGS cpu cuda mps)
 option(PTXAS_VERBOSE "Pass through -v flag to PTX Assembler" OFF)
 
 if(APPLE)
   set(CMAKE_OSX_DEPLOYMENT_TARGET 13.1)
 endif()
 
+option(BUILD_CUDA "Build bitsandbytes with CUDA support" OFF)
+option(BUILD_MPS "Build bitsandbytes with MPS support" OFF)
+option(NO_CUBLASLT "Disable CUBLAS" OFF)
```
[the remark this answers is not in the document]
wkpark: `NO_CUBLASLT` was old `Makefile` stuff and previously implemented `CMakeLists.txt` does not break its old behavior.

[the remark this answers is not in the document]
rickardp: Is it important not to break the old Makefile flags?

wkpark:
The documentation for the command has been updated in PR #1047, close this issue.

PR #1048, opened 2024-02-06
Updated documentation on compiling from source

### LINKED ISSUE #1040: Compilation doc page is obsolete
Currently the compilation guidelines in https://huggingface.co/docs/bitsandbytes/main/en/compiling are obsolete as the makefile has been deleted due to the cmake refactor. It would be great if one could re-write that documentation page so that it is up-to-date with the current way of compiling bnb from source

cc @Titus-von-Koeller

rickardp:
For cmake update

As per issue #1040

Fixes #1040


younesbelkada (approved):
Great work, thank you !

Issue #1043, opened 2024-02-06
Distribute pip wheels for the architecture they are built for


MatthiasKohl:
### System Info

Any platform other than x86_64

### Reproduction

On e.g. aarch64, `pip install bitsandbytes`.

### Expected behavior

`pip install bitsandbytes` will succeed at first on aarch64 because the distribution on PyPi has the `any` architecture.
However, the distributed library is built for x86_64, and this cannot work.

It would be expected that there is a pip distribution for each architecture supported by bitsandbytes.

matthewdouglas:
Hi @MatthiasKohl,

At the moment the only supported platform for the PyPI wheels is manylinux1_x86_64. However, please see #997 and the ongoing discussion in #1032 where this is being addressed.

Status overview:
* Windows x86_64 + CUDA support is well underway. The `main` branch is buildable!
* Linux aarch64 + CUDA builds are part of the GH Actions workflows, though I am not sure of status.
* Support for macOS arm64 + MPS is discussed in #1020.
* Support for CPU-only x86_64 (Windows/Linux) and aarch64 (Linux) is discussed in #1021
* Intel has been working on backends for their CPUs and GPUs:  #894, #898
* Draft for Linux x86_64 + AMD ROCm is #756

rickardp:
@MatthiasKohl Can you try the wheels built here?

https://github.com/TimDettmers/bitsandbytes/actions/runs/7792455224

They are built for Linux aarch64, but I don't have an Arm device that can run Cuda so I never tested them. It would be very valuable to get some feedback on how they run. I believe the source level changes were tested on a Jetson device earlier

wkpark:
Linux cuda arm64 build is not arm64. it is just x64_86. (Linux cpu arm64 is just fine.)

rickardp:
> Linux cuda arm64 build is not arm64. it is x64_86. (Linux cpu arm64 is just fine.)

https://github.com/TimDettmers/bitsandbytes/blob/88ab630315d9a79973302182d79653b1dfa0918a/.github/workflows/python-package.yml#L138

The cuda versions are indeed built for arm64. Or are you saying that the arm64 image produces x64 output

matthewdouglas:
@rickardp 
> The cuda versions are indeed built for arm64. Or are you saying that the arm64 image produces x64 output

[...]

PR #1050, opened 2024-02-07
Fix cross compilation on linux

rickardp:
As mentioned in #1043, cross compilation was missing the necessary cmake flags to actually build for arm64


Titus-von-Koeller:
Thanks @rickardp!

PR #1051, opened 2024-02-07
Add concurrency lock to Python wheel workflow

rickardp:
To not waste precious build minutes when modifying PRs frequently.

As added in #1035 

Also fixes an unnecessary path filter, but possibly it would be better to turn the filter around later.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,9 +15,12 @@ on:
       - 'setup.py'
       - 'pyproject.toml'
       - 'pytest.ini'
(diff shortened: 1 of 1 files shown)


wkpark:
original author is @akx by PR #1011 by commit https://github.com/TimDettmers/bitsandbytes/pull/1011/commits/f718770592c7541d703b8a82b696d365fb1d7412#1055 
thank you!

Titus-von-Koeller:
Cool! Always better to have shorter iteration times :)

PR #1052, opened 2024-02-07
Define CUDA versions for workflows centrally

rickardp:
This PR defines CUDA versions to be built in workflows centrally.

It uses a Dockerfile as a source of truth, as this can be updated automatically by Dependabot.

Currently, the Dockerfile is not used for building the compiler images, but we could do this in the future if we want to speed up buiilds (by caching on ghcr.io and e.g stripping static libs that we don't use, as they constitute about 50% of the image size), but the bottleneck is Windows anyway so this is not critical at the moment


scripts/get_docker_info.sh
```diff
@@ -0,0 +1,31 @@
...
+tag=$(grep "AS $1\$" $dockerfile | sed -e 's/FROM *//' -e 's/ *AS .*//')
+if [ -z "$tag" ]; then
+  echo "Target $1 not defined" >> /dev/stderr
+  exit 1
+fi
+if [ "$version_only" = "1" ]; then
+   echo $tag | sed -e 's/.*://'
+else
+   echo $tag
+fi
```
akx, line 31: 😳 Could we write this in Python, pretty please? 😂 

Something like

```python
import argparse
import re

from_line_re = re.compile(r"FROM\s+(?P<image>\S+)\s+AS\s+(?P<target>\S+)")


def find_image_in_dockerfile(dockerfile, target):
    with open(dockerfile) as f:
        for line in f.readlines():
            if (m := from_line_re.match(line)) and m.group("target") == target:
                return m.group("image")
    raise ValueError(f"Target {target} not defined in {dockerfile}")


def main():
    ap = argparse.ArgumentParser()
    ap.add_argument("dockerfile", required=True)
    ap.add_argument("target", required=True)
    ap.add_argument("-v", "--version-only", action="store_true")
    args = ap.parse_args()
    image = find_image_in_dockerfile(args.dockerfile, args.target)
    if args.version_only:
        image = image.rpartition(":")[-1]
    print(image)
```
, I think...

akx (commented):
Shell script heebie-jeebies within 😁

scripts/get_docker_info.sh
rickardp: No objection from me 😃   I just took this script off my library of useful stuff TBH

matthewdouglas:
Hi,

I am closing this as it is stale and we have a build workflow where this is defined in one place here: https://github.com/bitsandbytes-foundation/bitsandbytes/blob/d47553356f59c9e3e27059cd5872679e30293ece/.github/workflows/python-package.yml#L75

PR #1053, opened 2024-02-07
Skip checkout nvidia cub

rickardp:
This dependency is not needed anymore for building with CUDA, so there is no need to clone this repo anymore


younesbelkada (approved):
Thanks so much for the cleanup !

PR #1061, opened 2024-02-13
Fix race condition in kEstimateQuantiles

pnunna93:
Hi, there is a race condition in kEstimateQuantiles kernel and it needs additional synchronization step. The kernel is giving unexpected results on rocm because of this, this PR adds the sync step to avoid it.


Titus-von-Koeller:
Thanks for raising this and the quick fix!

PR #1072, opened 2024-02-21
[docs] Reorganize docs

stevhliu:
This PR reorganizes the current structure of the docs and consolidates all the installation instructions (compiling from source and non-PyTorch CUDA versions) into a single doc so users can easily find everything.

cc @Titus-von-Koeller


Titus-von-Koeller:
Strange that the workflow didn't post a link to the docs page of the PR. It also doesn't exist actually. [A functioning example would be here](https://moon-ci-docs.huggingface.co/docs/bitsandbytes/pr_1012/en/index), whereas the page of this PR doesn't exist: https://moon-ci-docs.huggingface.co/docs/bitsandbytes/pr_1072/en/index

Titus-von-Koeller:
Ok, rerunning the workflow solved it. Really strange, cc @younesbelkada, that there was a silent failure here, as the publishing workflow had displayed as green.

younesbelkada:
Hmmm interesting, however I am able to see the changes in https://moon-ci-docs.huggingface.co/docs/bitsandbytes/pr_1072/en/index - i think all is good no?

Titus-von-Koeller:
Thanks, this looks great. I just made a small tweak. I'll see if everything compiled correctly after going to sports and then merge it.

Thanks again for your work. Really great to have your collaboration on further improving the docs 🤗

Titus-von-Koeller:
Thanks for your work! Made some small tweaks in the existing text that I became aware of when reviewing everything. I think this restructuring definitely improves the status quo. Looking forward to the next steps as well :)

Titus-von-Koeller:
> Hmmm interesting, however I am able to see the changes in https://moon-ci-docs.huggingface.co/docs/bitsandbytes/pr_1072/en/index - i think all is good no?

@younesbelkada Yeah, the thing is that the link gave back a 404 at first, even though the workflow for publishing had been green for a couple of days. The docs and the automatic post here in the PR linking to it only appeared when I triggered a rerun of the action.

I think it's a bit sup-optimal that the job can fail silently. At a quick glance I couldn't even figure out where the code is that is responsible to posting back to the PR. Maybe we should take a look when we get around to it.

PR #1076, opened 2024-02-22
(docs) Clarify Python and CUDA Toolkit version requirement

matthewdouglas:
* Clarifies in the documentation that Python 3.8 is supported
* Updates CUDA Toolkit download URLs for newer patch releases
* Minor improvements to some error log messages in CUDA setup

N.B. I haven't considered the maximum Python range; not sure if anything special needs to be done for upcoming Python 3.13.


Titus-von-Koeller:
Thanks @matthewdouglas that was super helpful! Very much appreciated.

PR #898, opened 2023-12-05
Enable common device abstraction for 8bits/4bits

jianan-gu:
As one of the plans in the RFC https://github.com/TimDettmers/bitsandbytes/issues/894. This PR intends to add a device abstraction that **allows the supports of non-CUDA devices to be added into bitsandbytes (for 8bits/4bits functionality) more commonly and easily**. 

To create a device backend abstraction, this PR creates a common backend folder, which contains the key kernel interfaces class to implement by each backend, a backend registration interface to add new device backends, and a kernel dispatching mechanism.
```
Key interfaces from bitsandbytes.functional  that are for the scope of 8bits and 4bits:
| F.igemmlt | F.double_quant| F.mm_dequant| F.transform| F.extract_outliers| F.quantize_4bit| F.dequantize_4bit |
```
image

The common backends will further be used in bitsandbytes.functional to implement the 8bits/4bits functionality.

Currently, the backend only contains CUDA as it is without this PR.

PR #1077, opened 2024-02-23
DRAFT: allow for multiple hardware backends

Titus-von-Koeller:
Due to Tim's extensive feedback and request to implement an alternative approach based on his preferences, this is a (rough) draft of my understanding of his solution sketch.

This isn't complete, but I'm publishing it for early feedback: At this point especially from Tim, as I want to verify that I didn't misunderstand anything before getting feedback from everyone else.

It's late and I'll have to continue with the rest in the morning (meaning that this is far from complete).


bitsandbytes/functional.py
```diff
@@ -771,33 +862,72 @@ def quantize_blockwise(
...
-    if A.device.type != 'cpu':
+    if A.device.type != "cpu":
         assert blocksize in [4096, 2048, 1024, 512, 256, 128, 64]
         cblocksize = ct.c_int32(blocksize)
         prev_device = pre_call(A.device)
         code = code.to(A.device)
         is_on_gpu([code, A, out, absmax])
         if A.dtype == torch.float32:
-            lib.cquantize_blockwise_fp32(get_ptr(code), get_ptr(A), get_ptr(absmax), get_ptr(out), cblocksize, ct.c_int(A.numel()))
+            lib.cquantize_blockwise_fp32(
```
akx, line 872: Feels like these calls should also be delegated to a backend?

bitsandbytes/functional.py
```diff
@@ -2016,73 +2491,25 @@ def mm_dequant(
...
-        assert rowidx.numel() == nnz
-        assert colidx.numel() == nnz
-
-        self.rows = rows
-        self.cols = cols
-        self.nnz = nnz
-        self.rowidx = rowidx
-        self.colidx = colidx
-        self.values = values
+@deprecated
```
akx, line 2510: Why does this (and the other similar trampolines) need to be `deprecated`? 🤔

pyproject.toml
```diff
@@ -8,6 +8,22 @@ src = [
...
+ignore = [
+    "B007",    # Loop control variable not used within the loop body (TODO: enable)
+    "B028",    # Warning without stacklevel (TODO: enable)
+    "E501",    # Supress line-too-long warnings: trust yapf's judgement on this one.
+    "E701",    # Multiple statements on one line (TODO: enable)
+    "E712",    # Allow using if x == False, as it's not always equivalent to if x.
+    "E731",    # Do not use lambda
+    "F841",    # Local assigned but not used (TODO: enable, these are likely bugs)
+    "RUF012",  # Mutable class attribute annotations
+    "ISC001",  # String concatination warning: may cause conflicts when used with the formatter
```
akx, line 24:
```suggestion
    "ISC001",  # String concatenation warning: may cause conflicts when used with the formatter
```

... but this shouldn't be added anyway IMO, the single-line concatenations should be fixed (and Ruff can autofix them).

bitsandbytes/backends/_base.py
```diff
@@ -0,0 +1,143 @@
...
+        raise NotImplementedError
+
+    def extract_outliers(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def igemmlt(self, *args, **kwargs):
+        raise NotImplementedError
+
+    def mm_dequant(self, *args, **kwargs):
+        raise NotImplementedError
```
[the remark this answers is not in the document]
Titus-von-Koeller: > am a great proponent of making some functions at least partially kwarg-only

Totally agree, I also like locking APIs with at least the `*`, like `def fn(positional_only, /, keyword_and_positional, *, keyword, only):`..

Yes, the args, kwargs are only as scaffolding to make a point that these methods still come and will be part of the API (based on Tim's input).

rickardp:
My main concern with this vs the original PR is the singleton backend. It feels like a fundamental architectural decision. I say this because I think in addition to the actual code changes, assumptions to this will be made in many places (as well as in code using this library, making backtracking on this decision "breaking").

Is it certain that this is not coming back to bite us in the end? Are we sure we do not want to support heterogeneous device setups?

I would gravitate towards suggesting following the torch device as was suggested in the other PR, but maybe I do not understand the cons of a multiple backends approach.

matthewdouglas:
> Edit: I see now this comment from @TimDettmers:
> 
> > There is also currently the problem that the backend design breaks paged buffers that have a CPU device type but need to be executed on a CUDA device. I think this can be fixed by assigning a singleton backend upon initialization rather than dynamically checking the device type.
> 
> Could we solve this by sending in a device parameter instead of or as an optional complement to inferring from the device? Or is the concern a breaking API change here?

I think there's already a need to track a device index, so this seems feasible to add device type, unless there's something I'm misunderstanding. E.g. I already see the signature `void cprefetch(void *ptr, size_t bytes, int device)` along with a `page_deviceid` property. Would we just be able to look at when `tensor.is_paged` to resolve that?

It may also just make sense to start out by assuming we're using a CUDA or (maybe ROCm?) device with paged optimizers. It doesn't make sense for MPS or CPU backends. Intel GPU is the question mark for me on that. But even then, is another option to globally select a backend just for this? I.e. `GlobalOptimManager.get_instance().set_backend("ipex")`

rickardp:
> It doesn't make sense for MPS or CPU backends. Intel GPU is the question mark for me on that.

Paged optimizers is a pointless feature for all systems with true unified memory. 

> I.e. GlobalOptimManager.get_instance().set_backend("ipex")

On the other hand, even with NVIDIAs "virtual UMA", wouldn't it be possible to first allocate the tensor on the device it is meant to go to, then being able to access it from the CPU?

Or is the problem is that PyTorch does not support it which is why this would be needed? TBH I've never used these new CUDA features so I was just looking at the docs (https://developer.nvidia.com/blog/unified-memory-cuda-beginners/)

TimDettmers:
## Cross-posted reply to #898

> 1) Device common abstraction interface

The four sub-interfaces look good to me. We probably want to create higher-level abstractions for 4-bit quantization/dequantization and some other functions should be deprecated such as percentile clipping, and the non-tensor-core 8-bit matrix multiplication.

It seems more people like the tensor-driven dispatch. I find it difficult to read and debug and it can lead to more accidental bugs compare to a singleton. There are also further problems. For example, for paged tensor, the device is `cpu` but it needs to be executed on a `cuda` device. We can create special branching functions for this case. But then we already have a special branch for AMD and paged tensors. A singleton would have no branching at all and bugs can be avoided it the user forgot to push the tensor to the correct device. I think a singleton is much superior from reducing bugs and enhancing the user experience.

For a tensor type dispatch, how would you handle the case where the user uses the wrong device? If a tensor is on the CPU, I guess it would silently be executed on the CPU (if supported). This would be slow without the user knowing that something is wrong and we have no way of detecting if something is wrong because we just dispatching based on device type.

> 2) Device-specific functions/class

Some functions specific to the device would be specializations in the base class and would not be provided in the interface. So these should be "private" functions for the base class.

I think a runtime singleton would be overkill. The case that someone has multiple accelerators of different device type would be very rare (probably only developers have this). We should have a special manual hack for this rather than degrading the overall system for this use-case. So, a static one-time singleton would be preferable from my view.

> 3) Device common tool functions/classes

Agreed. I am not sure if all device type would work the same way. For example, does `get_ptr()` generalize to all devices? It might be a special case though and if something is wrong we can fix it later. So I think moving everything to something like `utils.py` might work well.

> I've noticed that the other implementation https://github.com/TimDettmers/bitsandbytes/pull/1077 which has the same scope has just started to be reviewed. Have you decided which design will be applied? It would be nice if we could get your feedback soon, we would love to help make progress for both designs. Thx!

The PR #1107 was based on a discussion that I had with @Titus-von-Koeller. It was my preferred view at the time, but we should discuss what would be best for everyone and decide based on that.

## Cross-posted reply to #1077

> I think there's already a need to track a device index, so this seems feasible to add device type, unless there's something I'm misunderstanding. E.g. I already see the signature void cprefetch(void *ptr, size_t bytes, int device) along with a page_deviceid property. Would we just be able to look at when tensor.is_paged to resolve that?

I think this might be the way to go. If we use a singleton we do not need this, but for a dispatch scenario we need to do this and it is probably the best solution to do an if-else on `tensor.is_paged`.

> On the other hand, even with NVIDIAs "virtual UMA", wouldn't it be possible to first allocate the tensor on the device it is meant to go to, then being able to access it from the CPU?

This would be more general, and we could even abstract it to disk -> unified memory architectures like Apple silicon. It requires more work, but would be a great feature for Apple silicon to work with larger models on smaller Macbooks and the iPhone etc. The main problem here is that it requires quite a bit of work, but I think it could well be worth it to create this abstraction as it is more general and quite powerful (it can be used beyond optimizers, for example also for matrix multiplication / weights).

> My main concern with this vs the original PR is the singleton backend. It feels like a fundamental architectural decision. I say this because I think in addition to the actual code changes, assumptions to this will be made in many places (as well as in code using this library, making backtracking on this decision "breaking").. Is it certain that this is not coming back to bite us in the end? Are we sure we do not want to support heterogeneous device setups?

I think this is the main advantage of non-static singleton. I would optimize for the user-side rather than developer-side. A singleton can prevent a lot of accidental bugs and enhances the user-experience for cases where the user forgot to move to the correct device. This can become a problem in particular if we support CPU devices.

__I think the main discussion point should be which architectural design will help with the best user experience. That is really what I want to optimize for with the design. Lets discuss.__

PS: @Titus-von-Koeller had more details on the sub-interfaces and what will be deprecated and we will post soon.

jiqing-feng:
Hi @Titus-von-Koeller . Can you share the user behavior here? Like how to assign the singleton or device.

If environmental parameters or some global variables can pass the singleton, how should I use assisted_decoding, which has 2 models, one running on the GPU and another running on the CPU?

If we can assign a singleton on the object like `model.set_singleton(device)`, then what's the difference btw `model.to(device)`.

Do you have any method that will not break any user behavior so it can be easily applied by users?

akx:
>  I would optimize for the user-side
[...]
>  A singleton can prevent a lot of accidental bugs

What accidental bugs are you referring to? The user would not be interacting with the backend object by hand in any regular use case.

matthewdouglas:
Not to be overly pedantic, but I wonder if we need to differentiate between backend and device, too. #747 has me thinking even more about this.

We have to consider what the PyTorch build supports, which devices are available, and which libraries/compute backend the user wants to (or should by default) use for those devices. I think there should be sensible defaults based on the hardware available, but maybe additional choices via a Python API and/or env variables for users to have more control if needed.

PyTorch is built for CUDA, ROCm, or MPS. We're not going to combine these.

Some sensible mutually exclusive default backends based on PyTorch build and library availability, in a priority order:

1. macOS arm64, MPS
2. Linux aarch64, CUDA
3. Linux/Windows x86-64, CUDA or ROCm

We can do simple tests for those and pick at most one.

Next up is Intel GPUs (IPEX). That might depend on torch version, as it seems like torch 2.3.0+ will have `torch.xpu` built-in, while older will take from the `ipex` package. In this case we would load an IPEX backend, maybe with a SYCL binary to go with it from #747. It's not clear to me if this can coexist with CUDA/ROCm.

From there is XLA/PJRT (see #1119). My understanding there is limited, but I think AWS Tranium/Inferentia can be enabled through XLA and `torch-neuronx`.

For Intel Gaudi, the device type should end up being `hpu`, and we'd see a PyTorch package that is built for that, so I'm not sure that would ever be combined with other accelerators.

We should also always have some kind of minimal CPU implementation to fallback to, whether it's hyper-optimized or just naively implemented and missing some ops.

I think it is reasonable to detect the support for these devices and load any python modules, binaries, etc, and then make dispatch decisions based on type of device (`mps`, `cuda`, `xpu`, `xla`, `hpu`, `cpu`), device index, and some supplemental information, i.e. `is_paged`.

Then within some of these maybe there's further options based on build, env vars, hardware support, etc. Do we want to let you do something odd like run with torch + ROCm but invoke SYCL kernels? I'll try to elaborate soon, but in general I think there's a balance to be made here.

PR #1078, opened 2024-02-23
Migrate build data to pyproject.toml

matthewdouglas:
* Moves most metadata to `pyproject.toml`, conforming with [PEP 517](https://peps.python.org/pep-0517/), [PEP 518](https://peps.python.org/pep-0518/), and [PEP 621](https://peps.python.org/pep-0621/)
* Removes `requirements.txt` files (but not the conda `environment.yml` ones just yet)
* Updates docs to instruct on usage of editable install for development (see: [PEP 660](https://peps.python.org/pep-0660/))
* Updates workflow to skip unnecessary dependency installation for building wheels


.github/workflows/python-package.yml
```diff
@@ -177,9 +177,9 @@ jobs:
     - name: Install build package
       shell: bash
       run: pip install build
-    - name: Install Python test dependencies
-      shell: bash
-      run: pip install -r requirements-ci.txt
+    #- name: Install Python test dependencies
+    #  shell: bash
+    #  run: pip install -r requirements-ci.txt
```
akx, line 182: Dead code?

docs/source/installation.mdx
```diff
@@ -51,10 +51,9 @@ You'll need to build bitsandbytes from source. To compile from source, you need
 
 ```bash
 git clone https://github.com/TimDettmers/bitsandbytes.git && cd bitsandbytes/
-pip install -r requirements-dev.txt
 cmake -DCOMPUTE_BACKEND=cuda -S .
 cmake --build . --config Release
-python -m build --wheel
+pip install .
```
akx, line 56: This is a different thing. `python -m build --wheel` will build a wheel and not install it. `pip install .` builds a throwaway wheel and installs it.

pyproject.toml
```diff
@@ -1,7 +1,80 @@
 [build-system]
-requires = [ "setuptools", "wheel" ]
+requires = ["setuptools >= 63.0.0"]
```
akx, line 2: At some point, maybe not now, we might want [`hatchling`](https://hatch.pypa.io/1.9/) as a build backend instead, because it's easier to hook into, and we could easily ensure the resulting wheel has the binary component included.

.github/workflows/python-package.yml
matthewdouglas: Thanks, removed this.

pyproject.toml
matthewdouglas: Agreed. I've only worked with Python for a few years now, but my general impression is that the packaging ecosystem leaves much to be desired. No idea what the best tool is, but for right now personally I favor smaller incremental changes over bigger ones.

docs/source/installation.mdx
Titus-von-Koeller: > In general though I think the docs on that page could be expanded a little bit and offer more information.

Yes, I think that would be good. Do you have anything in specific in mind that you feel must be included?

matthewdouglas: A few things that come to mind that I may be able to contribute to:
* Explain there's separate builds for NO_CUBLASLT, CUDA, and CPU-only
** _Aside_ I'm wanting to get rid of the need for the NO_CUBLASLT build, future PR may happen for that.
* Specify compilers we know to work
** E.g. for MSVC, we've been able to build with VS2022 (MSVC 193X), but as far as I am aware have not validated older versions.
** To that point, we haven't built against a lot of CUDA toolkit versions yet. CUDA 11.6 appears to be the minimum for VS2022
** For gcc on x86-64, the bare minimum is probably 6.x?
** I haven't verified but I think MSVC 1934 will be required for 12.0+.
* Should maybe explain how to set up build environment on Windows, with the [Developer Command Prompt](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022) probably being easiest.
* Specify some hardware constraints on CPU: e.g. we expect AVX2 support on x86-64 (at least for Windows build, but would expect the same on Linux in the future, i.e. we expect minimum `x86-64-v2`)
* Minimum OS requirements
** For Windows, this would be Windows Server 2019+, Windows 10+, x86-64 only.
** Not clear to me for macOS, but still WIP

pyproject.toml
```diff
@@ -1,7 +1,80 @@
...
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Topic :: Scientific/Engineering :: Artificial Intelligence"
+]
+dependencies = [
+    "torch",
```
rickardp, line 41: Don't we need any version constraint at all? Maybe at least a minimum version?

pyproject.toml
```diff
@@ -1,7 +1,80 @@
...
+    "numpy"
+]
+
+[project.optional-dependencies]
+benchmark = ["pandas", "matplotlib"]
+docs = ["hf-doc-builder", "black"]
+dev = [
+    "bitsandbytes[test]",
+    "build",
```
rickardp, line 50: In anticipation of a new build system, could we set some basic version constraints. Maybe minimum version what used to be in requirements.txt and up to next major? If we trust semantic versioning

pyproject.toml
```diff
@@ -1,7 +1,80 @@
...
+test = [
+    "einops~=0.6.0",
+    "lion-pytorch==0.0.6",
+    "pytest<8",
+    "pytest-cases",
+    "pytest-sugar",
+    "scipy",
+    "transformers"
+]
+triton = ["triton; sys_platform=='linux' and platform_machine=='x86_64'"]
```
rickardp, line 64: This is a new optional dependency. Do we use it in this codebase?

pyproject.toml
```diff
@@ -1,7 +1,80 @@
...
-requires = [ "setuptools", "wheel" ]
+requires = ["setuptools >= 63.0.0"]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "bitsandbytes"
+dynamic = ["version"]
+description = "k-bit optimizers and matrix multiplication routines."
+authors = [{name="Tim Dettmers", email="dettmers@cs.washington.edu"}]
+requires-python = ">=3.8"
```
rickardp, line 10: We don't actually build/test for 3.8 in the workflow. Should we specify 3.9 as a minimum version, or verify for 3.8 in the workflow?

pyproject.toml
matthewdouglas: I do think we should have at least minimums, but I'm not sure what they should be. I took these runtime dependencies out of `setup.py` just as they were.

If it were up to me, I would pin `torch>=1.13.0` on the low end for all platforms/backends assuming everything works. Why?

* Oldest release which did not ship any wheels with CUDA < 11
  * Migrated to CUDA 11.6 and 11.7, which is right in line where I think bitsandbytes cutoff should be [1]
  * Apple silicon support (MPS) promoted from experimental/prototype to beta
    * 🤗 [Accelerate docs](https://huggingface.co/docs/accelerate/en/usage_guides/mps) "strongly recommend" torch>=1.13 for MPS
* AWS has [HuggingFace containers](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#huggingface-training-containers)with the oldest having `torch==1.13.1+cu117` on py3.9
** In general, all of the Python 3.9+ containers with PyTorch appear to be torch>=1.13.0
* [Azure Container for PyTorch](https://learn.microsoft.com/en-us/azure/machine-learning/resource-azure-container-for-pytorch?view=azureml-api-2) still has a 1.13. configuration

But it could also be a little more permissive for some environments; e.g. transformers recently adopted a constraint of `torch>=1.11,!=1.12.0` in huggingface/transformers#28207, which was also adopted in UKPLab/sentence-transformers#2432.

pyproject.toml
matthewdouglas: I think it's just experimental, but there's code in `bitsandbytes.triton` that is optional. See e67bfccbcd9490f51628bb3a7fac9cfb9c31310d.

pyproject.toml
matthewdouglas: I was going off of the environment.yml for these which didn't have any constraints. I'm seeing #1081 calls for `ruff==0.2.1`, and in general I agree that this extra should be more constrained. I'll make some changes here.

pyproject.toml
matthewdouglas: It should still work work 3.8, and we even have `ruff` targeting for that too. So that is why I submitted #1076, based also on https://github.com/TimDettmers/bitsandbytes/discussions/990#discussioncomment-8558070. So agreed: 3.8 should probably be validated in CI or dropped. I would like to make 3.9 the minimum personally, but not sure what everyone else thinks?

pyproject.toml
rickardp: IMHO (and I recognize this might ba a bit controversial) we should set minimum to what we test on, then pin to these when we test. I don't think we should overstretch to keep supporting old versions of dependencies at the expense of stability or complexity. So I would say ideally we set minimum to what we currently build on and either no upper bound or < next major.

As a user of any package I prefer to know what works or not and not having to be the guinea pig.

But this is my personal opinion, not sure what the rest of you think.

pyproject.toml
rickardp: Ah, yes. Better to target 3.8 in tests until decided to drop it

pyproject.toml
```diff
@@ -1,7 +1,82 @@
...
+    "wheel>=0.42,<1"
+]
+test = [
+    "einops~=0.6.0",
+    "lion-pytorch==0.0.6",
+    "pytest==7.4.4",
+    "pytest-cases>=3.7.0,<4",
+    "pytest-sugar>=1.0.0,<1.1.0",
+    "scipy>=1.10,<2",
+    "transformers>=4.30.1,<5"
```
akx, line 64: Cases isn't used, and Sugar is only for nicer CLI output. Pytest doesn't need to be strictly pinned IMO.

```suggestion
    "einops~=0.6.0",
    "lion-pytorch==0.0.6",
    "pytest~=7.4",
    "scipy>=1.10,<2",
    "transformers>=4.30.1,<5"
```

pyproject.toml
```diff
@@ -1,7 +1,82 @@
...
+]
+dependencies = [
+    "torch>=1.11,!=1.12.0",
+    "numpy>=1.17"
+]
+
+[project.optional-dependencies]
+benchmark = ["pandas", "matplotlib"]
+docs = [
+    "hf-doc-builder @ git+https://github.com/huggingface/doc-builder.git@main"
```
akx, line 48: I dislike that this dependency is not pinned at all (targeting `main`) 😞 (and the last tagged version is from 2022: https://github.com/huggingface/doc-builder/releases/tag/v0.4.0)

I tried to raise this concern over at HF, but no reply in a month. https://github.com/huggingface/doc-builder/pull/434#issuecomment-1920949249

pyproject.toml
```diff
@@ -1,7 +1,82 @@
...
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "bitsandbytes"
+dynamic = ["version"]
+description = "k-bit optimizers and matrix multiplication routines."
+authors = [{name="Tim Dettmers", email="dettmers@cs.washington.edu"}]
+requires-python = ">=3.8"
+readme = "README.md"
+license = {file="LICENSE"}
```
akx, line 12:
```suggestion
license = "MIT"
```

pyproject.toml
matthewdouglas: Agreed. I tried with this pinned to 0.4.0 and it just didn't work. Plus the workflow pulls from `main` too. 🤷‍♂️

docs/source/installation.mdx
matthewdouglas: Reference for VS support

Reference

pyproject.toml
matthewdouglas: The v0.5.0 release is on PyPI now so I've updated to use that.

matthewdouglas:
Rebased

matthewdouglas:
> > Rebased
> 
> This is really weird, somehow the diff of the PR got all messed up now. This makes it really hard to review.
> 
> It seems to me this PR is otherwise ready for review, right?

Not sure what happened there but I think it should look more reasonable now. Apart from that, yes, it's ready.

.github/workflows/python-package.yml
```diff
@@ -122,37 +122,44 @@ jobs:
...
+    - name: Determine and Set Platform Tag, then Tag Wheel
+      shell: bash
+      run: |
+        PLATFORM_TAG=$(python .github/scripts/set_platform_tag.py "${{ matrix.arch }}")
+        echo "PLATFORM_TAG=$PLATFORM_TAG"
+        wheel tags --remove --abi-tag=none --python-tag=py3 --platform-tag=$PLATFORM_TAG dist/bitsandbytes-*.whl
+    - name: Upload build artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: bdist_wheel_${{ matrix.os }}_${{ matrix.arch }}_${{ matrix.python-version }}
```
rickardp, line 160: I am not sure if wheels per Python version is ever needed. Some tests/linting would be good per Python version to validate the claimed support, but that is maybe a separate step?

matthewdouglas: Good catch, this must be something I messed up in a conflict merge. Just above this you can see we're removing the ABI and python version tags from the wheel, so this should be:
`bdist_wheel_${{ matrix.os }}_${{ matrix.arch }}`

pyproject.toml
matthewdouglas: Ideally we would have a CI setup that could test a wider matrix so we can be confident about how we pin these, for sure.

For me part of the point of this `bitsandbytes` project is to try and make things accessible for a wide range of users - so that's across hardware, software, and user's expertise. Since many people, particularly in academia, don't have much control over things like OS, python version, or even torch version, we've got to support a large range in order to make things easy for users. That does come at the cost of maintenance, obviously. Need to find a good balance here.

So at least with this change it's being restricted to `torch>=1.11,!=1.12.0`, which is in line with `transformers`. I don't know that all of those versions of torch will work, but it's an incremental step forward compared to `install_requires=["torch", "numpy"]`, and given the lack of issues around torch versions it does seem likely that it would work.

matthewdouglas:
Superseded by #1373.

PR #1080, opened 2024-02-23
(cmake) Update library output directory

matthewdouglas:
Related to #1035, this change was needed for outputs to be located correctly when building within VS2022 or VSCode on Windows (MSVC, Ninja, CMakePresets in use).


Titus-von-Koeller:
Makes sense, thank you!

PR #1089, opened 2024-02-26
Lit-GPT integration docs

rasbt:
Add a reference to Lit-GPT as discussed [here](https://github.com/Lightning-AI/lit-gpt/issues/955)


Titus-von-Koeller:
Yay! Thanks for the quick response and content is looking great.

PR #1099, opened 2024-02-28
adding whole Linear8bitLt/Linear4bit module save/load serialization

rdyro:
The purpose of this pull request is to allow `torch.save`/`torch.load` directly on modules containing `Linear4bit` and `Linear8bitLt` submodules.

Currently, `torch.save`, then `torch.load` on `Linear8bitLt` (after first forward) causes a missing field `CB` error in the `Int8Params` class. This PR makes torch aware of the `CB` and `SCB` fields in `Int8Params` class.

The core of this PR is
```python
        ~~return torch.Tensor._make_subclass(cls, data, requires_grad)~~
        obj = torch.Tensor._make_subclass(cls, data, requires_grad)
        obj.CB, obj.SCB = cls.CB, cls.SCB
        return obj
```
in  `class Int8Params`

I also added the `torch.save` -> `torch.load` test to the `Linear4bit` (this was already working) and `Linear8bitLt` (this is not yet working).

While saving modules directly in Pytorch with `save` and `load` is not good practice, the change to make this work is minimal and makes disk caching modules for development easier.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -449,7 +449,9 @@ class Int8Params(torch.nn.Parameter):
         cls.SCB = None
         if data is None:
             data = torch.empty(0)
-        return torch.Tensor._make_subclass(cls, data, requires_grad)
+        obj = torch.Tensor._make_subclass(cls, data, requires_grad)
+        obj.CB, obj.SCB = cls.CB, cls.SCB
+        return obj
 
     def cuda(self, device):
         if self.has_fp16_weights:
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -1,4 +1,5 @@
 import copy
+from io import BytesIO
 import os
 import pickle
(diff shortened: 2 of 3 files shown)


Titus-von-Koeller:
Yes, I agree, this is looking good and should be merged before the release. I'll review it more in depth soon.

Thanks @rdyro for the good work and taking the initiative to contribute, really appreciated 🤗

rdyro:
Thanks for the positive feedback! I really like your work with `bitsandbytes`.

Let me know if ideally, the new tests should extend to all Linear layers, not just `Linear4bit` and `Linear8bitLt`.

Titus-von-Koeller:
Dear @rdyro,

I just reviewed your proposed changes and everything really looks good! I don't think any additional tests are needed, what you did already looks good the way it is.

I also ran the `transformers` integration tests and everything came through clean.

Thanks so much for your contribution and if you feel like contributing more, we'd be happy to support you!

PR #1107, opened 2024-03-05
Deduplicate helpers & fix lint issues from #1099

akx:
#1099 was merged with lint failing (missing a trailing newline), and while fixing that, I figured the copy-pasted helper functions could do with deduplicating.


akx:
cc @Titus-von-Koeller

Titus-von-Koeller:
Thanks @akx, good catches and thanks for the quick, proactive fix 🤗

PR #1114, opened 2024-03-07
add audit wheel to CI/CD for visibility of many_linux conformance

Titus-von-Koeller:


.github/workflows/python-package.yml
```diff
@@ -192,6 +192,19 @@ jobs:
...
+        pip install -q auditwheel
+        for whl in dist/*.whl; do
+          printf '%.0s=' {1..60}  # separator line, same as below
+          echo -e "\nChecking wheel: $whl"
+          printf '%.0s-' {1..60}
+          echo
+          auditwheel show "$whl"
+          printf '%.0s-' {1..60}
+          echo
+        done
```
akx, line 207: A future version of this could use https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary for :sparkles:markdown:sparkles: but this is a good start. This could also be a separate step that's done after all of the wheels are done.

akx (approved):

.github/workflows/python-package.yml
Titus-von-Koeller: ok, took a moment to give you the fancy output you desire when waiting for other results..

.github/workflows/python-package.yml
```diff
@@ -198,6 +204,31 @@ jobs:
...
+
+  audit-wheels:
+    needs: build-wheels
+    runs-on: ubuntu-latest
+    env:
+      PIP_DISABLE_PIP_VERSION_CHECK: 1
+    steps:
+    - uses: actions/checkout@v4
+    - name: Download all wheel artifacts
+      uses: actions/download-artifact@v2
```
akx, line 216: You can't use download-artifact@v2 with upload-artifact@v4, which is probably why you're not getting anything here :)

Titus-von-Koeller:
closed in favor of #1116

PR #1075, opened 2024-02-21
[docs] APIs

stevhliu:
This PR expands the API docs to showcase high-level classes and core lower-level building blocks, starting with the optimizers.

- [x] LinearKbit (Linear4bit, Linear8bitLt, etc.)
- [x] Parameters (Params4bit, Int8Params)
- [x] Embeddings (Stable Embedding, Embedding)

PR #1128, opened 2024-03-13
Minor type/doc fixes

akx:
This PR fixes a couple of small typing/docs related things that popped up while running PyCharm's Find Problems on the repo.


akx:
cc @stevhliu: mind double-checking that `args` change (since you apparently added the docs in #1075) – by my reasoning, `args` can't be a dict since those classes are (occasionally) accessing it with `self.args.foo`?

stevhliu:
Hmm, I'm actually not entirely sure about this one 😅 . Here it looks like it is a `dict`:

https://github.com/TimDettmers/bitsandbytes/blob/b03ce0e0be74888e3a604ec749b761621a6c8407/bitsandbytes/optim/optimizer.py#L407

stevhliu:
Ah ok, wasn't aware of `types.SimpleNamespace`. In that case, it looks good to me. Thanks for fixing the types! 🤗

PR #1041, opened 2024-02-06
Rework CUDA/native-library setup and diagnostics

akx:
This PR reworks the native-library setup code that was partially enmeshed with the diagnostics code. I'm sorry it's all in one big commit, this was a bit too hard to pull into smaller ones so it still made sense 😓 

The main thing is that `cuda_setup/` is gone; in its place are 

* `cuda_specs`, which is a simple dataclass containing information about the current cuda environment, and the native-library loading code is now in `cextension.py`. (This would be further reworked when more backends are introduced, of course; see #898.)
  * `cextension` introduces a minimal proxy object for the `ctypes.CDLL` (a precursor to a `Backend`), so a global CUDA-specific constant `COMPILED_WITH_CUDA` is not needed.
* `diagnostics/`, which is a separate thing that's being run by `python -m bitsandbytes`.
  * There was overlapping DLL/.so discovering code in both `cuda_setup` and `__main__.py`; I got rid of the latter implementation in favor of the stuff that had already been there in `cuda_setup`. As far as I could tell, though, that code was only ever used to sniff around for CUDA runtime libraries, but that information was only shown to the user as a diagnostic.
    * Finding the libraries was made a bit looser; anything `nvcuda` or `libcudart` now smells like CUDA.
  * I think the new diagnostics are pretty much on par with the old ones – definitely not bit-by-bit exact though. I don't presently have a machine with broken CUDA, so I can't really tell. (I think the formatting should be made more GitHub friendly though; the `########` headers turn into monstrous Markdown headings. That's for another PR, though.)

Additionally, this PR trivially enables the Mac `libbitsandbytes_cpu.dylib` native library to be found by adding it to the `DYNAMIC_LIBRARY_SUFFIX` mapping.

I tested locally that:

* on my Mac, no tests fail and the library gets loaded as expected
* on my WSL2 machine, `libbitsandbytes_cuda121_nocublaslt.so` gets loaded as expected and tests pass as before.

This is still in no way perfect, but it's better IMO :)

Refs https://github.com/TimDettmers/bitsandbytes/issues/918

PR #1134, opened 2024-03-18
Fix diagnostics error within vscode on windows

matthewdouglas:
This fixes a small issue out of #1041 when running `python -m bitsandbytes` from an integrated VSCode terminal on Windows.

```
OSError: [WinError 231] All pipe instances are busy: '\\\\.\\pipe\\vscode-git-bc89f4fbf9-sock'
```


akx (approved):
Good catch. Can you also see which envvar contains that path so we can add it to the ignore list?

PR #1136, opened 2024-03-18
Add CUDA 12.4 to docs/install helper

matthewdouglas:
CUDA 12.4 was released on 2024-03-06. I've been running builds on both Windows and Linux with this successfully.

The Docker image hasn't shown up yet, so it's not added the build pipeline.


Titus-von-Koeller:
Thanks @matthewdouglas, this is a valuable update!

PR #1147, opened 2024-03-25
Tests: improve memory usage

matthewdouglas:
This simple change frees up GPU memory after each unit test. Prior to this change, I would experience a system crash when attempting to run the full unit test suite on RTX 3060 6GB. It now runs comfortably.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,5 @@
+import gc
+
 import pytest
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller:
Important detail to get right for lower resource environments, thanks for the helpful contribution ❤️

PR #1149, opened 2024-03-26
[docs] Install from source

stevhliu:
From an internal discussion, it may make more sense to split off compilation instructions into it's own section because it's pretty much the same on Linux/Windows, and so we can add more details. Feel free to discuss and suggest what we can add, and we can iterate on this PR! 🙂 

cc @matthewdouglas @akx @Titus-von-Koeller if you have any ideas about other things we can add here

### DIFF
--- a/docs/source/installation.mdx
+++ b/docs/source/installation.mdx
@@ -2,6 +2,14 @@
 
 bitsandbytes is only supported on CUDA GPUs for CUDA versions **11.0 - 12.3**.
 
+The latest version of bitsandbytes (v0.43.0) builds on:
+
+| OS | CUDA | Compiler |
+|---|---|---|
+| Linux | 11.7 - 12.3 | GCC 11.4 |
+|  | 12.4+ | GCC 13.2 |
+| Windows | 11.7 - 12.4 | MSVC 19.38+ (VS2022 17.8.0+) |
+
 > [!TIP]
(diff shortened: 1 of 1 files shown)


matthewdouglas:
Some info we can add regarding validated compilers.

The v0.43.0 release is known to build on:

Windows: MSVC 19.38+ (VS2022 17.8.0+) with CUDA 11.7-12.4
Linux: GCC 11.4 (with CUDA 11.7-12.3), 13.2 (with CUDA 12.4+)

Other notes:

* Linux: Expected CUDA Toolkit compatibility is currently with 11.1+.
* Linux: GCC >= 7.3 recommended. GCC >= 6 is required.
* Linux: CUDA Toolkit >= 11.4.1 is required when building with GCC >= 11.
* Linux: CUDA Toolkit >= 12.0 is required when building with GCC >= 12.
* Linux: CUDA Toolkit >= 12.4 is required when building with GCC >= 13.
* Windows: CUDA Toolkit >= 11.6 is required when building with MSVC 19.30+ (VS2022).

stevhliu:
Thanks, I've added that info as a table in the installation

Titus-von-Koeller:
Great, thank you all, this is looking very good. I'm happy that we have more details here now, I think this is important, so that we properly cover the edge-cases.

PR #1171, opened 2024-04-08
Add CUDA 12.4 to Linux x86-64 Build Workflow

matthewdouglas:
It took a little while, but NVIDIA published the `nvidia/cuda:12.4.0-devel-ubuntu22.04` container. This PR enables building with CUDA 12.4.0 on Linux x86-64.

Windows builds use an action that hasn't been updated yet, so will need to enable at a future point.


Titus-von-Koeller:
Very good, thank you @matthewdouglas! 🤗

PR #1187, opened 2024-04-23
Fixed optim update error with non-contiguous grads/params

### LINKED ISSUE #1185: 32 bit optimizer update error despite gradients being the same
### System Info

A100 GPU, torch 2.1, cuda 12.1, bitsandbytes 0.43.1

### Reproduction

The tensors to be loaded are zipped here:
[grads.zip](https://github.com/TimDettmers/bitsandbytes/files/15041673/grads.zip)
```
import torch
low_rank_grad = torch.load("low_rank_grad.pt")
dist_low_rank_grad = torch.load("dist_low_rank_grad.pt")
def update(low_rank_grad):
    p = torch.zeros_like(low_rank_grad, dtype=torch.float32, device="cuda")
    p.grad = low_rank_grad
    lr = 1e-2
    state = {}
    state["state1"] = torch.zeros_like(low_rank_grad, dtype=torch.float32, device="cuda")
    state["state2"] = torch.zeros_like(low_rank_grad, dtype=torch.float32, device="cuda")
    beta1 = 0.9
    beta2 = 0.999
    step = 1
    args = None
    eps = 1e-8
    weight_decay = 1e-2


    F.optimizer_update_32bit(
        "adam",
        p.grad,
        p,
        state["state1"],
        beta1,
        eps,
        step,
        lr,
        state["state2"],
        beta2,
        weight_decay
    )
    return p
print(low_rank_grad.shape, dist_low_rank_grad.shape)
assert (low_rank_grad[:, :32] == dist_low_rank_grad).all()
# adam update step
low_rank_grad = update(low_rank_grad)
dist_low_rank_grad = update(dist_low_rank_grad)
low_rank_grad[:, :32] == dist_low_rank_grad
```
My result showing that most updates on the same grad chunk diverged
image

### Expected behavior

This comes from adapting the  Galore optimizer for Tensor parallel, when testing precision of the distributed and original optimizer.
Here  the gradient is shared along dim 1 by tensor parallel, but the corresponding grad chunk clearly matches. However after the optim step the chunks are not exactly the same. I first doubted this is due to quantization statistics, but using 32 bit and disabling quantization stably leads to this bug.

Edenzzzz:
Fixes #1185 
Non-contiguous params/gradients resulting from `torch.chunk` and `all_gather` etc. are ubiquitous in distributed training frameworks such as ZeRO. This avoids update errors as the C++ kernels assume row-major inputs. 

image
cc @Titus-von-Koeller

### DIFF
--- a/bitsandbytes/optim/optimizer.py
+++ b/bitsandbytes/optim/optimizer.py
@@ -474,9 +474,13 @@ class Optimizer2State(Optimizer8bit):
 
     @torch.no_grad()
     def update_step(self, group, p, gindex, pindex):
+        # avoid update error from non-contiguous memory layout
+        p.data = p.data.contiguous()
+        p.grad = p.grad.contiguous()
+        
         state = self.state[p]
         grad = p.grad
-
+        
         config = self.get_config(gindex, pindex, group)
 
         state["step"] += 1
@@ -685,6 +689,10 @@ class Optimizer1State(Optimizer8bit):
 
     @torch.no_grad()
     def update_step(self, group, p, gindex, pindex):
+        # avoid update error from non-contiguous memory layout
+        p.data = p.data.contiguous()
+        p.grad = p.grad.contiguous()
+        
         state = self.state[p]
         grad = p.grad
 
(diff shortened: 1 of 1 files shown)


matthewdouglas:
@Edenzzzz Thanks for the PR! This looks good to me!
cc: @Titus-von-Koeller

Titus-von-Koeller:
Just ran the optimizer test suite as well, everything passing.

Thanks @Edenzzzz and @matthewdouglas for tracking this down and providing a fix 🙌🏻 🚀, this was really helpful!

PR #1188, opened 2024-04-23
TESTS: add `is_flaky` decorator on flaky tests

younesbelkada:
# What does this PR do?

This PR introduces a new pytest decorator `is_flaky` that runs tests that have been decorated as being flaky. The decorator will run the test `max_attempts` times and if the test failed during all these attempts the test will be marked as failed.

Plan forward: identify which tests are flaky and decorate them

cc @Titus-von-Koeller

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -2,6 +2,8 @@ import functools
 from io import BytesIO
 from itertools import product
 import random
+import sys
+import time
 from typing import Any, List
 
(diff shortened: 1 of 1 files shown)


akx:
You could also use https://pypi.org/project/pytest-retry/...

Titus-von-Koeller:
Closing this in favor of pytest-retry approach.

PR #1196, opened 2024-05-06
WIP: test docs pr workflow (sth about a missing dependency)

Titus-von-Koeller:


Titus-von-Koeller:
was fixed in separate PR

PR #1197, opened 2024-05-06
fix doc builder workflow by reverting to old image

Titus-von-Koeller:
doc-builder image had been changed, need to revert to old one due to dependency issues, see https://huggingface.slack.com/archives/C021H1P1HKR/p1714469081588139


younesbelkada (commented):
Thanks ! Can you also propagate these changes to other doc-build workflows ? 🙏

Titus-von-Koeller:
@younesbelkada Jup, thanks for reminding me of that 🤗 I added it for the job that runs for `main`. The other one that uploads it and posts on the PR is fine as is. It still works successfully.

PR #1206, opened 2024-05-10
Support 4bit on CPU backend

Xia-Weiwen:
Adds implementation for the following ops on CPU backend:
- quantize_4bit
- dequantize_4bit
- gemv_4bit

Limitations:
- `quant_storage` must be torch.uint8
- `compress_statistics` is not supported yet (`bnb_4bit_use_double_quant` must be false)
- `fp4` is slow currently because there is no fused kernel yet.

Difference from CUDA implementation:
- On CPU backend, it is not required that A is a vector to go to the fused dequant-gemm kernel. CUDA requires that. So, the op is called `gemv_4bit`. But on CPU backend, it's actually GEMM.
- Different numerical accuracy due to different kernel implementations

Here is the code snippet of an example to run HuggingFace models with 4bit on CPU backend: https://gist.github.com/Xia-Weiwen/592d6e24e03f904a18692b3e27794c53. You will have to bypass CUDA checks in transformers to run.

---
cc @jiqing-feng @jgong5 @jianan-gu

### DIFF
--- a/bitsandbytes/backends/cpu_xpu_common.py
+++ b/bitsandbytes/backends/cpu_xpu_common.py
@@ -1,11 +1,11 @@
-import warnings
-import torch
 from typing import Optional
+import warnings
+
+import torch
+
 from bitsandbytes.functional import (
-    get_4bit_type,
-    quantize_blockwise,
-    dequantize_blockwise,
     QuantState,
+    get_4bit_type,
 )
 
 try:
@@ -237,23 +237,35 @@ def mm_dequant_impl(
 
 
 NF4_QUANT_TABLE = [
-    -1.0 - 1e-2,           # 0b0000
-    -0.8480964004993439,   # 0b0001
-    -0.6106329262256622,   # 0b0010
-    -0.4599952697753906,   # 0b0011
+    -1.0 - 1e-2,  # 0b0000
+    -0.8480964004993439,  # 0b0001
+    -0.6106329262256622,  # 0b0010
+    -0.4599952697753906,  # 0b0011
     -0.33967943489551544,  # 0b0100
     -0.23460740596055984,  # 0b0101
     -0.13791173323988914,  # 0b0110
-    -0.045525018125772476, # 0b0111
-    0.03979014977812767,   # 0b1000
-    0.1202552504837513,    # 0b1001
-    0.2035212516784668,    # 0b1010
-    0.2920137718319893,    # 0b1011
-    0.3893125355243683,    # 0b1100
-    0.5016634166240692,    # 0b1101
-    0.6427869200706482,    # 0b1110
-    0.8614784181118011,    # 0b1111
+    -0.045525018125772476,  # 0b0111
+    0.03979014977812767,  # 0b1000
+    0.1202552504837513,  # 0b1001
+    0.2035212516784668,  # 0b1010
+    0.2920137718319893,  # 0b1011
+    0.3893125355243683,  # 0b1100
+    0.5016634166240692,  # 0b1101
+    0.6427869200706482,  # 0b1110
+    0.8614784181118011,  # 0b1111
 ]
+
+
+FP4_QUANT_TABLE = {
+    0 - 1e-2: 0,  # 0b0000
+    0.00260417: 1,  # 0b0001
+    0.0859375: 6,  # 0b0110
+    0.20833333: 7,  # 0b0111
+    0.29166667: 4,  # 0b0100
+    0.4166667: 5,  # 0b0101
+    0.583333: 2,  # 0b0010
+    0.8333333: 3,  # 0b0011
+}
 
 
 # It's faster not to use torch.compile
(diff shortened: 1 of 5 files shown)


bitsandbytes/backends/cpu_xpu_common.py
```diff
@@ -228,3 +234,282 @@ def mm_dequant_impl(
...
+        out = torch.empty(quant_state.shape, dtype=quant_state.dtype, device=A.device)
+
+    n = out.numel()
+    # Map nf4 to [-1, 1]
+    out_uint8 = torch.empty(A.size(0) * 2, dtype=torch.uint8, device=A.device)
+    out_uint8[::2] = A.bitwise_and(0xF)
+    out_uint8[1::2] = A.bitwise_right_shift(4)
+    out_dq = torch.empty(out_uint8.shape).to(quant_state.dtype)
+    for i in range(len(quant_state.code)):
+        out_dq[out_uint8 == i] = quant_state.code[i]
```
jiqing-feng, line 460: Using index select will be faster `out_dq = quant_state.code[out_uint8.to(torch.int32)]`.

Xia-Weiwen: Looks like `torch.compile` result of this code gives wrong results. And removing `torch.compile` results in lower performance. Let's keep this implementation for now.

jgong5: A bug in torch.compile? Can you submit a bug to PyTorch? I will try to fix it.

Xia-Weiwen: However, I cannot reproduce the issue with the script below. May need more investigation.
```python
import torch


NF4_DEQUANT_TABLE = torch.Tensor([
  -1.0,
  -0.6961928009986877,
  -0.5250730514526367,
  -0.39491748809814453,
  -0.28444138169288635,
  -0.18477343022823334,
  -0.09105003625154495,
  0.0,
  0.07958029955625534,
  0.16093020141124725,
  0.24611230194568634,
  0.33791524171829224,
  0.44070982933044434,
  0.5626170039176941,
  0.7229568362236023,
  1.0,
])


@torch.compile
def dequant_nf4_compile(t_in: torch.Tensor, out_dtype):
  return NF4_DEQUANT_TABLE[t_in.to(torch.int)].to(out_dtype)


def dequant_nf4_eager(t_in: torch.Tensor, out_dtype):
  return NF4_DEQUANT_TABLE[t_in.to(torch.int)].to(out_dtype)


x = torch.randint(0, 16, (1024, 1024), dtype=torch.uint8)

y1 = dequant_nf4_compile(x, torch.bfloat16)
y1 = dequant_nf4_compile(x, torch.bfloat16)
y2 = dequant_nf4_eager(x, torch.bfloat16)

print(torch.equal(y1, y2))
print("max diff =", torch.abs(y1 - y2).max())
```

jiqing-feng:
Hi @Titus-von-Koeller . Here is the test results on Intel 4th Gen Xeon CPU of this PR:
image

The big difference between NF4 and FP4 is that we can use fused ops in NF4, but they are not prepared in FP4. FP4 will also support fused ops and is supposed to get the same performance as NF4, maybe in the next Ipex release. Would you please review it? Thx!

Titus-von-Koeller:
Dear @Xia-Weiwen et al,

Unfortunately we're (mostly me alone) quite resource constrained and humbled by the workload associated with the `multi-backend-refactor`. I just talked with my colleague @younesbelkada about the topic how to best handle the next steps.

We both took a look at this PR and the one from AMD and think that at first glance everything looks really good. At this time, both me and Younes are not in a position to give detailed feedback and I need to focus on concretizing the path forward on how to integrate with the PyTorch dispatcher (tensor driven dispatch, as requested) through the torch.library Python-level APIs. After extensive research and yesterday's consultation with 3 PyTorch devs at Meta that are experts on the topic I need to focus on making this new input concrete.

However, for the purpose of iterative progress (as agreed in our prior conversations), we've decided to already go ahead and merge both the open Intel and AMD branches into `multi-backend-refactor`, where interested parties can then compile from source and give the new functionality (we're so excited and grateful about this!) a thorough testing.

Once we've made some progress on the `torch.library` based refactor, I'll next focus on enabling the nightly releases for that branch as well. We're also looking forward to your feedback on the this torch.library / tensor-driven dispatch topic once the code is there on the basis of which to discuss (and refactor the backend specific code towards that new target, after we agreed with all of you that this is the right path).

Among other things, there's also been extensive ongoing work in the background on things like moving BNB to a new independent/non-profit Github org, but under the umbrella of Hugging Face and the support of their infra team for managing the complexities of the CI/CD backend and runners. Also, we're working to make Github runners for the different hardware platforms a reality (thanks for your help on that!).

Thanks again for the good work and active collaboration! ❤️ 🚀

Titus-von-Koeller:
P.S. Also see this: [README: asking for help from volunteer alpha testers](url)

Let us know if you have further thoughts on this and how you think it's best to communicate about this.

Xia-Weiwen:
Hi @Titus-von-Koeller Thanks a lot for your help on this. It's so weird that as the owner of this PR I didn't receive any notifications. Anyway, we are glad to provide feedbacks on the adoption of `torch.library`. Please let use know when there is any update.
Also, we would love to volunteer to conduct regular tests on Intel CPU/GPU as alpha testers. I think we will need to be aligned on many aspects of the tests, such as test code base, methods, frequency, scopes and how we sync and publish the results. Maybe we can create an issue to track this. I will have discussions with my colleagues and come back later.

jgong5:
> At this time, both me and Younes are not in a position to give detailed feedback and I need to focus on concretizing the path forward on how to integrate with the PyTorch dispatcher (tensor driven dispatch, as requested) through the torch.library Python-level APIs. After extensive research and yesterday's consultation with 3 PyTorch devs at Meta that are experts on the topic I need to focus on making this new input concrete.

Hi @Titus-von-Koeller May I learn more details about how you are going to refactor things via `torch.library`? I guess this is one of the official ways of integrating native backend implementations with PyTorch, which provides python bindings and backend dispatching mechanism. I shared similar comments earlier too: https://github.com/TimDettmers/bitsandbytes/pull/898#issuecomment-1909439116.

Meanwhile, it would be beneficial to allow the flexibility of backend integration without adding native code explicitly to bitsandbytes too, like optimizing via `torch.compile` as what this PR does. For simpler ops without involving gemms, this is a more light-weight approach than adding native code.

PR #1211, opened 2024-05-14
[docs] Clarify FSDP-QLoRA

stevhliu:
Fix and clarify the docs around:

1.  FSDP only supports floating data type in `quant_storage`
2. show a more common way to access and configure this option with `BitsAndBytesConfig` in Transformers
3. explain `compute_dtype` and the importance of matching `quant_storage` data type with the data types used everywhere else
4. link to code example from PEFT docs

cc @Titus-von-Koeller

### DIFF
--- a/docs/source/fsdp_qlora.md
+++ b/docs/source/fsdp_qlora.md
@@ -9,21 +9,26 @@ This guide provides a brief guide on how bitsandbytes supports storing quantized
 
 ## Quantized data storage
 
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller:
In the first `!TIP` box in the Training section, I would add a sentence (maybe in bold?) like "Note that FSDP is a distributed framework and therefore needs to be kicked off as a distributed training job, which is described in the following resources."

Otherwise it doesn't become clear for novices that these code examples alone (on this BNB doc page) don't suffice and FSDP is not run as standard single threaded Python that they can run in an interpreter or notebook, but needs to be kicked as a distributed way through Accelerate (or torchrun or alike if we weren't using HF libs).

Titus-von-Koeller:
Hey @stevhliu, thanks a lot for taking the time to further polish this page, really appreciated! 🤗 

I left a few suggestions. Please feel free to integrate my feedback at your discretion and then we'll merge.

Titus-von-Koeller (approved):

PR #1238, opened 2024-06-05
CPU: add torch.compile for F.double_quant and F.quantize_4bit

Xia-Weiwen:
We didn't use `torch.compile` for these two functions because of some issue. These issues are now fixed in PyTorch.
Using `torch.compile` brings performance gain and also avoid some issues during finetuning.

### DIFF
--- a/bitsandbytes/backends/cpu_xpu_common.py
+++ b/bitsandbytes/backends/cpu_xpu_common.py
@@ -55,7 +55,7 @@ def _maybe_torch_compile(func):
     return func
 
 
-# Don't use torch.compile for now due to PyTorch issue https://github.com/pytorch/pytorch/issues/124382
+@_maybe_torch_compile
 def double_quant_impl(A, col_stats=None, row_stats=None, out_col=None, out_row=None, threshold=0.0):
     """
     Find absolute max values of each row/column of a tensor, and symmetrically quantize it to int8.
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller:
Cool! Is this ready? Because it's still marked as draft.

Xia-Weiwen:
> Cool! Is this ready? Because it's still marked as draft.

Thanks. I think @jiqing-feng will collect some data before we mark it ready.

jiqing-feng:
For `quantize_4bit_impl`, it only happens in the load model, so these are no difference with nf4/fp4 performance before and after we applied torch.compile.

For `double_quant_impl`, I have tested it on llama-7b inference and opt-6.7b finetune;found no difference in performance.

We plan to add torch.compile in quantization op but didn't integrate in the first version, this is because the issue in the comment:`Don't use torch.compile for now due to PyTorch issue https://github.com/pytorch/pytorch/issues/124382`. And now the issue has been fixed, so we can add torch.compile again. The torch.compile will be optimized constantly in pytorch :)


Thx!

Titus-von-Koeller:
Thanks @jiqing-feng for explaining. Then it's ready to merge, @Xia-Weiwen, right?

Xia-Weiwen:
> Thanks @jiqing-feng for explaining. Then it's ready to merge, @Xia-Weiwen, right?

Yes. Please. Thanks.

PR #1251, opened 2024-06-17
CPU/XPU: disable torch.compile if g++ is not available

Xia-Weiwen:
`torch.compile` requires `g++`. On platforms like Windows, `g++` is not available. In that case, `torch.compile` is disabled to avoid runtime errors. With this patch, the CPU/XPU can be enabled on Windows.

I have validated the patch by running UTs for CPU backend in `test_functional.py` and a real model (OPT-1.3B) on my local machine (Windows 11 Enterprise, 12th Gen Intel(R) Core(TM) i7-1270P  2.20 GHz, 16.0 GB RAM).

### DIFF
--- a/bitsandbytes/backends/cpu_xpu_common.py
+++ b/bitsandbytes/backends/cpu_xpu_common.py
@@ -1,5 +1,6 @@
 from typing import Optional
 import warnings
+import subprocess
 
 import torch
 
@@ -17,6 +18,14 @@ try:
 except BaseException:
     ipex_cpu = None
     ipex_xpu = None
+
+
+gxx_available = False
+try:
+    subprocess.run(["g++", "--version"])
+    gxx_available = True
+except BaseException:
+    warnings.warn(f"g++ not found, torch.compile disabled for CPU/XPU.")
 
 
 Tensor = torch.Tensor
@@ -45,8 +54,8 @@ def _ipex_xpu_version_prereq(major, minor):
 
 
 def _maybe_torch_compile(func):
-    # torch.compile requires pytorch >= 2.0
-    if _torch_version_prereq(2, 0):
+    # torch.compile requires g++ and pytorch >= 2.0
+    if gxx_available and _torch_version_prereq(2, 0):
         options = {}
         # fx_graph_cache requires pytorch >= 2.2
         if _torch_version_prereq(2, 2):


Xia-Weiwen:
Hi @Titus-von-Koeller Could you please review? Thanks. Looks like the CI failure is about ROCM. I am not sure if the failure is caused by this PR.

matthewdouglas:
Related: pytorch/pytorch#124245

I'm not sure what PyTorch release that's going to land in, but good news is that it looks like it'll eventually support MSVC on Windows and we can update this as appropriate. Until then, I think this will be OK.

xuhancn:
https://github.com/pytorch/pytorch/blob/main/torch/_inductor/cpp_builder.py#L123-L134 @Xia-Weiwen FYI.

Xia-Weiwen:
Hi @matthewdouglas Yes, it will still take some time to enable `torch.compile` on Windows, and the clang compiler is required. The contributor is @xuhancn . So, I would agree that it's better to disable it on Windows now and enable it later when everything is ready.

xuhancn:
> Hi @matthewdouglas Yes, it will still take some time to enable `torch.compile` on Windows, and the clang compiler is required. The contributor is @xuhancn . So, I would agree that it's better to disable it on Windows now and enable it later when everything is ready.

Windows default compiler is msvc(cl.exe), please correct it. :)

Xia-Weiwen:
> > Hi @matthewdouglas Yes, it will still take some time to enable `torch.compile` on Windows, and the clang compiler is required. The contributor is @xuhancn . So, I would agree that it's better to disable it on Windows now and enable it later when everything is ready.
> 
> Windows default compiler is msvc(cl.exe), please correct it. :)

Thanks for the correction.

Titus-von-Koeller:
> > > Hi @matthewdouglas Yes, it will still take some time to enable `torch.compile` on Windows, and the clang compiler is required. The contributor is @xuhancn . So, I would agree that it's better to disable it on Windows now and enable it later when everything is ready.
> > 
> > Windows default compiler is msvc(cl.exe), please correct it. :)
> 
> Thanks for the correction. 

@Xia-Weiwen does this still need correction or is this ready to merge? In the diff it still shows gcc for the compiler...

matthewdouglas:
@Titus-von-Koeller That's just a correction on comment about the compiler requirement for Windows (when we get there). It will use MSVC by default and not clang. I think it's OK to merge now as it is and my commentary here is just a note for future refinement.

Xia-Weiwen:
Hi @Titus-von-Koeller . This PR does not need correction. `torch.compile` is not ready on Windows yet. G++ is required for `torch.compile` now. So, this PR just disable it on Windows by checking if G++ is available. @matthewdouglas is correct. Thanks.

Xia-Weiwen:
Hi @Titus-von-Koeller Do you have more comments? Thanks

Xia-Weiwen:
Hallo @Titus-von-Koeller It's been a while since last update. Do you think there are any blocking issues preventing this PR merging? Thanks.

Titus-von-Koeller:
@Xia-Weiwen Yes, everything is right the way it is then, thanks for the clarification!

Xia-Weiwen:
> @Xia-Weiwen Yes, everything is right the way it is then, thanks for the clarification!

Thanks!

PR #1255, opened 2024-06-21
Create build job for ROCm

pnunna93:
This PR adds a new job to python-package.yml file which compiles and builds bitsandbytes for hip backend. This is to test compilation and build, and doesn't upload artifacts, I will open a separate PR for wheels build using .so file from this build.
cc: @Titus-von-Koeller @amathews-amd @sunway513


Titus-von-Koeller:
thanks @pnunna93 !

PR #1257, opened 2024-06-25
Add xpu support for 4bit (nf4/fp4) and 8bit quantization

zhuhong61:


zhuhong61:
@jiqing-feng Could you please help review? thanks!

Titus-von-Koeller:
@jiqing-feng @zhuhong61 Is this PR orphaned? In that case I would close it and you can resubmit whenever you're ready.

PR #1259, opened 2024-06-26
Fixed tests for cpu only platforms

galqiwi:
`test_4bit_warnings` is the only test that fails on cpu only platforms with CUDA-enabled torch.

The reason for this is that other tests fall into [this](https://github.com/TimDettmers/bitsandbytes/blob/dada530149212d64d4b69534716202659ef37ec8/tests/conftest.py#L20) branch
```
except RuntimeError as re:
    # CUDA-enabled Torch build, but no CUDA-capable device found
    if "Found no NVIDIA driver on your system" in str(re):
        pytest.skip("No NVIDIA driver found")
    raise
```

While `test_4bit_warnings` fails with `Failed: DID NOT WARN. No warnings of type (<class 'UserWarning'>,) were emitted`, as the error is catched by `pytest.warns`.

### DIFF
--- a/tests/test_linear8bitlt.py
+++ b/tests/test_linear8bitlt.py
@@ -182,7 +182,7 @@ def test_linear_serialization(
 
 
 @pytest.fixture
-def linear8bit():
+def linear8bit(requires_cuda):
     linear = torch.nn.Linear(32, 96)
     linear_custom = Linear8bitLt(
         linear.in_features,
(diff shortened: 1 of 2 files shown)


galqiwi:
Sorry, looked over 3 tests that fail during setup in `linear8bit` fixture.

Now `pytest ./tests` truly works
```
Results (295.12s (0:04:55)):
      54 passed
    3209 skipped
```

Titus-von-Koeller:
Thank you so much @galqiwi, really appreciated! It's always good to have folks with a hands-on attitude and spot-on :)

Sorry for the delay, unfortunately was out sick, etc..

PR #1270, opened 2024-07-05
chore: delete useless buffered activation

Ther-nullptr:
For QLoRA models, we do not need to update the $\mathbf{W}$, so the buffered activation of $\mathbf{A}$ is useless. It is suggested not to save $\mathbf{A}$ in `ctx` to save the memory.

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -526,7 +526,7 @@ class MatMul4Bit(torch.autograd.Function):
             return torch.zeros_like(ctx.A), torch.zeros_like(ctx.B), None, bias_grad, None
 
         req_gradA, _, _, req_gradBias, _ = ctx.needs_input_grad
-        None, B = ctx.tensors
+        _, B = ctx.tensors
 
         grad_A, grad_B, grad_bias = None, None, None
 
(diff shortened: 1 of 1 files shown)


TimDettmers:
Great catch! This is the bug that caused a lot of memory issues with QLoRA. I cannot believe I missed this bug for such a long time. This will very significantly reduce the memory footprint of QLoRA fine-tuning with larger batch sizes or longer sequence lengths.

Titus-von-Koeller:
Jup, great work @Ther-nullptr

Just out of curiosity, how did you become aware of this, track this down?

Titus-von-Koeller:
I'm aiming to do a patch release this week to make sure this reaches all users asap. We've had this on our backlog for a bunch of time, but it's been a bit tricky to track down and then we had to deprioritize it as our limited resources were tied up in the multi-backend-refactor stuff (and being sick)..

Thank you so much for this extremely helpful contribution, @Ther-nullptr !

Titus-von-Koeller:
Ah, just saw [your repo and this commit](https://github.com/thu-ee-acts-lab/HyC-LoRA/commit/21e825412de1a5ede5c7b19a436048a9e70d8955), really interesting!

Please tag me on any topics that supports your use-case, helps unblock you. We've had a bit of an impasse with maintenance recently, this way you can be sure it doesn't slip our radar.

Titus-von-Koeller:
Thank you so much for this extremely helpful contribution, @Ther-nullptr ! Saw [your repo and this commit](https://github.com/thu-ee-acts-lab/HyC-LoRA/commit/21e825412de1a5ede5c7b19a436048a9e70d8955), really interesting! Please tag me if there's anything else we can help you with in that effort.

Release with this fix just went live!

AaronZLT:
I'm quite new to qlora's repo, what is the $A$ and $B$ in this context? Are they from LoRA's $LoRA_A$ and $LoRA_B$? If yes why that we do not need to store the activation of $A$ since $A$ and $B$ is trainable?

PR #1273, opened 2024-07-10
Fix CUDA 12.5 build issue

HennerM:
pythonInterface.cpp depends on ops.cuh
which in turn depends on some thrust headers.
It is defined as a C++ compilation unit
which is problematic  becuase thrift doesn't guarantee compatibility with a host compiler.

This is starting to cause compiler issues like this:
```
/usr/local/cuda/targets/x86_64-linux/include/cub/util_ptx.cuh:377:37: error: invalid output constraint '=f' in asm
    asm ("mul.rz.f32 %0, %1, %2;" : "=f"(d) : "f"(a), "f"(b));
                                    ^
/usr/local/cuda/targets/x86_64-linux/include/cub/util_ptx.cuh:388:41: error: invalid output constraint '=f' in asm
    asm ("fma.rz.f32 %0, %1, %2, %3;" : "=f"(d) : "f"(a), "f"(b), "f"(c));
                                        ^
/usr/local/cuda/targets/x86_64-linux/include/cub/util_ptx.cuh:415:39: error: use of undeclared identifier 'threadIdx'
    return ((block_dim_z == 1) ? 0 : (threadIdx.z * block_dim_x * block_dim_y)) +
                                      ^
/usr/local/cuda/targets/x86_64-linux/include/cub/util_ptx.cuh:416:40: error: use of undeclared identifier 'threadIdx'
            ((block_dim_y == 1) ? 0 : (threadIdx.y * block_dim_x)) +
                                       ^
/usr/local/cuda/targets/x86_64-linux/include/cub/util_ptx.cuh:417:13: error: use of undeclared identifier 'threadIdx'
            threadIdx.x;
```

A similar issue is being reported in https://github.com/NVIDIA/cccl/issues/1373 and https://forums.developer.nvidia.com/t/when-upgrade-from-cuda12-4-to-12-5-the-compilation-became-broken/295814 indicating that cuda 12.5 introduced some breaking changes here.

We don't actually depend on any symbols from the thrift headers, removing them in pythonInterface.cpp and kernels.cu fixes the issue

### DIFF
--- a/csrc/kernels.cu
+++ b/csrc/kernels.cu
@@ -12,8 +12,6 @@
 #include <cub/block/block_reduce.cuh>
 #include <cub/cub.cuh>
 #include <math_constants.h>
-#include <thrust/host_vector.h>
-#include <thrust/device_vector.h>
 #include <mma.h>
 
 
--- a/csrc/ops.cuh
+++ b/csrc/ops.cuh
@@ -19,10 +19,6 @@
 #include <vector>
 #include <functional>
 
-#include <thrust/host_vector.h>
-#include <thrust/device_vector.h>
-
-
 
 #define CUDA_CHECK_RETURN(value) {                      \
   cudaError_t _m_cudaStat = value;                    \


Titus-von-Koeller:
Thanks @HennerM for raising this and immediately providing a fix, this proactive attitude is super helpful and appreciated ❤️ 

cc @matthewdouglas

HennerM:
@matthewdouglas thanks for the review - I wasn't even checking if the headers are needed at all. Have removed them now and all still compiles fine. Thanks for the shout.

matthewdouglas (approved):

matthewdouglas:
@HennerM Thank you!

@Titus-von-Koeller We should be good merging this. It may still fail on Windows because of an MSVC upgrade. I will open separate PRs to take care of that and enable CUDA 12.5 builds in CI.

Titus-von-Koeller:
Ok, thanks a lot both! 🤗  We have to introduce some linting, auto-fixing tool to avoid such cases where dependencies or variables are declared but not used. Just have to do some research what tools exists that fully cover CUDA as well. Let me know if you have suggestions..

PR #1276, opened 2024-07-12
Fix Windows CUDA build compatibility with newest MSVC

matthewdouglas:
A recent update to CI build image for Windows has broken the Windows CUDA builds. This is because the latest version of the Visual Studio compiler is version is v1940 [1]. For CUDA versions below 12.4, the check that `nvcc` uses for support is to only allow up to v193x [2]. The intent of that check is to support up to VS2022.

There's a change in CMake 3.29.4 for this when using the VS generator [3], but this PR will work around this in a similar way for older CMake versions and when using ninja.

cc: @Titus-von-Koeller 

**References**
[1] https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/
[2] https://github.com/pytorch/vision/pull/8492
[3] https://gitlab.kitware.com/cmake/cmake/-/issues/26003

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,9 +188,15 @@ if(WIN32)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
-# Weird MSVC hacks
 if(MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2 /fp:fast")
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller (approved):

Titus-von-Koeller:
Perfect, @matthewdouglas, looking good! Thanks a lot for handling this ❤️

PR #1282, opened 2024-07-17
add job to upload wheels to continuous pre-release

Titus-von-Koeller:
After plenty of discussion around different ways of making pre-built wheels available for preview, I'm trying out an approach that seems the most promising so far.

I'm starting simple to try things out and iterate from there, but the idea is that we can have multiple 'continuous' pre-releases. In practice this means that we're using `pyuploadtool` which I came across towards the end of my extensive research to create and consequently keep updating pre-releases and their assets.

I spent quite some time with another approach that would have allowed users to download the artifacts of a branch, unzip them, install the wheel. However that approach needs a PAT to access the wheels, sth that became apparent too late (the GH docs seem to say sth else...).

Anyways, this approach doesn't have that problem, as (pre-)release assets are globally readable without rate limit or needing to be logged in (PAT). We can also attach the wheels directly, meaning there's no intermediate zip to unpack:

Therefore users will simply be able to install via sth like 

`pip install https://github.com/bitsandbytes-foundation/bitsandbytes/releases/download/continuous-build_multi-backend-refactor/bnb-linux-x86_64.whl`

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -103,6 +103,7 @@ jobs:
           name: shared_library_cuda_${{ matrix.os }}_${{ matrix.arch }}_${{ matrix.cuda_version }}
           path: output/*
           retention-days: 7
+
   build-wheels:
     needs:
       - build-shared-libs
@@ -123,7 +124,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - name: Download build artifact
+      - name: Download build artifacts
         uses: actions/download-artifact@v4
         with:
           merge-multiple: true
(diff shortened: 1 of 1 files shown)


.github/workflows/python-package.yml
```diff
@@ -152,6 +153,52 @@ jobs:
...
+              exit 1
+            fi
+          done
+      - name: Inspect wheels directory after renaming files
+        run: ls -alFR wheels/
+      - name: Create release and upload artifacts
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_CONTINUOUS_RELEASE_TYPE: prerelease
+          GITHUB_CONTINUOUS_RELEASE_TAG: continuous-release_main
```
winglian, line 196: @Titus-von-Koeller Would it be possible to have versioned tags by including the GitHub SHA? something like 
`continuous-release_main-${{ github.sha }}`

PR #1292, opened 2024-07-24
Embedding4bit and Embedding8bit implementation

galqiwi:
Hi! I've been [researching](https://arxiv.org/abs/2405.14852) LLM quantization and found a bottleneck that I think this PR can fix.

When using extreme 1 bit and 2 bit LLM quantization (which have seen many improvements recently [1](https://arxiv.org/abs/2307.13304), [2](https://arxiv.org/abs/2401.06118v2), [3](https://arxiv.org/abs/2402.04396), [4](https://arxiv.org/abs/2405.14852), [5](https://arxiv.org/abs/2406.11235)), uncompressed embeddings can start to take up too much space (in some cases more than 50%).

I've documented this bottleneck in huggingface/transformers [issue](https://github.com/huggingface/transformers/issues/31474), and it looks like the bitsandbytes library can be a good place to start dealing with it.

In this PR I implement embedding modules for 4-bit and 8-bit quantizations from this library. Currently, they only support `_load_from_state_dict` API and can't be saved, but I think they still can be useful.

After that, I plan to integrate this functionality into the transformers library by extending HfQuantizer functionality.

What do you think?

----------

There is also one thing I want to implement before going back to the transformers library: support for shared weights in 8-bit quantization.

While 4-bit quantization linear layer does not seem to change `self.weight` parameter during forward pass, 8-bit quantization linear layer changes it dramatically in `init_8bit_state` [method](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/774d0656110c49107559da6f53c3c9910fd3f357/bitsandbytes/nn/modules.py#L797).

So, while 4-bit embedding and linear layer can share the same `Params4bit` parameter, it is not the case for 8-bit.

I think that [this](https://github.com/galqiwi/bitsandbytes/commit/6c52fd03911ec45efdbece5d35f1e89b60367e98) patch should fix the problem, but this part of the code is very tightly coupled to everything around itself, and I need your help with advice. Do you think it can break something important I don't see?


galqiwi:
Bump

P.S. Part about shared embeddings can be discussed later in another PR or issue

matthewdouglas:
Hi @galqiwi! Thank you for the PR! I think this would be a very useful addition and will review this week. I agree that the shared embeddings can be deferred to follow up discussion/PRs.

bitsandbytes/nn/modules.py
```diff
@@ -658,6 +665,191 @@ def maybe_rearrange_weight(state_dict, prefix, local_metadata, strict, missing_k
...
+    """
+
+    def __init__(self, num_embeddings, embedding_dim, device=None, dtype=None):
+        super().__init__(num_embeddings, embedding_dim, device=device, dtype=dtype)
+        self.dtype = self.weight.data.dtype
+
+        self.weight = Int8Params(self.weight.data, has_fp16_weights=False, requires_grad=False)
+
+    def _save_to_state_dict(self, destination, prefix, keep_vars):
+        raise NotImplementedError("saving Embedding4bit module is not implemented")
```
matthewdouglas, line 695:
```suggestion
        raise NotImplementedError("saving Embedding8bit module is not implemented")
```

bitsandbytes/nn/modules.py
```diff
@@ -658,6 +665,191 @@ def maybe_rearrange_weight(state_dict, prefix, local_metadata, strict, missing_k
...
+        compressed_output_stats = F.embedding(input, row_stats.view(self.num_embeddings, 1))
+
+        output = compressed_output * (compressed_output_stats / 127.0)
+
+        return output.to(self.dtype)
+
+
+class Embedding4bit(nn.Embedding):
+    """
+    This is the base class similar to Linear4bit. It implements 4 bit 4-bit quantization algorithm presented in
```
matthewdouglas, line 717:
```suggestion
    This is the base class similar to Linear4bit. It implements the 4-bit quantization algorithm presented in
```

bitsandbytes/nn/modules.py
```diff
@@ -658,6 +665,191 @@ def maybe_rearrange_weight(state_dict, prefix, local_metadata, strict, missing_k
...
+    [QLoRA](https://arxiv.org/abs/2305.14314) for embeddings.
+
+    Quantization API is similar to Linear4bit:
+    ```python
+    import torch
+    import torch.nn as nn
+
+    from bitsandbytes.nn import Embedding8bit
```
matthewdouglas, line 725:
```suggestion
    from bitsandbytes.nn import Embedding4bit
```

matthewdouglas:
Thanks @galqiwi! Overall, this looks great! I just left a few minor nits, but otherwise happy to merge!

galqiwi:
Thank you for reviewing my PR, @matthewdouglas! I've fixed all the typos you found

matthewdouglas:
Thanks @galqiwi! This is a great contribution, and the unit tests here are really appreciated!

galqiwi:
Hi again! Are you planning on publishing new release of bnb?

PR #1294, opened 2024-07-25
script to estimate qlora mem usage

Titus-von-Koeller:
sketch of a script to estimate qlora mem usage (w/o taking FSDP into account)


AaronZLT:
Could you explain the calculation in the estimation?

Titus-von-Koeller:
@AaronZLT Sry, this was just a draft based on a formula that Tim came up with, but it didn't really prove to be useful. I don't have a good explanation off the top of my head right now and need to focus on other tasks right now, sry!

Closing this due to the above reason.

PR #1325, opened 2024-08-20
fix 4bit dtype

jiqing-feng:
Fix error check for model weight dtype.


jiqing-feng:
Hi @Titus-von-Koeller , please merge this PR for 4bit saving bug.

PR #1328, opened 2024-08-20
Cuda source cleanup , refactor and fixes

abhilash1910:
This is a draft Pr for cleanup of cuda kernels, refactor template assignments and fix some AOT issues. 
The cuda kernels have some scope of cleanup in terms of redundant code pathways and unused methods.
cc @matthewdouglas , @Titus-von-Koeller 
@TimDettmers (pinging for awareness).


matthewdouglas:
Thanks! We actually just recently were discussing some cleanup here so this fits nicely timing-wise!

Titus-von-Koeller:
@abhilash1910

Thanks for the PR, really appreciate the pro-activity in helping us clean things up!

I just had a chat with Tim and he reviewed the PR and said everything looks good to him.

I see you put the PR as draft, are there any more changes coming or can wrap things up?

abhilash1910:
Hi @Titus-von-Koeller , I am planning to wrap this up in a couple of hours from now. 
Will ping you . Thanks

csrc/ops.cu
```diff
@@ -693,9 +673,9 @@ template <typename T> void gemm_host(int m, int n, int k, T * A,  T* B,  T * out
 	//cout << m << endl;
 	//cout << n << endl;
 	//cout << k << endl;
-  //if(bits == 32)
+  if(bits == 32)
     //gemm_device<T, 32, 128><<< num_blocks, 128, 0, 0 >>>(m,  n,  k, A,  B,  out, lda, ldb, ldc);
-    //gemm_device<T, 32, 32><<< num_blocks, 32, 0, 0 >>>(m,  n,  k, A,  B,  out, lda, ldb, ldc);
+    gemm_device<T, 32, 32><<< num_blocks, 32, 0, 0 >>>(m,  n,  k, A,  B,  out, lda, ldb, ldc);
```
matthewdouglas, line 678: Is there a particular issue this is addressing? I think we're likely to deprecate this anyway, but would be good to know more about expected impact in the meantime.

abhilash1910: Yes I was thinking about removing the 32 bit gemm_device template altogether as well. Similar case is in gemm_4 bit_inference . Since the template implementation is there in kernels.cu, I had uncommented this . But in reality this & the equivalent template invocation can be done away with . What do you suggest?

matthewdouglas: I'm not opposed to keeping it for now and taking another look in a follow up PR that I'm planning. In that PR `gemm_4bit_inference` is going to be removed.

abhilash1910: > I'm not opposed to keeping it for now and taking another look in a follow up PR that I'm planning. In that PR `gemm_4bit_inference` is going to be removed.

In that case, I would suggest to keep it for now. BTW, will the interface of gemm_4bit_inference be changed? 
I think this PR is complete, barring just the nit format which I am updating.
Also a note that I might be picking up the cutlass port soon for the gemms, just going through some housekeeping for now.

abhilash1910: The lint should be fixed now . @matthewdouglas

matthewdouglas (approved):
Looks good! Thanks!

PR #1330, opened 2024-08-21
Enable certain CUDA kernels to accept specified cuda stream

jeejeelee:
FIX https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1308

By passing specified `stream` to certain kernel functions, `cudagraph` can correctly capture these kernels, enabling downstream repo  `vLLM ` to run inference in cudagraph mode, resulting in significant speed improvements for BNB models.
ping @matthewdouglas  @Titus-von-Koeller  @TimDettmers 
cc @chenqianfzh

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -438,9 +438,11 @@ def is_on_gpu(tensors):
         )
     return on_gpu
 
+
 def get_tensor_stream(tensor: Tensor) -> int:
     stream = torch.cuda.current_stream(tensor.device).cuda_stream
     return stream
+
 
 def get_ptr(A: Optional[Tensor]) -> Optional[ct.c_void_p]:
     """
@@ -985,7 +987,7 @@ def dequantize_blockwise(
                 get_ptr(out),
                 ct.c_int(quant_state.blocksize),
                 ct.c_int(A.numel()),
-                ct.c_uint64(stream)
+                ct.c_uint64(stream),
             )
         elif out.dtype == torch.float16:
             lib.cdequantize_blockwise_fp16(
@@ -995,7 +997,7 @@ def dequantize_blockwise(
                 get_ptr(out),
                 ct.c_int(quant_state.blocksize),
                 ct.c_int(A.numel()),
-                ct.c_uint64(stream)
+                ct.c_uint64(stream),
             )
         elif out.dtype == torch.bfloat16:
             lib.cdequantize_blockwise_bf16(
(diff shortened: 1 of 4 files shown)


jeejeelee:
@Titus-von-Koeller , Thank you for the feedback, I've corrected the error mentioned above. I'm verifying whether all the unit tests are passing.

jeejeelee:
On my machine with a 3090 GPU, my test results are as follows:
```
=========================================================== 13 failed, 3264 passed, 35 skipped, 1061 warnings, 16 errors in 875.66s (0:14:35) ==========================================================
```
All tests in `test_generation.py` failed due to a network connection error.
All tests in `test_triton.py` failed because my  local triton version is 3.0.0.
The other errors are likely due to precision issues. I'm not certain if these are caused by this PR

jeejeelee:
@Titus-von-Koeller  please review again,  thanks~

matthewdouglas:
@danielhanchen I believe you're directly calling some of these C-API functions in Unsloth, so I want to make sure you've got a heads up here since this changes their signatures.

bitsandbytes/functional.py
```diff
@@ -439,6 +439,11 @@ def is_on_gpu(tensors):
     return on_gpu
 
 
+def get_tensor_stream(tensor: Tensor) -> int:
+    stream = torch.cuda.current_stream(tensor.device).cuda_stream
+    return stream
+
+
```
matthewdouglas, line 446: It would be cleaner here to use `ct.c_void_p` instead of uint64. In fact `._as_parameter_` is defined on PyTorch's `Stream` wrapper, so it should just work with ctypes function calls.

E.g. change this to `return torch.cuda.current_stream(tensor.device)` and replace `ct.c_uint64(stream)` in the C invocations with just `stream`. This would effectively be the same as `ct.c_void_p(stream.cuda_stream)` and could remove the need for the `reinterpret_cast` etc.

matthewdouglas (changes requested):

matthewdouglas:
@jeejeelee Thank you for the contribution! The only nit I have is the one that I noted about using `c_void_p` instead of `uint64`. 

A few test failures in test_kbit_backprop and test_gemv_4bit is OK and not related to this PR. I see similar results on my 4090. The generation tests passed for me. Looks nice!

bitsandbytes/functional.py
jeejeelee: thank you very much, I will handling it asap

danielhanchen:
> @danielhanchen I believe you're directly calling some of these C-API functions in Unsloth, so I want to make sure you've got a heads up here since this changes their signatures.

Super thanks for the heads up!! Yep we use the C API directly!

Titus-von-Koeller:
I'll be off until Monday, @matthewdouglas  will be taking the lead. Thanks both!

matthewdouglas (approved):

PR #1344, opened 2024-08-31
Add `move_to_device` kwarg to the optimizer's `load_state_dict`

koute:
This PR makes it possible to load an optimizer checkpoint without automatically moving the optimizer's state to the GPU.

Some background as to why: I'm keeping the optimizer's state on the CPU to save on VRAM and I manually move it to the GPU as needed. Unfortunately the `load_state_dict` will move all of the optimizer's tensors to whatever device the model's parameters are currently on, which results in an OOM crash. So currently before loading an optimizer checkpoint I have to unnecessarily move my model to the CPU, call the optimizer's `load_state_dict`, and then move the model back to the GPU. With this PR I can skip this silly dance.

### DIFF
--- a/bitsandbytes/optim/optimizer.py
+++ b/bitsandbytes/optim/optimizer.py
@@ -153,12 +153,14 @@ class Optimizer8bit(torch.optim.Optimizer):
     def __setstate__(self, state):
         super().__setstate__(state)
 
-    def load_state_dict(self, state_dict):
+    def load_state_dict(self, state_dict, move_to_device=True):
         """Load an optimizer state.
 
         Arguments:
             state_dict (`dict`):
                 An optimizer state (should be returned from a call to `state_dict`) to load.
+            move_to_device (`bool`, defaults to `True`):
+                Whether to move the optimizer's state to the device.
         """
         # deepcopy, to be consistent with module API
         state_dict = deepcopy(state_dict)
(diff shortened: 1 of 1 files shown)


bitsandbytes/optim/optimizer.py
```diff
@@ -195,7 +197,8 @@ def cast(param, value):
             elif isinstance(value, dict):
                 for k, v in value.items():
                     if k in self.non_castable_tensor_keys:
-                        value[k] = v.to(param.device)
+                        if move_to_device:
+                            value[k] = v.to(param.device)
```
hansonw, line 201:
```suggestion
                        buffer = self.get_state_buffer(v, v.dtype)
                        buffer.copy_(v)
                        value[k] = buffer
```

hansonw (commented):
I encountered this issue as well but when using the paged variants of the optimizers (`load_state_dict` should be re-creating paged tensors instead of just using `.to(param.device)`).

My solution (see suggestion below) was to alter the initialization to use `self.get_state_buffer` instead. It's still kind of orthogonal to this PR (but the intent is similar) - I can submit a separate PR, but curious what the maintainers think.

bitsandbytes/optim/optimizer.py
matthewdouglas: Thanks @hansonw! This seems reasonable as a separate PR!

PR #1348, opened 2024-09-05
Nf4 reload

jiqing-feng:
Fix nf4 memory issue when using state_dict().

Still has memory issue when loading nf4 model.


jiqing-feng:
- [ ]

PR #1349, opened 2024-09-05
fix nf4 memory issue by init op_context in forward

jiqing-feng:
Fix nf4 memory issue when using state_dict().

Will open it once it's ready


Titus-von-Koeller:
Dear @jiqing-feng, please [install the pre-commit hooks](https://huggingface.co/docs/bitsandbytes/main/en/contributing#setup-pre-commit-hooks). Otherwise, you'll always have broken checks going forward and it's quick to install.

jiqing-feng:
Fix a small memory issue for nf4, the PR will not block tests from transformers.

PR #1360, opened 2024-09-16
Add AdEMAMix optimizer

matthewdouglas:
Adds support for the AdEMAMix optimizer described here: https://arxiv.org/abs/2409.03137

Includes blockwise 8bit and 32bit versions, each supporting paged operation.

AdEMAMix is a modification to Adam which introduces an additional EMA component. It is observed that AdEMAMix can forget training data at a slower pace and can reach similar loss as AdamW with significantly less training data.

~~TODO: Implement scheduler for alpha/beta3~~

### DIFF
--- a/bitsandbytes/optim/ademamix.py
+++ b/bitsandbytes/optim/ademamix.py
@@ -1,0 +1,254 @@
+from typing import Iterable, Literal, Optional, Tuple
+
+import torch
+
+import bitsandbytes.functional as F
+from bitsandbytes.optim.optimizer import Optimizer2State
+
+
+class _ReferenceAdEMAMix(torch.optim.Optimizer):
+    """
+    Reference: https://hf.co/papers/2409.03137
+    """
+
+    def __init__(
+        self,
+        params: Iterable[torch.nn.Parameter],
+        lr: float = 1e-3,
+        betas: Tuple[float, float, float] = (0.9, 0.999, 0.9999),
+        alpha: float = 5.0,
+        eps: float = 1e-8,
+        weight_decay: float = 1e-2,  # default 0.0 or 1e-2?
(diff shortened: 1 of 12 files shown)


bitsandbytes/optim/ademamix.py
```diff
@@ -0,0 +1,254 @@
...
+                    continue
+
+                grad = p.grad
+                state = self.state[p]
+
+                # State initialization
+                if len(state) == 0:
+                    # For parity with bnb implementation we combine both fast
+                    # and slow EMA stats into one stacked tensor.
+                    state["m1_m2"] = p.new_zeros((2, *p.size()))
```
matthewdouglas, line 62: This is done for ease of compatibility with the existing test suite. In most other implementations we'll see two separate buffers here.

csrc/kernels.cu
```diff
@@ -885,9 +885,16 @@ __global__ void kOptimizer32bit2State(T* g, T* p,
...
   T p_vals[NUM_PER_THREAD];
 
+
   float s1_vals[NUM_PER_THREAD];
   float s2_vals[NUM_PER_THREAD];
 
+  // AdEMAMix has an additional state buffer, which we packed
+  // into state1. We need thread-local storage here for these.
+  // TODO: Mark with [[maybe_unused]] after upgrade to min compiler.
+  float s3_vals[NUM_PER_THREAD];
```
matthewdouglas, line 895: There's a few extra memory allocations like this to support AdEMAMix. Have not confirmed if the compiler is optimizing these out for instantiations with `OPTIMIZER=ADAM`, but if not, the overhead isn't very much.

TimDettmers (approved):
This looks all good to me.

PR #1365, opened 2024-09-19
Change 8bit optimizer blocksize 2048->256; additional bf16 support

matthewdouglas:
This PR stacks on top of #1360 with an update to the blocksize for the 8bit blockwise optimizers. Additionally, bf16 support is added for RMSprop, Adagrad, and Momentum.

@TimDettmers @Titus-von-Koeller


TimDettmers (commented):
This looks all good. The only thing that we might want to check is if the error boundaries in the tests are now improved. We want to keep the absolute and relative error tight, so that if we have a slight degradation, the tests fail again.

matthewdouglas:
@TimDettmers I was able to tighten some of the tolerances.

PR #1372, opened 2024-09-26
replace tl.libdevice.llrint with tl.extra.cuda.libdevice.rint

bjmsong:


matthewdouglas:
Closing in favor of #1871 which removes this functionality instead.

PR #1383, opened 2024-10-03
[TESTS] Add global random seed for deterministic test

allrob23:
### Description

This PR introduces a global random seed set after imports in the test file. This change aims to make the tests more deterministic and easier to debug by ensuring consistent random behavior across test runs.

### Current Issues (Before this PR):

The tests currently use random without setting a fixed seed, which causes several problems:

a. Flaky Tests: Tests sometimes pass and sometimes fail due to different random values;
b. Hard to Reproduce: When a test fails, it's difficult to reproduce the exact conditions;
c. Inconsistent CI/CD: Different CI runs may have different outcomes;
d. Time Wasted: Developers spend time debugging issues that are hard to replicate.

### Solution

Set a global random seed right after imports in the test file.

### Benefits

a. Reproducibility: All test runs will use the same random values;
b. Easier Debugging: When a test fails, we can easily reproduce the issue;
c. Consistent CI/CD: Every CI run will have the same behavior;
d. Clear Intent: It's immediately obvious that we're using controlled randomness;
e. Time Saved: Less time spent on debugging non-deterministic test failures.

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -22,6 +22,8 @@ from tests.helpers import (
 torch.set_printoptions(precision=5, sci_mode=False, linewidth=120, edgeitems=20, threshold=10000)
 k = 20
 
+random.seed(42)
+
 
 def assert_all_approx_close(a, b, rtol=1e-3, atol=1e-3, count=0, throw=True):
     idx = torch.isclose(a, b, rtol=rtol, atol=atol)


matthewdouglas:
Hi @allrob23! Thank you for your suggestion and the contribution. This type of change has been applied in #1576, superseding this PR.

PR #1385, opened 2024-10-10
README: Replace special Unicode text symbols with regular characters

akx:
The release announcement from 9b3c2b34f9a0c0cc82c17bf1c14ed200dfc76c0b uses Unicode Mathematical Alphanumeric Symbols, which render in a jarringly different typeface (not to mention I wouldn't be surprised if accessibility tools have difficulties with them).

This switches them to regular characters and Markdown formatting.


matthewdouglas:
LGTM, thanks!

PR #1393, opened 2024-10-16
[Feature]Cutlass (re) integration

abhilash1910:
Hi @TimDettmers , @Titus-von-Koeller & @matthewdouglas , 
This is part of our initial discussion to (re) integrate and try out cutlass integration support for gemm device kernel. 
Currently this is a WIP as I am testing out standalone cutlass gemms (without cute) and would incorporate cute if required for performance.  Yes this took some time as was caught up in some other work , but plan to actively add/integrate this feature to cuda  backend.  In longer run , we plan to integrate this in sycl backend as well.
cc @mehdi-goli for review and feedbacks .

### DIFF
--- a/csrc/kernels.cu
+++ b/csrc/kernels.cu
@@ -3054,213 +3054,39 @@ template <int FORMAT> __global__ void kExtractOutliers(char *A, int *idx, char *
 	}
 }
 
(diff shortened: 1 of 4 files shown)


lamikr:
Will this be behind some kind of build layer so that it can be disabled for example from AMD's rocm build target which does not have direct cutlass replacement available. In rocm case, composable kernel could perhaps be used for providing a similar kinf of features what cutlass have, but it would need some porting work.

matthewdouglas:
Closing as this is stale and quite out of date.

PR #1401, opened 2024-10-24
LLM.int8() Refactoring: Part 1

### LINKED ISSUE #538: [Bug] Exception: cublasLt ran into an error! during fine-tuning LLM in 8bit mode
## Problem

Hello, I'm getting this weird cublasLt error on a lambdalabs H100 with cuda 118, pytorch 2.0.1, python3.10 Miniconda while trying to fine-tune a 3B param open-llama using LORA with 8bit loading. This only happens if we turn on 8bit loading. Lora alone or 4bit loading (qlora) works.

The same commands did work 2 weeks ago and stopped working a week ago.

I've tried bitsandbytes version 0.39.0 and 0.39.1 as prior versions don't work with H100. Source gives me a different issue as mentioned in Env section.

## Expected

No error

## Reproduce

Setup Miniconda then follow https://github.com/OpenAccess-AI-Collective/axolotl 's readme on lambdalabs and run the default open llama lora config.

## Trace

0.39.0
```bash
File "/home/ubuntu/axolotl/scripts/finetune.py", line 352, in <module>
    fire.Fire(train)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/fire/core.py", line 141, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/fire/core.py", line 475, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/fire/core.py", line 691, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)   
  File "/home/ubuntu/axolotl/scripts/finetune.py", line 337, in train
    trainer.train(resume_from_checkpoint=resume_from_checkpoint)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/trainer.py", line 1531, in train
    return inner_training_loop(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/trainer.py", line 1795, in _inner_training_loop
    tr_loss_step = self.training_step(model, inputs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/trainer.py", line 2640, in training_step
    loss = self.compute_loss(model, inputs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/trainer.py", line 2665, in compute_loss
    outputs = model(**inputs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/utils/operations.py", line 553, in forward
    return model_forward(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/utils/operations.py", line 541, in __call__
    return convert_to_fp32(self.model_forward(*args, **kwargs))
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/amp/autocast_mode.py", line 14, in decorate_autocast
    return func(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/peft/peft_model.py", line 827, in forward
    return self.base_model(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/hooks.py", line 165, in new_forward
    output = old_forward(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py", line 691, in forward
    outputs = self.model(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/hooks.py", line 165, in new_forward
    output = old_forward(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py", line 579, in forward
    layer_outputs = decoder_layer(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/hooks.py", line 165, in new_forward
    output = old_forward(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py", line 293, in forward
    hidden_states, self_attn_weights, present_key_value = self.self_attn(
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/accelerate/hooks.py", line 165, in new_forward
    output = old_forward(*args, **kwargs)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/transformers/models/llama/modeling_llama.py", line 195, in forward
    query_states = self.q_proj(hidden_states).view(bsz, q_len, self.num_heads, self.head_dim).transpose(1, 2)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs) 
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/peft/tuners/lora.py", line 942, in forward
    result = super().forward(x)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/nn/modules.py", line 402, in forward
    out = bnb.matmul(x, self.weight, bias=self.bias, state=self.state)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/autograd/_functions.py", line 562, in matmul
    return MatMul8bitLt.apply(A, B, out, bias, state)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/autograd/function.py", line 506, in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/autograd/_functions.py", line 400, in forward
    out32, Sout32 = F.igemmlt(C32A, state.CxB, SA, state.SB)
  File "/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/functional.py", line 1781, in igemmlt
    raise Exception('cublasLt ran into an error!')
Exception: cublasLt ran into an error!
```

## Env

`python -m bitsandbytes`

- on main branch: I get error same as here https://github.com/TimDettmers/bitsandbytes/issues/382

- on 0.39.0

```bash
bin /home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/cuda_setup/main.py:149: UserWarning: Found duplicate ['libcudart.so', 'libcudart.so.11.0', 'libcudart.so.12
.0'] files: {PosixPath('/home/ubuntu/miniconda3/envs/py310/lib/libcudart.so.11.0'), PosixPath('/home/ubuntu/miniconda3/envs/py310/lib/libcudart.so')}.. We'll flip a coin and try one of
 these, in order to fail forward.
Either way, this might cause trouble in the future:
If you get `CUDA error: invalid device function` errors, the above might be the cause and the solution is to make sure only one ['libcudart.so', 'libcudart.so.11.0', 'libcudart.so.12.0
'] in the paths that we search based on your env.
  warn(msg)
CUDA SETUP: CUDA runtime path found: /home/ubuntu/miniconda3/envs/py310/lib/libcudart.so.11.0
CUDA SETUP: Highest compute capability among GPUs detected: 9.0
CUDA SETUP: Detected CUDA version 118
CUDA SETUP: Loading binary /home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so...
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
++++++++++++++++++ BUG REPORT INFORMATION ++++++++++++++++++
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

+++++++++++++++++++ ANACONDA CUDA PATHS ++++++++++++++++++++
/home/ubuntu/miniconda3/envs/py310/lib/libcudart.so
/home/ubuntu/miniconda3/envs/py310/lib/stubs/libcuda.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/lib/libtorch_cuda.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/lib/libtorch_cuda_linalg.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/torch/lib/libc10_cuda.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda110.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda114.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda120.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda111_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda112_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda116.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda117.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda120_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda118_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda114_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda112.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda111.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda110_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda115_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda116_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda121_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda121.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda115.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda113_nocublaslt.so
/home/ubuntu/miniconda3/envs/py310/lib/python3.10/site-packages/bitsandbytes/libbitsandbytes_cuda113.so
/home/ubuntu/miniconda3/envs/py310/nsight-compute/2023.1.1/target/linux-desktop-glibc_2_19_0-ppc64le/libcuda-injection.so
/home/ubuntu/miniconda3/envs/py310/nsight-compute/2023.1.1/target/linux-desktop-glibc_2_11_3-x64/libcuda-injection.so
/home/ubuntu/miniconda3/envs/py310/nsight-compute/2023.1.1/target/linux-desktop-t210-a64/libcuda-injection.so

++++++++++++++++++ /usr/local CUDA PATHS +++++++++++++++++++


+++++++++++++++ WORKING DIRECTORY CUDA PATHS +++++++++++++++


++++++++++++++++++ LD_LIBRARY CUDA PATHS +++++++++++++++++++
+++++++++++ /usr/lib/x86_64-linux-gnu CUDA PATHS +++++++++++
/usr/lib/x86_64-linux-gnu/libcudart.so
/usr/lib/x86_64-linux-gnu/stubs/libcuda.so
/usr/lib/x86_64-linux-gnu/libcuda.so

++++++++++++++++++++++++++ OTHER +++++++++++++++++++++++++++
COMPILED_WITH_CUDA = True
COMPUTE_CAPABILITIES_PER_GPU = ['9.0']
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
++++++++++++++++++++++ DEBUG INFO END ++++++++++++++++++++++
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Running a quick check that:
    + library is importable
    + CUDA function is callable


WARNING: Please be sure to sanitize sensible info from any such env vars!

SUCCESS!
Installation was successful!
```

```bash
$ nvcc -V
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2022 NVIDIA Corporation
Built on Wed_Jun__8_16:49:14_PDT_2022
Cuda compilation tools, release 11.7, V11.7.99
Build cuda_11.7.r11.7/compiler.31442593_0
```

## Misc

All related issues: 
- lambdalabs h100 https://github.com/tloen/alpaca-lora/issues/174
- older gpu: https://github.com/oobabooga/text-generation-webui/issues/379
- says due to card issue: https://github.com/Facico/Chinese-Vicuna/issues/3
- 3070: https://github.com/Facico/Chinese-Vicuna/issues/41
- 3060: https://github.com/oobabooga/text-generation-webui/issues/569
- no resolution: https://github.com/Lightning-AI/lit-llama/issues/315
- low vram https://github.com/deep-diver/LLM-As-Chatbot/issues/16
- no resolution: https://github.com/huggingface/transformers/issues/21371

Also tried install `cudatoolkit` via conda.

matthewdouglas:
This PR is the initial phase of a set of changes aimed at improving the LLM.int8() implementation.

@TimDettmers @Titus-von-Koeller 

#### Primary Purpose
* Introduces support for Hopper (H100, H200) GPUs.
* Resolves #538 

#### Enhancements
* Removes the usage of Turing and Ampere specific memory layouts while retaining compatibility across sm_75 through sm_89.
  * Simplifies the code and surface area needing to be maintained.
  * Reduced overhead by removing layout transformation operations.
* Removes the separate `NO_CUBLASLT` build while retaining compatibility for targets below sm_75. <sup>verification in progress</sup>
  * This simplifies building and packaging, and **trims the size of binary wheels in ~half**.
* Support for CUDA Graph tracing to bring parity with 4bit.
* Improved kernels for inference:
  * Fused kernel for activation scale calibration and quantization. (Exposed as op `F.int8_vectorwise_quant`)
  * Other kernels simplified to operate with row-major data.
* Makes many unit tests more reliable with increased determinism.

#### Deprecations
The following functions from `bitsandbytes` are deprecated:
```
mm_cublas
bmm_cublas
matmul_cublas
```

The following functions from `bitsandbytes.functional` are deprecated:
```
_mul
arange
dequant_min_max
dequantize_no_absmax
extract_outliers
get_special_format_str
get_tensor_stream (moved to internal API)
get_transform_buffer
get_transform_func
mm_dequant (replacement: int8_mm_dequant)
igemmlt (replacement: int8_linear_matmul)
nvidia_transform
post_call
pre_call
transform
quantize_no_absmax
vectorwise_dequant
vectorwise_quant (~replacement: int8_vectorwise_quant)
vectorwise_mm_dequant (~replacement: int8_mm_dequant)
```

Further testing and benchmarking will be coming. At the moment unit tests are passing.

### Next steps

- [x] Clean up and reorganize unit tests
- [x] Documentation for public APIs
- [x] Ensure fallback path for shapes that don't work well with cuBLASLt (i.e. m/k not multiples of 4).
- [x] Add an int8 dequantize op
- [ ] ~Further improvement of sparse decomposition performance~ (Deferred to future PRs)
- [x] Conduct profiling, benchmarks, and evaluations
  - [x] Build benchmark/evaluation scripts
  - [x] Prepare analysis of results


bitsandbytes/autograd/_functions.py
```diff
@@ -1,6 +1,5 @@
 from dataclasses import dataclass
-from functools import reduce  # Required in Python 3
-import operator
+from math import prod
```
matthewdouglas, line 2: We support Python 3.8+ only, so use the builtin.

bitsandbytes/autograd/_functions.py
```diff
@@ -245,11 +238,11 @@ class MatmulLtState:
     _tile_indices: Optional[torch.Tensor] = None
     force_no_igemmlt: bool = False
     CB = None
-    CxB = None
+    CxB = None  # TODO: Deprecate/remove
```
matthewdouglas, line 241: This won't be used anymore but I'm not sure of the side-effects of removing these properties either. Could be downstream integrations accessing them, maybe used in serialization etc. Any tips/thoughts here are welcome.

bitsandbytes/autograd/_functions.py
```diff
@@ -283,12 +276,17 @@ def tile_indices(self):
...
-        using_igemmlt = supports_igemmlt(A.device) and not state.force_no_igemmlt
+    def forward(
+        ctx: torch.autograd.function.FunctionCtx,
+        A: torch.Tensor,
+        B: torch.Tensor,
+        out=None,
+        bias: Optional[torch.Tensor] = None,
+        state: MatmulLtState = None,
+    ):
+        state = state or MatmulLtState()
```
matthewdouglas, line 288: This seemed like a bug/oversight to default `state=MatmulLtState` so I changed it here to use a new empty state.

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +299,108 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
+        if ctx.needs_input_grad[1]:
+            # Slower path
+            CA, CAt, SCA, SCAt, coo_tensorA = F.double_quant(A.to(torch.float16), threshold=state.threshold)
         else:
-            if not state.has_fp16_weights and state.CxB is None and using_igemmlt:
-                state.CxB, state.SB = F.transform(state.CB, to_order=formatB)
-            subA = None
+            # Fast path
+            CA, SCA, coo_tensorA = F.int8_vectorwise_quant(A.to(torch.float16), threshold=state.threshold)
+            CAt = SCAt = None
```
matthewdouglas, line 318: I've been focused on improving for inference, so the new fused kernel in this path only gathers the scales and quantizes row-wise.

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +299,108 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
-                outliers = F.extract_outliers(state.CxB, state.SB, state.idx.int())
-            else:
-                outliers = state.CB[:, state.idx.long()].clone()
 
-            state.subB = (outliers * state.SCB.view(-1, 1) / 127.0).t().contiguous().to(A.dtype)
-            CA[:, state.idx.long()] = 0
-            CAt[:, state.idx.long()] = 0
-            subA = A[:, state.idx.long()]
+                # 2. Quantize B
+                state.CB, state.SCB, _ = F.int8_vectorwise_quant(B.to(torch.float16))
```
matthewdouglas, line 332: Mostly for simplicity I'm not using `CBt` and `SCBt` in backward anymore and just doing the compute in fp16.

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +299,108 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
-                output = output.to(A.dtype)
-            else:  # apply bias separately
-                output = F.mm_dequant(out32, Sout32, SCA, state.SCB, bias=None)
-                output = output.to(A.dtype).add_(bias)
+        if state.threshold > 0.0 and coo_tensorA is not None:
+            state.idx = torch.unique(coo_tensorA._indices()[1]).long()
+
+            # Zero out the outliers in the transposed 8bit inputs.
+            if CAt is not None:
+                CAt[:, state.idx] = 0
```
matthewdouglas, line 347: We skip this for inference now as it's also not needed.

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +299,108 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
 
+            # Extract the input outliers in original precision
+            subA = A[:, state.idx]
+
+            # Extract the corresponding weights
+            if state.has_fp16_weights:
+                state.subB = B[:, state.idx].t()  # .contiguous()
+            else:
+                outliers = state.CB[:, state.idx]  # .clone()
+                state.subB = (7.874016e-3 * outliers * state.SCB.view(-1, 1)).t().to(A.dtype)
```
matthewdouglas, line 357: Equivalent to dividing by 127, but multiplication is faster.

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +299,108 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
 
-        ctx.formatB = formatB
         ctx.grad_shape = input_shape
         ctx.dtype_A, ctx.dtype_B, ctx.dtype_bias = A.dtype, B.dtype, None if bias is None else bias.dtype
 
         if any(ctx.needs_input_grad[:2]):
             ctx.tensors = (CAt, subA, A)
             ctx.tensor_states = (SCAt, state.idx)
         else:
-            ctx.tensors = [None, None, A]
```
matthewdouglas, line 428: Similar idea to #1270 - we shouldn't need this here, and my assumption is that it is just an oversight. @TimDettmers

bitsandbytes/functional.py
```diff
@@ -421,41 +414,36 @@ def create_quantile_map(A, total_bits=8):
...
-        if t is None:
-            continue  # NULL pointers are fine
-        is_paged = getattr(t, "is_paged", False)
-        on_gpu &= t.device.type == "cuda" or is_paged
-        if not is_paged:
+        # NULL pointers and paged tensors are OK.
+        if t is not None and not getattr(t, "is_paged", False):
+            on_gpu &= t.is_cuda
             gpu_ids.add(t.device.index)
+
```
matthewdouglas, line 431: This isn't specific for int8, but while I was profiling I noticed an opportunity to slightly improve some of the overhead here.

bitsandbytes/functional.py
```diff
@@ -421,41 +414,36 @@ def create_quantile_map(A, total_bits=8):
...
         )
     return on_gpu
 
 
-def get_tensor_stream(tensor: Tensor) -> torch.cuda.Stream:
-    stream = torch.cuda.current_stream(tensor.device)
-    return stream
+def get_tensor_stream(tensor: Tensor) -> ct.c_void_p:
+    # We use the raw stream for performance reasons.
+    return ct.c_void_p(torch._C._cuda_getCurrentRawStream(tensor.device.index))
```
matthewdouglas, line 446: There was far more overhead with constructing a `torch.cuda.Stream()` for this than I would have thought. Certainly more than I would have liked. So while not pretty, this is another performance improvement that reaches beyond int8.

bitsandbytes/functional.py
```diff
@@ -2302,179 +2274,171 @@ def batched_igemm(
...
-
-    if dimsA == 2 and out is None:
-        out, Sout = get_transform_buffer((shapeA[0], shapeB[0]), dtype, A.device, "col32", "row")
-    elif dimsA == 3 and out is None:
-        out, Sout = get_transform_buffer((shapeA[0], shapeA[1], shapeB[0]), dtype, A.device, "col32", "row")
-
-    assert dimsB != 3, "len(B.shape)==3 not supported"
-    assert A.device.type == "cuda"
-    assert B.device.type == "cuda"
+def igemmlt(A, B, out=None, Sout=None, dtype=torch.int32):
```
matthewdouglas, line 2277: This is the big int8 change that enables most of the other simplifications.

bitsandbytes/functional.py
```diff
@@ -2552,94 +2516,64 @@ def coo_zeros(rows, cols, nnz, device, dtype=torch.half):
...
 
-    coo_tensor = None
-    ptrA = get_ptr(A)
-    ptrColStats = get_ptr(col_stats)
-    ptrRowStats = get_ptr(row_stats)
-    ptrOutCol = get_ptr(out_col)
-    ptrOutRow = get_ptr(out_row)
+        # A.masked_fill(A.abs() < threshold, 0.0).to_sparse_coo()
+        # coo_tensor = extract_outliers_new(A, threshold)
+        coo_tensor = torch.masked_fill(A, A.abs() < threshold, 0.0).to_sparse_coo()
```
matthewdouglas, line 2560: This is the slowest part right now.

csrc/kernels.cu
```diff
@@ -722,24 +717,25 @@ __global__ void kDequantizeBlockwise(float *code, unsigned char * A, float * abs
...
     else
     {
-      valid_items_load = n - i > TILE_SIZE ? TILE_SIZE : n - i;
-      valid_items_store = n - i > TILE_SIZE ? TILE_SIZE : n - i;
+      valid_items_load = min(TILE_SIZE, n - i);
+      valid_items_store = valid_items_load;
     }
-    local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH)/(blocksize)]);
+    local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH) >> (31 - __clz(blocksize))]);
+    //local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH)/(blocksize)]);
```
matthewdouglas, line 733: Micro-optimization here. Considering reverting in favor of legibility.

csrc/kernels.cu
```diff
@@ -2134,6 +2130,106 @@ kOptimizerStatic8bit1StateBlockwise(T* p, T* __restrict__ const g, unsigned char
...
+__launch_bounds__(1024, BNB_MAX_THREADS_PER_SM / 1024)
+__global__ void kInt8VectorQuant(T * __restrict__ A, int8_t* out, float* rowStats, float threshold, int rows, int cols) {
+  using BlockReduceT = cub::BlockReduce<float, THREADS>;
+
+  // One block per row.
+  // Threads load column values in a striped arrangement.
+  // e.g. t0 reads row[0], row[0+nthreads], ..
+  // and  t1 reads row[1], row[1+nthreads], ..
+  // Each thread will determine its local absmax.
+  // We then do a blockwise reduction to determine the row's absmax.
```
matthewdouglas, line 2148: I'm considering a future iteration with a different approach for GEMV or smaller batches where more blocks work cooperatively on the same row. At the moment this works pretty well though.

csrc/kernels.cu
```diff
@@ -3525,17 +3568,19 @@ template <typename T, int THREADS, int BITS> __global__ void kgemm_4bit_inferenc
   __shared__ T quant_map[16];
 	T local_absmax = T(0.0f);
 
-  for(int i = threadIdx.x; i < 16; i++)
-    quant_map[i] = T(datatype[i]);
+  if (threadIdx.x < 16)
+    quant_map[threadIdx.x] = T(__ldg(&datatype[threadIdx.x]));
+  //for(int i = threadIdx.x; i < 16; i++)
+    //quant_map[i] = T(__ldg(&datatype[i]));
```
matthewdouglas, line 3574: Not int8 but another small 4bit change that wanted to sneak its way in. @TimDettmers I'm just looking for a sanity check here that this makes sense.

bitsandbytes/autograd/_functions.py
```diff
@@ -1,20 +1,14 @@
 from dataclasses import dataclass
-from functools import reduce  # Required in Python 3
-import operator
+from math import prod
 from typing import Callable, Optional, Tuple
 import warnings
 from warnings import warn
 
 import torch
+from typing_extensions import deprecated
```
akx, line 8: I'd avoid a new dependency just for this, but it's very likely `typing_extensions` will be in the venvs for users anyway...

bitsandbytes/autograd/_functions.py
```diff
@@ -283,12 +284,17 @@ def tile_indices(self):
...
     @staticmethod
-    def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
-        using_igemmlt = supports_igemmlt(A.device) and not state.force_no_igemmlt
+    def forward(
+        ctx: torch.autograd.function.FunctionCtx,
+        A: torch.Tensor,
+        B: torch.Tensor,
+        out=None,
+        bias: Optional[torch.Tensor] = None,
+        state: MatmulLtState = None,
```
akx, line 294:
```suggestion
        state: Optional[MatmulLtState] = None,
```

bitsandbytes/autograd/_functions.py
akx: What about e.g. bf16?

bitsandbytes/autograd/_functions.py
akx: That should be a comment in the code. The magic number is opaque to readers – I'd maybe even make it a constant `ONE_OVER_127 = 1 / 127` and use that...

bitsandbytes/autograd/_functions.py
```diff
@@ -301,146 +307,100 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState):
...
+            ctx.tensors = [None, None, None]
             ctx.tensor_states = (None, None)
             ctx.save_for_backward(None, None)
 
-        clone_func = torch.clone if len(output_shape) == 3 else lambda x: x
-        return clone_func(output.view(output_shape))
+        output_shape = (*input_shape[:-1], state.CB.shape[0])
+
+        if len(input_shape) == 3:
+            return output.reshape(output_shape)  # .clone()
```
akx, line 391: What does the comment mean?

bitsandbytes/autograd/_functions.py
```diff
@@ -452,35 +412,29 @@ def backward(ctx, grad_output):
         if len(grad_output.shape) == 3:
             grad_output = grad_output.reshape(-1, grad_output.shape[-1]).contiguous()
 
-        Cgrad, Cgradt, SCgrad, SCgradt, coo_tensor = F.double_quant(grad_output.to(torch.float16))
+        # if req_gradB:
+        # grad_B = torch.matmul(grad_output.t(), A)
+        # if state.threshold > 0.0 and subA is not None:
+        #     grad_B[:, idx] += torch.matmul(grad_output.t(), subA)
+        # Cgrad, Cgradt, SCgrad, SCgradt, _ = F.double_quant(grad_output.to(torch.float16))
```
akx, line 419: Dead code?

bitsandbytes/autograd/_functions.py
```diff
@@ -452,35 +412,29 @@ def backward(ctx, grad_output):
...
-                if state.CxBt is None:
-                    state.CxBt, state.SBt = F.transform(state.CBt, to_order=formatB, transpose=True)
-                gradA32, SgradA32 = F.igemmlt(C32grad, state.CxBt, Sgrad, state.SBt)
-                grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape).to(ctx.dtype_A)
-
-            elif state.CB is not None:
+            # grad_output @ B.T
+            # if state.CBt is not None:
+            #    gradA32, SgradA32 = F.igemmlt(Cgrad, state.CBt.t())
+            #    grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape).to(ctx.dtype_A)
```
akx, line 432: Dead code?

bitsandbytes/cextension.py
```diff
@@ -67,6 +63,9 @@ def __init__(self, lib: ct.CDLL):
     def __getattr__(self, item):
         return getattr(self._lib, item)
 
+    def __getitem__(self, item):
+        return getattr(self._lib, item)
```
akx, line 67: Why is this required? I don't think the library should look mapping-like.

`getattr(lib, "foo")` would work.

bitsandbytes/functional.py
```diff
@@ -421,41 +417,36 @@ def create_quantile_map(A, total_bits=8):
...
-    if major <= 7:
-        return "col_turing"
-    if major == 8:
-        return "col_ampere"
-    return "col_turing"
+    return "row"
 
 
-def is_on_gpu(tensors):
+def is_on_gpu(tensors: Iterable[torch.Tensor]):
```
akx, line 425: Since you check for `t is not None`, I think the signature should note that nones are OK:
```suggestion
def is_on_gpu(tensors: Iterable[Optional[torch.Tensor]]):
```

bitsandbytes/functional.py
akx: This changing the rtype for a public function, which is probably not OK.

I'd add a new `_get_raw_tensor_stream` function and use that...

bitsandbytes/functional.py
```diff
@@ -858,49 +861,39 @@ def quantize_blockwise(
...
+            torch.float16: "cquantize_blockwise_fp16",
+        }
+
+        if A.dtype not in fn_map.keys():
+            raise ValueError(f"Blockwise quantization only supports 16/32-bit floats, but got {A.dtype}")
+
+        fn = fn_map[A.dtype]
+
+        with torch.cuda.device_of(A):
+            lib[fn](
```
akx, line 888:
```suggestion
        fn_map = {
            torch.float32: "cquantize_blockwise_fp32",
            torch.bfloat16: "cquantize_blockwise_bf16",
            torch.float16: "cquantize_blockwise_fp16",
        }

        fn = getattr(lib, fn_map.get(A.dtype), None)
        if not fn:
             raise ValueError(f"Blockwise quantization only supports 16/32-bit floats, but got {A.dtype}")

        with torch.cuda.device_of(A):
            fn(
```

to avoid the silly `lib[fn]`. `fn_map` should probably also be module-level so it isn't recomputed in this function over and over again.

bitsandbytes/functional.py
```diff
@@ -985,47 +978,34 @@ def dequantize_blockwise(
...
+            torch.float16: "cdequantize_blockwise_fp16",
+        }
+
+        if out.dtype not in fn_map.keys():
+            raise ValueError(f"Blockwise quantization only supports 16/32-bit floats, but got {out.dtype}")
+
+        fn = fn_map[out.dtype]
+
+        with torch.cuda.device_of(A):
+            lib[fn](
```
akx, line 1000: Same comment here as above.

bitsandbytes/functional.py
```diff
@@ -1197,68 +1176,72 @@ def quantize_4bit(
...
         if quant_type == "fp4":
-            lib.cquantize_blockwise_fp32_fp4(
-                get_ptr(None),
-                get_ptr(A),
-                get_ptr(absmax),
-                get_ptr(out),
-                ct.c_int32(blocksize),
-                ct.c_int(n),
-            )
+            with torch.cuda.device_of(A):
```
akx, line 1182: This looks like the `with torch.cuda.device_of(A):` block should be wrapping the whole thing, not each invocation of the function.

Looks like this could also use the mapping approach championed above:

```
fn_map = {
  (torch.float32, "fp4"): "cquantize_blockwise_fp32_fp4",
  ...
}
```

bitsandbytes/functional.py
```diff
@@ -1376,83 +1359,87 @@ def dequantize_4bit(
...
-                get_ptr(None),
-                get_ptr(A),
-                get_ptr(absmax),
-                get_ptr(out),
-                ct.c_int(quant_state.blocksize),
-                ct.c_int(n),
-                stream,
-            )
+            with torch.cuda.device_of(A):
+                lib.cdequantize_blockwise_fp32_fp4(
```
akx, line 1367: Same with-wrap-and-mapping comment as above.

bitsandbytes/functional.py
```diff
@@ -1376,83 +1359,87 @@ def dequantize_4bit(
...
     else:
         raise ValueError(f"Blockwise quantization only supports 16/32-bit floats, but got {A.dtype}")
-    post_call(A.device)
 
-    is_transposed = True if A.shape[0] == 1 else False
+    is_transposed = A.shape[0] == 1
     if is_transposed:
         return out.t()
     else:
         return out
```
akx, line 1439:
```suggestion
    if A.shape[0] == 1:  # is transposed, transpose back
        return out.t()
    return out
```

bitsandbytes/functional.py
```diff
@@ -2008,10 +2004,9 @@ def gemv_4bit(
     transposed_B=False,
     state=None,
 ):
-    prev_device = pre_call(A.device)
     # sout = check_matmul(A, B, out, transposed_A, transposed_B, expected_type=A.dtype)
     if state is None:
-        raise ValueError("state cannot None. gem_4bit( ) requires the state from quantize_4bit( )")
+        raise ValueError("state cannot None. gemv_4bit() requires the state from quantize_4bit()")
```
akx, line 2009:
```suggestion
        raise ValueError("state cannot be None. gemv_4bit() requires the state from quantize_4bit()")
```

bitsandbytes/functional.py
```diff
@@ -2045,62 +2040,64 @@ def gemv_4bit(
...
-                ldb,
-                ldc,
-                ct.c_int32(state.blocksize),
-                stream,
-            )
+
+    with torch.cuda.device_of(A):
+        if B.dtype in [torch.uint8, torch.bfloat16, torch.float16, torch.float32]:
+            if A.dtype == torch.float16:
+                lib.cgemm_4bit_inference_naive_fp16(
```
akx, line 2047: This too could use the mapping thing.

bitsandbytes/functional.py
akx: Might be nice to add type annotations here, both for the args and rtype.

bitsandbytes/functional.py
```diff
@@ -2302,179 +2299,171 @@ def batched_igemm(
...
-            has_error = lib.cigemmlt_turing_8(ptr, m, n, k, ptrA, ptrB, ptrC, ptrRowScale, lda, ldb, ldc)
-    elif formatB == "col_ampere":
+
+    with torch.cuda.device_of(A):
+        ctx = CUBLAS_Context.get_instance().get_context(A.device)
+        ptrA = get_ptr(A)
+        ptrB = get_ptr(B)
+        ptrC = get_ptr(out)
+        ptrRowScale = get_ptr(None)
+        m, n, k, lda, ldb, ldc = map(ct.c_int32, (m, n, k, lda, ldb, ldc))
```
akx, line 2350: This measures 12% faster in a microbenchmark.

```suggestion
        m = ct.c_int32(m)
        n = ct.c_int32(n)
        k = ct.c_int32(k)
        lda = ct.c_int32(lda)
        ldb = ct.c_int32(ldb)
        ldc = ct.c_int32(ldc)
```

bitsandbytes/functional.py
```diff
@@ -2302,179 +2299,171 @@ def batched_igemm(
...
     if has_error:
-        print(f"A: {shapeA}, B: {shapeB}, C: {Sout[0]}; (lda, ldb, ldc): {(lda, ldb, ldc)}; (m, n, k): {(m, n, k)}")
-        raise Exception("cublasLt ran into an error!")
-
-    torch.cuda.set_device(prev_device)
+        raise RuntimeError(
+            f"cublasLt ran into an error!\n"
+            f"\tA: {shapeA}, B: {shapeB}, C: {Sout[0]}\n"
+            f"\t(lda, ldb, ldc): {(lda, ldb, ldc)}\n"
+            f"\t(m, n, k): {(m, n, k)}"
```
akx, line 2366: Since we only support py3.8+, could use:

```suggestion
            f"cublasLt ran into an error!\n"
            f"\t{shapeA=}, {shapeB=}, {Sout[0]=}\n"
            f"\t{(lda, ldb, ldc)=}\n"
            f"\t{(m, n, k)=}"
```

bitsandbytes/functional.py
```diff
@@ -2302,179 +2299,171 @@ def batched_igemm(
...
-def get_colrow_absmax(A, row_stats=None, col_stats=None, nnz_block_ptr=None, threshold=0.0):
-    assert A.dtype == torch.float16
-    device = A.device
+def get_colrow_absmax(
+    A: torch.Tensor,
+    row_stats: torch.Tensor = None,
+    col_stats: torch.Tensor = None,
+    nnz_block_ptr: torch.Tensor = None,
+    threshold=0.0,
+):
```
akx, line 2414: Rtype?

bitsandbytes/functional.py
```diff
@@ -2302,179 +2299,171 @@ def batched_igemm(
...
+        if col_stats is None:
+            # shape [cols]; unsqueeze(0) gives [1,cols]
+            col_stats = absA.amax(dim=0, keepdim=False).float()
 
-    if threshold > 0.0:
-        nnz_block_ptr.cumsum_(0)
+    return row_stats, col_stats, outlier_mask
+
+
+def get_row_absmax(A, threshold=0.0):
```
akx, line 2440: Rtype?

bitsandbytes/functional.py
```diff
@@ -2552,96 +2541,61 @@ def coo_zeros(rows, cols, nnz, device, dtype=torch.half):
     return COOSparseTensor(rows, cols, nnz, rowidx, colidx, values)
 
 
-def double_quant(A, col_stats=None, row_stats=None, out_col=None, out_row=None, threshold=0.0):
-    device = A.device
-    assert A.dtype == torch.half
-    assert device.type == "cuda"
-    prev_device = pre_call(A.device)
+def double_quant(A: torch.Tensor, col_stats=None, row_stats=None, out_col=None, out_row=None, threshold=0.0):
```
akx, line 2544: Rest of the types?

bitsandbytes/functional.py
```diff
@@ -2690,7 +2644,22 @@ def transform(A, to_order, from_order="row", out=None, transpose=False, state=No
     return out, new_state
 
 
-def spmm_coo(cooA, B, out=None):
+def spmm_coo(cooA: Union[COOSparseTensor, torch.Tensor], B: torch.Tensor, out: torch.Tensor = None):
+    if not isinstance(cooA, COOSparseTensor):
+        assert (
+            cooA.is_sparse and cooA.layout == torch.sparse_coo
+        ), "Tensor must be `COOSparseTensor or a PyTorch COO tensor."
```
akx, line 2651: I wouldn't use an assert in runtime code, since they will be wiped out by `python -O`.
```suggestion
        if not (cooA.is_sparse and cooA.layout == torch.sparse_coo):
            raise ValueError("Tensor must be a `COOSparseTensor` or a PyTorch COO tensor.")
```

bitsandbytes/functional.py
```diff
@@ -2690,7 +2644,22 @@ def transform(A, to_order, from_order="row", out=None, transpose=False, state=No
...
+
+        # Convert to custom COOSparseTensor
+        cooA = COOSparseTensor(
+            rows=cooA.shape[0],
+            cols=cooA.shape[1],
+            nnz=cooA._nnz(),
+            rowidx=cooA.indices()[0].int(),
+            colidx=cooA.indices()[1].int(),
+            values=cooA.values(),
+        )
```
akx, line 2661: Might as well spare a couple calls and use locals. This will crash if the shape isn't 2D, so if that isn't OK, add a suitable `[:2]`...
```suggestion
        rows, cols = cooA.shape
        indices = cooA.indices()
        cooA = COOSparseTensor(
            rows=rows,
            cols=cols,
            nnz=cooA._nnz(),
            rowidx=indices[0].int(),
            colidx=indices[1].int(),
            values=cooA.values(),
        )
```

bitsandbytes/nn/modules.py
```diff
@@ -481,10 +481,8 @@ def forward(self, x: torch.Tensor):
             x = x.to(self.compute_dtype)
 
         bias = None if self.bias is None else self.bias.to(self.compute_dtype)
-        out = bnb.matmul_4bit(x, self.weight.t(), bias=bias, quant_state=self.weight.quant_state)
-
-        out = out.to(inp_dtype)
 
+        out = bnb.matmul_4bit(x, self.weight.t(), bias=bias, quant_state=self.weight.quant_state).to(inp_dtype)
         return out
```
akx, line 486: While you're at it:
```suggestion
        return bnb.matmul_4bit(x, self.weight.t(), bias=bias, quant_state=self.weight.quant_state).to(inp_dtype)
```

bitsandbytes/nn/modules.py
```diff
@@ -1008,12 +1003,9 @@ def forward(self, x: torch.Tensor):
 
         out = bnb.matmul(x, self.weight, bias=self.bias, state=self.state)
 
-        if not self.state.has_fp16_weights:
-            if self.state.CB is not None and self.state.CxB is not None:
-                # we converted 8-bit row major to turing/ampere format in the first inference pass
-                # we no longer need the row-major weight
-                del self.state.CB
-                self.weight.data = self.state.CxB
+        if not self.state.has_fp16_weights and self.state.CB is not None:
```
akx, line 1006: I think keeping the comment here might be a good idea.

bitsandbytes/research/autograd/_functions.py
```diff
@@ -186,7 +186,9 @@ class SwitchBackBnb(torch.autograd.Function):
     @staticmethod
     # TODO: the B008 on the line below is a likely bug; the current implementation will
     #       have each SwitchBackBnb instance share a single MatmulLtState instance!!!
-    def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState()):  # noqa: B008
+    def forward(ctx, A, B, out=None, bias=None, state: MatmulLtState = None):
```
akx, line 189:
```suggestion
    def forward(ctx, A, B, out=None, bias=None, state: Optional[MatmulLtState] = None):
```

bitsandbytes/research/autograd/_functions.py
```diff
@@ -186,7 +186,9 @@ class SwitchBackBnb(torch.autograd.Function):
     @staticmethod
     # TODO: the B008 on the line below is a likely bug; the current implementation will
     #       have each SwitchBackBnb instance share a single MatmulLtState instance!!!
```
akx, line 188: Remove this TODO since you fixed the bug?
```suggestion
```

bitsandbytes/research/autograd/_functions.py
```diff
@@ -216,25 +217,22 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState()):  # noqa: B00
...
+        CA, CAt, SCA, SCAt, outlier_cols = F.double_quant(A.to(torch.float16), threshold=state.threshold)
 
-        if state.threshold > 0.0 and coo_tensorA is not None:
+        if state.threshold > 0.0 and outlier_cols is not None:
             if state.has_fp16_weights:
-                idx = torch.unique(coo_tensorA.colidx).long()
+                idx = outlier_cols
                 CA[:, idx] = 0
-                CAt[:, idx] = 0
+                # CAt[:, idx] = 0
```
akx, line 226: Dead code?

bitsandbytes/research/autograd/_functions.py
```diff
@@ -245,34 +243,28 @@ def forward(ctx, A, B, out=None, bias=None, state=MatmulLtState()):  # noqa: B00
...
-            #    state.idx = state.outlier_pool.get_current_outlier_idx().to(A.device)
-            # else:
-            #    state.idx = outlier_idx
-            outliers = F.extract_outliers(state.CxB, state.SB, state.idx.int())
+            # outliers = F.extract_outliers(state.CxB, state.SB, state.idx.int())
+            outliers = state.CB[:, state.idx.long()].clone()
             state.subB = (outliers * state.SCB.view(-1, 1) / 127.0).t().contiguous().to(A.dtype)
             CA[:, state.idx.long()] = 0
-            CAt[:, state.idx.long()] = 0
+            # CAt[:, state.idx.long()] = 0
```
akx, line 267: Dead code?

bitsandbytes/research/autograd/_functions.py
```diff
@@ -336,24 +326,18 @@ def backward(ctx, grad_output):
...
-                    state.CxBt, state.SBt = F.transform(state.CBt, to_order=formatB, transpose=True)
-                # print('back B shape', state.CxBt.shape)
-                # print('back grad shape', C32grad.shape)
-                gradA32, SgradA32 = F.igemmlt(C32grad, state.CxBt, Sgrad, state.SBt)
-                grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape).to(ctx.dtype_A)
-
-            elif state.CB is not None:
+            # if state.CBt is not None:
+            #    gradA32, SgradA32 = F.igemmlt(Cgrad, state.CBt.t())
+            #    grad_A = F.mm_dequant(gradA32, SgradA32, SCgrad, state.SCBt).view(ctx.grad_shape).to(ctx.dtype_A)
```
akx, line 339: Dead code?

csrc/kernels.cu
```diff
@@ -219,7 +220,7 @@ __device__ half dhDequantizeNF4(unsigned char val)
 
 }
 
-__device__ float dDequantizeNF4(unsigned char val)
+__device__ __forceinline__ float dDequantizeNF4(unsigned char val)
```
akx, line 223: I suppose this `__forceinline__` has been measured to be Always A Good Idea?

csrc/kernels.cu
```diff
@@ -627,7 +628,7 @@ __global__ void kQuantizeBlockwise(float * code, T * __restrict__ const A, float
     for(int i = threadIdx.x; i < 256; i+=blockDim.x)
       smem_code[i] = code[i];
 
-  for (unsigned int i = base_idx; i < n_full; i += gridDim.x*BLOCK_SIZE)
+  for (int i = base_idx; i < n_full; i += gridDim.x*BLOCK_SIZE)
```
akx, line 631: Why the type change?

csrc/kernels.cu
```diff
@@ -722,24 +717,25 @@ __global__ void kDequantizeBlockwise(float *code, unsigned char * A, float * abs
   __shared__ typename LoadChar::TempStorage loadchar;
   __shared__ typename StoreT::TempStorage storet;
 
-  for (unsigned int i = base_idx; i < n_load; i += gridDim.x*TILE_SIZE)
+  for (int i = base_idx; i < n_load; i += gridDim.x*TILE_SIZE)
```
akx, line 720: Why the type change?

csrc/kernels.cu
akx: A comment above the less-legible line explaining what's happening should be enough.

csrc/kernels.cu
```diff
@@ -722,24 +717,25 @@ __global__ void kDequantizeBlockwise(float *code, unsigned char * A, float * abs
...
     }
-    local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH)/(blocksize)]);
+    local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH) >> (31 - __clz(blocksize))]);
+    //local_abs_max = __ldg(&absmax[(i+threadIdx.x*NUM_PER_TH)/(blocksize)]);
 
     __syncthreads();
     LoadChar(loadchar).Load(&(A[i]), qvals, valid_items_load, 128);
 
-    switch(DATA_TYPE)
+    switch (DATA_TYPE)
```
akx, line 738: Spurious change?

tests/test_autograd.py
```diff
@@ -249,13 +253,16 @@ def test_matmullt(dim1, dim2, dim3, dim4, funcs, dtype, req_grad, transpose, dec
...
+
+                state.CB, state.SCB, _ = bnb.functional.int8_vectorwise_quant(B2.to(torch.float16))
+
+                # (
+                #     state.CB,
+                #     CBt,
+                #     state.SCB,
+                #     SCBt,
+                #     coo_tensorB,
+                # ) = bnb.functional.double_quant(B2.to(torch.float16))
```
akx, line 265: Dead code?

akx (commented):
Some comments within – I didn't have the time to look at the CUDA code much or the tests this iteration.

Beyond the inline comments: I think the removal of nocublaslt should be split and out and merged in a separate PR.

bitsandbytes/autograd/_functions.py
matthewdouglas: This is a dependency of `torch` so it seems reasonable to me.

bitsandbytes/autograd/_functions.py
matthewdouglas: I'm thinking about it, but right now the existing implementation casts grad outputs and activations to fp16. Would be a separate feature.

csrc/kernels.cu
matthewdouglas: This is a super tiny optimization based on advice in the best practices guide here:

https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#loop-counters-signed-vs-unsigned

Eventually it might be the sort of thing we have linting rules for.

bitsandbytes/functional.py
matthewdouglas: We're planning to take a library-wide sweep through these separately. For now, I'm following the existing patterns, but it is on the backlog.

bitsandbytes/nn/modules.py
matthewdouglas: There's a flake8 rule set available in Ruff (RET) that covers things like this; may separately enable that.

bitsandbytes/nn/modules.py
matthewdouglas: It doesn't make sense as we always keep it row-major now.

csrc/kernels.cu
matthewdouglas: Same as the change above.

csrc/kernels.cu
matthewdouglas: Yeah, I think the IDE must have done that at some point. Soon I do plan to (separately) enforce some style rules.

csrc/kernels.cu
akx: Thanks, the doc link is exactly what I was hoping for, and explains why :)

bitsandbytes/nn/modules.py
akx: Yep, but figured since you were already coalescing the `.to` above, might as well do the obvious next thing now.

csrc/kernels.cu
akx: We could probably just enforce `clang-format` on .cu files?

bitsandbytes/functional.py
matthewdouglas: Thanks, fair point! I don't think we should have exposed this as a public facing API, so I've marked it deprecated and introduced a private `_get_tensor_stream` as the replacement.

bitsandbytes/autograd/_functions.py
matthewdouglas: cc: @Titus-von-Koeller An example here where we held on to activations when it was unnecessary, similar to 4bit.

bitsandbytes/functional.py
matthewdouglas: I've added that, but will also rename this function and have a deprecated igemmlt wrap it, since the signature did change.

bitsandbytes/functional.py
matthewdouglas: Updated!

bitsandbytes/functional.py
matthewdouglas: Planning to deprecate this. The only changes made here were for unit testing which will also have some updates in a follow up PR.

bitsandbytes/nn/modules.py
matthewdouglas: Updated!

csrc/kernels.cu
matthewdouglas: It's not always a good idea, but in this case I saw a small throughput improvement. The same improvement may exist for FP4 as well, but I did not test.

akx (commented):
Couple more comments, mostly regarding documenting the newly public-and-recommended functions.

There's also a bunch of commented-out code that probably should get removed, or it'll haunt the source forever otherwise, I think...

bitsandbytes/functional.py
```diff
@@ -2302,179 +2302,206 @@ def batched_igemm(
...
+    SA: Tuple[torch.Size, str],
+    SB: Tuple[torch.Size, str],
+    out: Optional[torch.Tensor] = None,
+    Sout: Optional[Tuple[torch.Size, str]] = None,
+    dtype=torch.int32,
+):
+    if SA is not None and SA[1] != "row":
+        raise NotImplementedError(f"Only row-major format inputs are supported, but got format `{SA[1]}`")
+    if SB is not None and SB[1] != "row":
+        raise NotImplementedError(f"Only row-major format is supported for matrix B, but got format `{SB[1]}`")
```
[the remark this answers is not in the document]
matthewdouglas: Those arguments were used in the old version, but serve no purpose now. So the main point here is to advise users that happen to try to invoke the old function with col32/ampere/turing layouts that it's not supported anymore. If you invoke with a row-major layout, it's still supported.

SA and SB were required; I'm just being defensive here with checking that they actually are not None.

Titus-von-Koeller:
Hey @matthewdouglas,

Thanks again for the insightful two-hour pairing session – it was great to walk through your code together. I’m impressed by your thoughtful review and the careful attention to detail in this work. There was a lot of complexity to handle pragmatically and I love the incremental refactoring approach that you took. The performance improvements are also really impressive. Great work!

---

Here's my feedback that I collected during our talk:

1. **Organize Test Scripts**  
   Consider moving scripty test parts under `bitsandbytes/scripts/8-bit`. Adding a reference to that in the main implementation would help guide developers to these “eval” scripts for future refactoring.

2. **Clarify absmax Logic**  
   In `get_row_absmax`, please add an explanation about why absmax only over rows is sufficient.

3. **Commentary in MatMul8bitLt**  
   You mentioned needing a comment in `MatMul8bitLt` – could you clarify the specific addition required here?

4. **Documenting Public Functions**  
   Ensure all public functions have clear, detailed docstrings and verify their proper rendering in the documentation.

5. **Deterministic Test Inputs**  
   It makes a lot of sense hard-coded test inputs to improve consistency over the prior approach of using randomization. Please make sure that this is true for all 8-bit related tests before concluding this PR. However, a follow-up PR applying this to other tests would help address ongoing flakiness and would be highly appreciated.

6. **Profiling Code Placement**  
   Please commit your profiling code to the main repo in a reasonable location and/or move more experimental/supplementary code to the `workbench` repo for future team reference.

7. **Benchmark Transparency for Users**  
   Adding benchmark results to the documentation would greatly benefit users, especially in a “deep-dive” section. Please clearly highlight performance comparisons with 16-bit, underscoring benefits with large context and batch sizes, where overhead remains constant. H100 benchmarks could add value but might be low priority. Focus on takeaways from performance, giving users accessible insights from your mental model, so they “know what we know”.

8. **Publicity-Worthy Performance Metrics**  
   Do we have any benchmark metrics from this refactor that might serve as release highlights?

---

Big thanks also to @akx for making time to review this work! We really appreciate your proactive contributions and helpful insights 🤗 Thanks ❤️

csrc/kernels.cu
matthewdouglas: That's something I was thinking about when I first started looking at these too. The fast GEMV kernel `kgemm_4bit_inference_naive` actually does it that way - the codebook for the datatype is loaded from global memory to shared. While the `kDequantizeBlockwise` kernel also has an argument for a code table, it's not used, and instead this hard-coded dequantization function is used instead.

I haven't spent much time profiling this but definitely share that curiosity. With that said, the compiler might be smart enough here to avoid divergence with predication or some other tricks.

bitsandbytes/functional.py
matthewdouglas: I went ahead and removed the lookup to keep this simple; maybe something we can do to further clean up in a separate PR.

bitsandbytes/functional.py
matthewdouglas: Simplified this.

bitsandbytes/functional.py
```diff
@@ -2302,179 +2302,206 @@ def batched_igemm(
...
+    result = int8_linear_matmul(A, B, out=out, dtype=dtype)
+    return result, (result.shape, "row")
+
+
+def int8_linear_matmul(A: torch.Tensor, B: torch.Tensor, out: Optional[torch.Tensor] = None, dtype=torch.int32):
```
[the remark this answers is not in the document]
matthewdouglas: Type checkers should be able to infer this one's return type as `tuple[torch.Tensor, torch.Tensor, torch.Tensor | None]` but I do agree and will be adding docstrings!

bitsandbytes/functional.py
```diff
@@ -2552,96 +2579,73 @@ def coo_zeros(rows, cols, nnz, device, dtype=torch.half):
...
-            None,
-            None,
-            None,
+        # TODO we could improve perf of this
+        outliers = A.abs() >= threshold
+
+        # argwhere needs host/device sync, so we skip when
+        # there aren't actually any outliers.
+        if outliers.any():
+            outlier_cols = torch.argwhere(outliers.any(dim=0)).view(-1)
```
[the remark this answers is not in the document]
matthewdouglas: This would be nice but it would still need a reduction op, e.g. `if dim0_outliers.any()`.

Ideally we could get this part accomplished within the quantization kernel where it's already reading the whole tensor. With that said, right now this new implementation already outperforms the current one, so I'm thinking that will be a future optimization exercise. Will share benchmark data soon!

bitsandbytes/functional.py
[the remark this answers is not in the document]
akx: Maybe it'd be better to be even more defensive: just fail altogether if these arguments are passed, as they won't be used/assigned to/...?

```suggestion
    if SA is not None:
        raise NotImplementedError("SA will not be used; please pass None")
    if SB is not None:
        raise NotImplementedError("SB will not be used; please pass None")
```

or similar..?

bitsandbytes/functional.py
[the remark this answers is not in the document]
akx: Maybe, but the semantics of the 3 returned values would still not be clear from that alone 😄

bitsandbytes/functional.py
matthewdouglas: Added docstring for this.

bitsandbytes/functional.py
```diff
@@ -2302,179 +2302,206 @@ def batched_igemm(
...
 
 
-def mm_dequant(A, quant_state, row_stats, col_stats, out=None, new_row_stats=None, new_col_stats=None, bias=None):
+def int8_mm_dequant(
+    A: torch.Tensor,
+    row_stats: torch.Tensor,
+    col_stats: torch.Tensor,
+    out: Optional[torch.Tensor] = None,
+    bias: Optional[torch.Tensor] = None,
+):
```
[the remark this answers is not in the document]
matthewdouglas: Updated!

bitsandbytes/functional.py
```diff
@@ -2552,96 +2579,73 @@ def coo_zeros(rows, cols, nnz, device, dtype=torch.half):
...
-def double_quant(A, col_stats=None, row_stats=None, out_col=None, out_row=None, threshold=0.0):
-    device = A.device
+def double_quant(
+    A: torch.Tensor,
+    col_stats: Optional[torch.Tensor] = None,
+    row_stats: Optional[torch.Tensor] = None,
+    out_col: Optional[torch.Tensor] = None,
+    out_row: Optional[torch.Tensor] = None,
+    threshold=0.0,
+):
```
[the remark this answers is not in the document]
matthewdouglas: Updated!

bitsandbytes/functional.py
```diff
@@ -2552,96 +2579,73 @@ def coo_zeros(rows, cols, nnz, device, dtype=torch.half):
...
+
+    if out_row is not None:
+        quant_row = out_row.copy_(quant_row)
+    if out_col is not None:
+        quant_col = out_col.copy_(quant_col)
+
+    return quant_row, quant_col, row_stats, col_stats.flatten().float(), outlier_cols
+
+
+def int8_vectorwise_quant(A: torch.Tensor, threshold=0.0):
```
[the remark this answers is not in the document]
matthewdouglas: Updated!

matthewdouglas:
There have now been some documentation updates both for the inline docstrings and the markdown-format public docs.

Additionally, tests related to 8bit now use static shapes. Certain tests related to benchmarking have been extracted away, and others have had a new `deprecated` marker applied where appropriate.

A more detailed look at benchmarking data will be provided with release materials. For now, an overview of inference benchmark results:
* **INT8**
  * On T4 and 4090, the per-token throughput is *improved by 60-85%* and per-token latency is *decreased by 40-45%*.
  * H100 is now supported. With Llama 3.1 70B and batch size >= 8, INT8 is consistently faster than NF4.
* **NF4**: 
  * On T4 and 4090, with batch size of 1, per-token throughput is *improved by 10-25%* and per-token latency is *decreased by 10-20%*.
  * On H100, across all batch sizes, per-token throughput is *improved by up to 28%* and per-token latency is *decreased by up to 22%*.

Titus-von-Koeller:
Really well done @matthewdouglas! What a thorough and well done refactor, which definitely needed some serious skill and dedication: Excellent work and results! 

Imo this is ready to merge now, just the docs deep dive and some small improvements that we discussed in Slack and then 🚀✨

Titus-von-Koeller (approved):

matthewdouglas:
Benchmark details have been added. I've also confirmed that everything is working on V100 without the separate NO_CUBLASLT build needed.

PR #1419, opened 2024-11-19
Fix issue that no valid semantic version tag found when installing from source via personal repo

statelesshz:
For the `multi-backend-refactor` branch, if we git clone the source code from our personal repository and install it, we will get `ValueError: No valid semantic version tag found`. 

This PR fixes this problem.

cc @Titus-von-Koeller @matthewdouglas


Titus-von-Koeller:
Thanks @statelesshz, makes sense. We appreciate your initiative in providing a fix!

PR #1445, opened 2024-12-11
(chore) Remove unused dotfiles

matthewdouglas:
Removing .buckconfig since we do not use [Buck](https://buck.build/).

Removing .yapf since we use [Ruff](url) for formatting.


Titus-von-Koeller (approved):
LGTM, thanks Matthew ❤️

PR #1453, opened 2024-12-16
Check input data device type to decide which op should be used

jiqing-feng:
We should enable the correct backend of ipex op based on the input data device.
Set default semver_tags to 1.0.0 because there are some version checks in transformers.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -481,7 +481,7 @@ class Linear4bit(nn.Linear):
             and not self.training
             and x.requires_grad == False
         ):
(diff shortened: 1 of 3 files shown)


notsyncing:
This PR fixed my error `NotImplementedError: Could not run 'torch_ipex::woq_linear.xpu' with arguments from the 'CPU' backend.` when running a model on cpu with xpu-enabled ipex. Thanks!

jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . Please review this PR, thanks.

matthewdouglas (approved):

PR #1471, opened 2025-01-08
FSDP-QLoRA doc updates for TRL integration

blbadger:
Small documentation changes to maintain compatibility as the `SFTTrainer` class received updates as of TRL v0.13.0:

1. `max_seq_length` and `dataset_text_field` moved to become `SFTConfig` class arguments from the `SFTTrainer`, causing errors when calling the trainer if these arguments are retained.
2. The `tokenizer` argument is being deprecated in favor of the more general `processing_class`, which can be a tokenizer.

The doc currently does not define the `training_arguments` arg, but we can assume that this is a an instance of the `SFTConfig` class in which case `max_seq_length` and `dataset_text_field` would be defined there. I have removed these args from the `SFTTrainer` instantiation because of this, although these arguments could be defined in a config if that is preferable. 

 Swapping the `tokenizer` for `processing_class` deals with the second change. This modification is optional as the `tokenizer` kwarg is deprecated and swapped automatically, but will probably be useful going forward.


Titus-von-Koeller (approved):
Thanks @blbadger for you first PR and this proactive attitude towards fixing things to the benefit of the community ❤️ Really appreciated!

blbadger:
Happy to contribute @Titus-von-Koeller !

PR #1484, opened 2025-01-23
Enable dequant+matmul 8bit path for Intel CPU and XPU

jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . This feature enables dequant 8bit weight and using float matmul. It speed-up the lora finetune for 3x on XPU and 2x on CPU by the [lora finetune script](https://github.com/huggingface/peft/blob/main/examples/olora_finetuning/olora_finetuning.py) on llama3-8b by the command `python olora_finetuning.py --base_model alokabhishek/Meta-Llama-3-8B-Instruct-bnb-8bit --init_lora_weights gaussian --seed 42 --torch_dtype bfloat16 --device_map cpu`. 

All tests in transformers have been passed, please review this PR. Thanks!

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -579,7 +579,8 @@ class MatMul8bitFp(torch.autograd.Function):
     @staticmethod
     def backward(ctx, grad_output):
         state = ctx.state
-        CB = state.CB.to(ctx.dtype_A).mul_(state.SCB.unsqueeze(1).mul(1.0 / 127.0))
+        B = state.CxB if state.CxB is not None else state.CB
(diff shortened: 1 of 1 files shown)


matthewdouglas:
On the mainline branch with the int8 refactoring that was done for v0.45.0, `Linear8bitLt` was simplified a bit. One of the optimizations made there was to only do the row-wise quantization for inference, and still do the "double quant" with row/col for training.

 But with that said the decomposition into separate int8 and fp16 matmuls has some overhead and I assume this is where the unsafe operations come from, particularly when threshold!=0.

Happy to merge; we can revisit doing int8 computation in the future if needed.

matthewdouglas (approved):

PR #1489, opened 2025-01-24
add xpu device index for `Int8Params`

faaany:
The following code will not move tensor to the desired `target_device`:

```bash
import torch
import bitsandbytes as bnb

random_tensor = torch.randn((3, 3), dtype=torch.float16)
target_device = 2
new_value = bnb.nn.Int8Params(random_tensor, requires_grad=False).to(target_device)
print(new_value)
# Parameter containing:
# Parameter(Int8Params([[ 127,  -97, -126],
#             [  60,   11,  127],
#             [-127,   90,  -35]], device='xpu:0', dtype=torch.int8))
```

To fix this, we need to pass the device index to the `xpu` method just like in the `cuda` method.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -660,9 +660,9 @@ class Int8Params(torch.nn.Parameter):
         self.SCB = SCB
         return self
 
-    def xpu(self):
+    def xpu(self, device):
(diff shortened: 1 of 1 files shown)


yao-matrix:
@jiqing-feng

matthewdouglas (approved):
LGTM, thanks!

PR #1413, opened 2024-11-12
Remove triton.ops, copy necessary bits here

bertmaher:
Summary: Triton upstream removed `triton.ops` and moved it to a semi-unmaintained `kernels` repo.  Since all that's needed here is the perf model, just add those bits here.

PR #1493, opened 2025-01-26
bitsandbytes, test existence of triton.ops.matmul_perf_model

lamikr:
This fixes the bitsandbytes to work with newer triton versions which does not anymore include the triton.ops.matmul_perf_model which has been moved to triton-langs kernels project. https://github.com/triton-lang/kernels

Fix is to simply catch the import error if triton is available but the triton.ops.matmul_perf_model is not. In such case the logic is implemented in a same way than earlier when the triton was not available at all.

To my understanding there is currently no a way to install opt_matmul_perf.py from new tritons kernel project directly with pip, so in future it could propably be good to include that python class directly in bitsandbytes.

https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1492

### DIFF
--- a/bitsandbytes/triton/int8_matmul_mixed_dequantize.py
+++ b/bitsandbytes/triton/int8_matmul_mixed_dequantize.py
@@ -2,7 +2,16 @@ import torch
 
 from bitsandbytes.triton.triton_utils import is_triton_available
 
-if not is_triton_available():
+use_triton = False
+if is_triton_available():
+    import triton
+    try:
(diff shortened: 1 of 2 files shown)


matthewdouglas:
Hi @lamikr, please see #1413 which was merged and released in [v0.45.1](https://github.com/bitsandbytes-foundation/bitsandbytes/releases/tag/0.45.1).

Titus-von-Koeller:
closing this, agree with @matthewdouglas's assertion. Fixed in #1413 and deprecating triton-based modules in the next versions (already marked for deprecation).

Thanks for the initiative @lamikr.

PR #1497, opened 2025-01-28
Improve guard for triton compatibility

matthewdouglas:
For the Triton functionality we need to guard beyond simply checking if the `triton` package is installed. This updates the guard to verify that there is also a supported triton backend.

@winglian @Titus-von-Koeller


Titus-von-Koeller (approved):
LGTM, thanks for taking the lead on that!

PR #1503, opened 2025-02-08
Remove error log for Intel CPU/XPU

jiqing-feng:
Intel CPU/XPU do not need compile any C++ file, so we can remove the error log.


bitsandbytes/cextension.py
```diff
@@ -70,6 +70,11 @@ def __init__(self, lib: ct.CDLL):
 
 
 def get_native_library() -> BNBNativeLibrary:
+    from bitsandbytes.backends.cpu_xpu_common import ipex_cpu, ipex_xpu
+
+    if ipex_cpu or ipex_xpu:
+        return None
```
akx, line 76: The type annotation for this function does not indicate it could ever return None.

jiqing-feng: done

akx: What was done?

jiqing-feng: Sorry for I forgot to remove `return None` in this function. Now, it's okay.

jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . We are seeking to remove this error log, I know BNB is under refactoring, but this is emergent. 

Many Intel customers reported this issue. Do you mind take a look? Thanks!

PR #1506, opened 2025-02-10
fix xpu woq linear dtype

jiqing-feng:
In XPU env, we usually cannot get ipex_cpu, so just set the dtype to 2 which indicate we use bf16 dtype to compute.

Hi @Titus-von-Koeller @matthewdouglas . Please review this PR. Thanks!


bitsandbytes/backends/cpu_xpu_common.py
```diff
@@ -562,7 +562,7 @@ def gemm_4bit_impl(
             None,
             None,
             state.blocksize,
-            ipex_cpu.quantization.WoqLowpMode.BF16,
+            2,
```
akx, line 565: Feels like there should be a comment about the magic number "2".

jiqing-feng: fixed

PR #1534, opened 2025-02-20
Add npu support for LLM.int8 forward

SlightwindSec:
**What does this PR do?**  
1. **LLM.int8 inference (forward-only)**  
   - Adds PyTorch-based `int8_vectorwise_dequant` (AscendC version WIP).  
   - Uses fused `npu_quant_matmul` for NPU-optimized matmul+dequant.  

2. **NF4 memory fix**  
   - Implements chunk-based processing to reduce memory usage and prevent OOM for large tensors.

**Notes**  
- Backward pass for LLM.int8 requires future AscendC kernels.  

**Collaborators**  
@ji-huazhong @Ginray @MatrixPlayer  

cc @Titus-von-Koeller @matthewdouglas

### DIFF
--- a/bitsandbytes/backends/npu.py
+++ b/bitsandbytes/backends/npu.py
@@ -11,6 +11,7 @@ except ImportError:
 
 from bitsandbytes.cextension import lib
 from bitsandbytes.functional import (
+    COOSparseTensor,
     get_4bit_type,
     get_ptr,
 )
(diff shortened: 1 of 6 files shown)


matthewdouglas:
Thanks! I'll provide a few comments. I also want to remind that we're quite close on moving to torch.library and custom ops for device dispatching and will provide more info soon on porting over to that. It should be a fairly simple process!

bitsandbytes/backends/npu.py
```diff
@@ -69,7 +181,7 @@ def int8_linear_matmul(
         out: Optional[torch.Tensor] = None,
         dtype=torch.int32,
     ) -> torch.Tensor:
-        raise NotImplementedError
+        return Int8AB(A, B)
```
matthewdouglas, line 184: Interesting and clever! While this does break the expected API as this isn't returning a Tensor (or performing any operations really), I can completely understand _why_ it is done this way. I think this will be OK for right now and we'll make the interface better in this regard later on.

bitsandbytes/nn/modules.py
```diff
@@ -320,9 +321,6 @@ def cpu(self, non_blocking: bool = False):
         return self.to(device="cpu", non_blocking=non_blocking)
 
     def npu(self, device: Optional[Union[int, device, str]] = None, non_blocking: bool = False):
-        # `torch.Tensor.to(<int num>)` is not supported by `torch_npu` (see this [issue](https://github.com/Ascend/pytorch/issues/16)).
-        if isinstance(device, int):
-            device = f"npu:{device}"
```
matthewdouglas, line 325: Does this require a bump in the minimum torch_npu to support now?

Titus-von-Koeller:
@SlightwindSec is this / NPU support something that's actively being worked on? Is this with corporate backing or a private endeavor?

PR #1547, opened 2025-02-28
Enable XPU int matmul

jiqing-feng:
XPU int matmul is available in torch2.6, it is faster than float matmul and also requires less memory.


matthewdouglas:
LGTM! We'll hold for the moment to wait on IPEX v2.6.0+xpu release. Thanks!

PR #1549, opened 2025-03-01
Fix CPU dequantization to use nested dequantized scaling constant

zyklotomic:
I believe that in the CPU case, we have forgotten to use the updated absmax if that absmax too was quantized, like in QLoRA's `doubleDequant()`.

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -1049,7 +1049,7 @@ def dequantize_blockwise(
         lib.cdequantize_blockwise_cpu_fp32(
             get_ptr(code),
             get_ptr(A),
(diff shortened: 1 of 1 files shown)


zyklotomic:
Similarly, I also noticed slightly above, in: https://github.com/bitsandbytes-foundation/bitsandbytes/blob/753e411fe60f0d059a02d742985bd73692593841/bitsandbytes/functional.py#L1020 the updated `code` variable is not used. Not sure if this is an error or extraneous.

matthewdouglas:
Thanks! The issue above seems like an oversight too, but it's refactored and addressed in #1544.

PR #1562, opened 2025-03-11
update comments

jiqing-feng:
update dequantize comment as both nf4 and fp4 are supported.


jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . Please review this PR. Thanks!

PR #1565, opened 2025-03-14
XPU backend support 8bit optimizer

Liangliang-Ma:
This pr adds support of 8bit optimizer for XPU backend.
The backend kernels is integrated in Intel_extension_for_pytorch now.
We have verified the whole path accuracy with 8bit Adam blockwise. 

Also add device synchronize func for every backend class to avoid cuda hardcode.

@jiqing-feng  @matthewdouglas @Titus-von-Koeller

### DIFF
--- a/bitsandbytes/backends/xpu.py
+++ b/bitsandbytes/backends/xpu.py
@@ -13,7 +13,10 @@ from .cpu_xpu_common import (
     int8_mm_dequant_impl,
     quantize_4bit_impl,
 )
-import intel_extension_for_pytorch as ipex
+try:
+    import intel_extension_for_pytorch as ipex
+except BaseException:
+    ipex_xpu = None
 
 Tensor = torch.Tensor
 
@@ -260,7 +263,6 @@ class XPUBackend(Backend):
             raise ValueError(
                 f"Gradient+optimizer bit data type combination not supported: grad {g.dtype}, optimizer {state1.dtype}",
             )
-
         optim_func(
             p,
             g,
--- a/bitsandbytes/backends/xpu.py
+++ b/bitsandbytes/backends/xpu.py
@@ -205,6 +205,9 @@ class XPUBackend(Backend):
         blocksize: int = 4096,
         nested=False,
     ) -> torch.Tensor:
+        if ipex_xpu is None:
+            raise RuntimeError("Please install intel_extension_for_ipex for 8bit optimizer backend on XPU device.")
+
         # void cdequantize_blockwise_fp32(float *code, unsigned char *A, float *absmax, float *out, int blocksize, const int n, cudaStream_t stream)
         if out.dtype == torch.float16:
             ipex.xpu.bitsandbytes.cdequantize_blockwise_fp16(code, A, absmax, out, blocksize, A.numel())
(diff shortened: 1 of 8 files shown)


jiqing-feng:
After I verified on ipex 2.7, we can add XPU tests on [test_optim](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/multi-backend-refactor/tests/test_optim.py).

matthewdouglas:
Thanks! 

Optimizer support isn't addressed yet on the new custom ops interface that we've mainlined, but we can keep dev on it here in this branch until that's ready.

Is there a plan to support any other optimizers? Completely understandable if not;  just curious!

bitsandbytes/backends/xpu.py
```diff
@@ -185,7 +205,19 @@ def dequantize_blockwise(
...
+
+        # void cdequantize_blockwise_fp32(float *code, unsigned char *A, float *absmax, float *out, int blocksize, const int n, cudaStream_t stream)
+        if out.dtype == torch.float16:
+            ipex.xpu.bitsandbytes.cdequantize_blockwise_fp16(code, A, absmax, out, blocksize, A.numel())
+        elif out.dtype == torch.bfloat16:
+            ipex.xpu.bitsandbytes.cdequantize_blockwise_bf16(code, A, absmax, out, blocksize, A.numel())
+        elif out.dtype == torch.float32:
+            ipex.xpu.bitsandbytes.cdequantize_blockwise_fp32(code, A, absmax, out, blocksize, A.numel())
+        else:
+            raise ValueError(f"Blockwise quantization only supports 16/32-bit floats, but got {out.dtype}")
```
matthewdouglas, line 219: This will be useful when porting over to the new custom ops as an implementation for `"bitsandbytes::dequantize_blockwise.out(Tensor A, Tensor absmax, Tensor code, int blocksize, ScalarType dtype, Tensor! out) -> ()`

jiqing-feng:
> Thanks!
> 
> Optimizer support isn't addressed yet on the new custom ops interface that we've mainlined, but we can keep dev on it here in this branch until that's ready.
> 
> Is there a plan to support any other optimizers? Completely understandable if not; just curious!

Currently no plan to enable other optimizers.

bitsandbytes/backends/xpu.py
jiqing-feng: Hard to understand. Could you please supply more details or instructions? Thanks!

matthewdouglas: @jinq-feng
What I meant by that is in the new interface, we define a custom op for the 8bit dynamic quantization that is used for the optimizers and nested absmax. Since there seems to exist an optimized implementation of this exact op in `ipex.xpu` now, we can just wrap it during our port.

Titus-von-Koeller:
@Liangliang-Ma I invited you to our bitsandbytes-intel slack channel. Could you join there to discuss if you're planning on supporting the PagedOptimizers of BNB?

The paged memory feature is what we have in functional.py:get_paged() using [cudaMallocManaged](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gd228014f19cc0975ebe3e0dd2af6dd1b) under the hood.

Liangliang-Ma:
@Titus-von-Koeller Due to changes in work content, I will not be doing related work in the near future. Thanks for invitation tho :)

PR #1567, opened 2025-03-17
Fix XPU 4bit

jiqing-feng:
It fix

Dequant value error: The default value is "fp4" and we checked the default value 1st so it is always "fp4". Now we fixed it by enable quant_state 1st.
Use ipex dequant_4bit kernel in XPU dequant 4bit. It can bring 1.5x speed-up on QLora.
We have at least 2x speed-up for the dequant_4bit kernel.


jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . This PR is ready to be merged, please review it. Thanks!

PR #1574, opened 2025-03-26
Drop Python 3.8 support.

matthewdouglas:
Python 3.8 reached EOL in October 2024. As such, it is appropriate to drop support and set the new minimum to Python 3.9.

This PR removes Python 3.8 support, updates the Ruff linter, and targets Python 3.9 as the minimum version.


Titus-von-Koeller (approved):
Looking great, thanks!

PR #1586, opened 2025-04-09
Specify blocksize

cyr0930:
Specify blocksize as it is a bit confusing just showing number in the code.
Also matching the code of same logic.


matthewdouglas:
Seems reasonable to me. Thanks!

PR #1594, opened 2025-04-15
Support LLM.int8() inference with torch.compile

matthewdouglas:
This PR introduces support for torch.compile with LLM.int8() inference. It removes graph breaks that would otherwise occur during the problem decomposition by introducing a new wrapper custom operator `bitsandbytes::int8_mixed_scaled_mm`.

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -210,36 +210,27 @@ class MatMul8bitLt(torch.autograd.Function):
                 # 2. Quantize B
                 state.CB, state.SCB, _ = F.int8_vectorwise_quant(B.to(torch.float16))
 
-        # Handle sparse decomposition. In some instances, we may have not found any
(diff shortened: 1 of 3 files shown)


bitsandbytes/autograd/_functions.py
```diff
@@ -210,37 +210,28 @@ def forward(
...
-        # Handle sparse decomposition. In some instances, we may have not found any
-        # outlier columns at all. In that case, we'll skip this part completely.
-        if state.threshold > 0.0 and outlier_cols is not None and outlier_cols.numel():
+        # Handle sparse decomposition
+        if state.threshold > 0.0:
             state.idx = outlier_cols
 
-            # Zero out the outliers in the transposed 8bit inputs.
-            if CAt is not None:
-                CAt[:, state.idx] = 0
```
matthewdouglas, line 220: This step is missing right now, but note that we''re planning to deprecate/remove support for full int8 training.

Issue #1513, opened 2025-02-13
torch.version.cuda.split error


Lier0:
### System Info

python3.12
rocm6.3
amdgpu : gfx1030

### Reproduction

```
$ python -m bitsandbytes

Could not load bitsandbytes native library: 'NoneType' object has no attribute 'split'
Traceback (most recent call last):
  File "/home/lier0/ia/ComfyUI/comfyui_env/lib64/python3.12/site-packages/bitsandbytes/cextension.py", line 85, in <module>
    lib = get_native_library()
          ^^^^^^^^^^^^^^^^^^^^
  File "/home/lier0/ia/ComfyUI/comfyui_env/lib64/python3.12/site-packages/bitsandbytes/cextension.py", line 64, in get_native_library
    cuda_specs = get_cuda_specs()
                 ^^^^^^^^^^^^^^^^
  File "/home/lier0/ia/ComfyUI/comfyui_env/lib64/python3.12/site-packages/bitsandbytes/cuda_specs.py", line 39, in get_cuda_specs
    cuda_version_string=(get_cuda_version_string()),
                         ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lier0/ia/ComfyUI/comfyui_env/lib64/python3.12/site-packages/bitsandbytes/cuda_specs.py", line 29, in get_cuda_version_string
    major, minor = get_cuda_version_tuple()
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lier0/ia/ComfyUI/comfyui_env/lib64/python3.12/site-packages/bitsandbytes/cuda_specs.py", line 24, in get_cuda_version_tuple
    major, minor = map(int, torch.version.cuda.split("."))
                            ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'

CUDA Setup failed despite CUDA being available. Please run the following command to get more information:

python -m bitsandbytes

Inspect the output of the command and see if you can locate CUDA libraries. You might need to add them
to your LD_LIBRARY_PATH. If you suspect a bug, please take the information from python -m bitsandbytes
and open an issue at: https://github.com/bitsandbytes-foundation/bitsandbytes/issues

++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
++++++++++++++++++ BUG REPORT INFORMATION ++++++++++++++++++
[...]

PR #1599, opened 2025-04-16
fix: Improve CUDA version detection and error handling

ved1beta:
fix: Improve CUDA/HIP version detection and error handling

This commit fixes issue #1513 by improving the version detection logic
in cuda_specs.py to handle both CUDA and ROCm/HIP systems more robustly.

Key changes:
- Add proper error handling for None values in version detection
- Make version string and tuple functions return Optional types
- Add robust version string parsing with try-except blocks
- Add validation checks for compute capabilities
- Improve error handling in get_cuda_specs function
- Add proper type hints and documentation

The changes ensure that:
1. NoneType errors are properly handled when version info is missing
2. ROCm/HIP systems are properly supported
3. Invalid version strings don't cause crashes
4. All error cases return None instead of raising exceptions

Test results show successful version detection for both CUDA and ROCm/HIP
systems, resolving the original issue where torch.version.cuda.split()
would fail on ROCm systems.

### DIFF
--- a/bitsandbytes/cuda_specs.py
+++ b/bitsandbytes/cuda_specs.py
@@ -1,6 +1,6 @@
 import dataclasses
 from functools import lru_cache
-from typing import Optional
+from typing import Optional, Tuple
 
 import torch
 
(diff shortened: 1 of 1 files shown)


matthewdouglas:
LGTM, thanks! We can merge after fixing the failing linter check.

PR #1604, opened 2025-04-21
fix log for ipex

jiqing-feng:
The previous lof info is not clear, we only indicate user didn't have library. But user do not need to build library, they can install ipex if they use Intel CPU or XPU.


jiqing-feng:
Hi @Titus-von-Koeller @matthewdouglas . Please review this PR.

Issue #1467, opened 2025-01-03
AttributeError: 'NoneType' object has no attribute 'cquantize_blockwise_fp16_fp4'


hessaAlawwad:
### System Info

working on google colab.
torch is installed using:
torch==1.13.0
torchvision==0.14.0
torchaudio==0.13.0
pytorchvideo @ git+https://github.com/facebookresearch/pytorchvideo.git@28fe037d212663c6a24f373b94cc5d478c8c1a1d

### Reproduction

I am trying to load a model using 4bit quantization while having another model in the pipline that requires specific version of torch(1.13.0). I try to load it using:

```
model_name = "Hessa/MMTQA-merged2"

# Load the model in 4-bit precision
model = AutoModelForCausalLM.from_pretrained(
    model_name,
    load_in_4bit=True,  # Enable 4-bit quantization
    device_map="auto",  # Automatically place the model on available devices (GPU/CPU)
    torch_dtype=torch.float16,  # Optionally use fp16 if needed
)
```

but I got the error:
```
ERROR:bitsandbytes.cextension:Could not load bitsandbytes native library: libcusparse.so.11: cannot open shared object file: No such file or directory
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/bitsandbytes/cextension.py", line 85, in <module>
    lib = get_native_library()
  File "/usr/local/lib/python3.10/dist-packages/bitsandbytes/cextension.py", line 72, in get_native_library
    dll = ct.cdll.LoadLibrary(str(binary_path))
  File "/usr/lib/python3.10/ctypes/__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "/usr/lib/python3.10/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libcusparse.so.11: cannot open shared object file: No such file or directory
WARNING:bitsandbytes.cextension:
CUDA Setup failed despite CUDA being available. Please run the following command to get more information:

python -m bitsandbytes

Inspect the output of the command and see if you can locate CUDA libraries. You might need to add them
to your LD_LIBRARY_PATH. If you suspect a bug, please take the information from python -m bitsandbytes
and open an issue at: https://github.com/bitsandbytes-foundation/bitsandbytes/issues

config.json: 100%
 5.27k/5.27k [00:00<00:00, 416kB/s]
[...]

PR #1607, opened 2025-04-24
Fix AttributeError in 4-bit quantization when CUDA library is not loaded

demoncoder-crypto:
Addresses #1467

### DIFF
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -958,7 +958,34 @@ def quantize_fp4(
     compress_statistics=False,
     quant_storage=torch.uint8,
 ):
-    return quantize_4bit(A, absmax, out, blocksize, compress_statistics, "fp4", quant_storage)
+    """Quantize tensor A in blocks to fp4 values.
+
+    Quantizes tensor A by dividing it into blocks which are independently quantized.
+
+    Args:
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller:
Hey @demoncoder-crypto thanks for the contribution! However, we're planning to handle this differently with a fallback mechanism, see this PR: https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1615

demoncoder-crypto:
Hi @Titus-von-Koeller, I just wanted to know if there are any issues that I can contribute too, any pointers are appreciated.

Titus-von-Koeller:
@demoncoder-crypto Cool that you're interested to contribute! It's a bit tough to figure out sth to do at the moment because we have so much going on, but most things that I'm aware of require a lot of background knowledge..

For example this PR has become stale somehow and would be important to make progress again and is somewhat self contained.

demoncoder-crypto:
@Titus-von-Koeller Thanks for replying it really means a lot. I just want to contribute as I am learning about this right now, I think I do have some decent knowledge on this subject if something requires learning I will definitely put in the work to learn it. I also do understand you are also quite busy so I really do appreciate the time you have given me thanks.

PR #1612, opened 2025-04-27
FSDP2 integration: torch.chunks(Params4bit)

### LINKED ISSUE #1424: torch.chunks(Params4bit) not returning Params4bit subclass
### System Info

Hi, I'am trying to make FSDP2 work with a llama model quantized with bitsandbytes but it seems that bitsandbytes' tensor subclasses like Params4bit are not compatible with the way FSDP2 shards the model.
When creating the DTensors to shard the model FSDP2 applies torch.chunk to the parameters which get returned by torch.chunk as ordinary Tensors instead of the original subclass (like Params4bit) which leads to errors down the line.

Is this a known issue and are there plans to make bitsandbytes composable with FSDP2?

### Reproduction

Created a simple repro:
```python
import torch
import torch.nn as nn

import bitsandbytes as bnb
from bitsandbytes.nn import Params4bit
        
blocksize=64
compress_statistics = True
quant_type = "fp4"
quant_storage=torch.uint8

w = torch.ones(4).to("cuda")

w_4bit, quant_state = bnb.functional.quantize_4bit(
    w,
    blocksize=blocksize,
    compress_statistics=compress_statistics,
    quant_type=quant_type,
    quant_storage=quant_storage,
    )

b = Params4bit.from_prequantized(w_4bit, quant_state.as_dict(packed=True))
print(f"{b=}")

chunks = torch.chunk(b, 2, dim=0)

print(f"{chunks=}")
```

Output:
```
b=Parameter containing:
Parameter(Params4bit([[51],
            [51]], device='cuda:0', dtype=torch.uint8))
chunks=(tensor([[51]], device='cuda:0', dtype=torch.uint8), tensor([[51]], device='cuda:0', dtype=torch.uint8))
```

### Expected behavior

Expecting the output of torch.chunk to be a a tuple of Params4bits instead of a Tensors.

ved1beta:
Fix Params4bit tensor subclass handling

- Replace torch._C._dispatch_without_tensor_functionalization() with torch._C._DisableTorchDispatch()
- Add custom __repr__ method to prevent .tolist() errors
- Implement chunk/cat operations that preserve Params4bit tensor type
- Ensure proper tensor subclass behavior in PyTorch 2.6.0
Fix #1424

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -234,12 +234,48 @@ class Params4bit(torch.Tensor):
         if kwargs is None:
             kwargs = {}
 
-        # Avoid recursion by using torch's dispatch mechanism
+        # Get the function name for specific handling
+        func_name = func.__name__
+
+        # Handle specific functions that need special treatment
+        if func_name in ["chunk", "cat"]:
+            # Use DisableTorchDispatch to avoid infinite recursion
+            with torch._C._DisableTorchDispatch():
+                # Get the output from the original function
+                out = func(*args, **kwargs)
+
+            # For chunk, we need to convert each chunk to a Params4bit
+            if func_name == "chunk" and isinstance(args[0], Params4bit):
+                param4bit = args[0]
+                return tuple(
+                    Params4bit(
+                        chunk,
+                        quant_state=param4bit.quant_state,
+                        requires_grad=chunk.requires_grad
+                    )
+                    for chunk in out
+                )
+
+            # For cat, we check if any tensor in the input sequence is a Params4bit
+            elif func_name == "cat":
+                # The first argument to cat is the sequence of tensors
+                tensors = args[0]
+                params4bit_tensors = [t for t in tensors if isinstance(t, Params4bit)]
+                if params4bit_tensors:
+                    # Use the quant_state from the first Params4bit tensor
+                    param4bit = params4bit_tensors[0]
+                    return Params4bit(
+                        out,
+                        quant_state=param4bit.quant_state,
+                        requires_grad=out.requires_grad
+                    )
+
+        # Use DisableTorchDispatch for general case
         with torch._C._DisableTorchDispatch():
             # Get the output from the original function
             out = func(*args, **kwargs)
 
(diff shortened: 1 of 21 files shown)


ved1beta:
can i get a feed back on this :)

Titus-von-Koeller:
Hey @ved1beta,

Thanks for the contribution. Could you tell us a bit more on why you deleted the lines that you deleted and how you validated your implementation?

ved1beta:
When changing from torch.nn.Parameter to torch.Tensor, we're now using the parent class's initialization methods which already handle basic tensor properties.

bitsandbytes/nn/modules.py
```diff
@@ -1063,3 +1096,44 @@ def forward(self, x):
...
+        quant_state = input.quant_state
+        # Create a new list of chunks that are Params4bit tensors
+        return tuple(Params4bit(chunk, quant_state=quant_state) for chunk in chunks)
+
+    # Otherwise, return the original chunks
+    return chunks
+
+# Replace the original torch.chunk with our custom implementation
+original_chunk = torch.chunk
+torch.chunk = _chunk_preserving_params4bit
```
matthewdouglas, line 1118: I'm a little bit hesitant to do patching like this. If we can avoid that and have this work with e.g. just `__torch_function__` and/or and `__torch_dispatch__` that'd be more ideal.

matthewdouglas (changes requested):

Titus-von-Koeller:
Really cool, ultimately a high-level test making sure we also notice when this functionality breaks would be really cool. For now to reproduce your exact results it would already be helpful to include an example script in `/examples`.

bitsandbytes/nn/modules.py
ved1beta: tried patching dispatch func itself please have a look : )

Titus-von-Koeller:
@ved1beta Just wondering why you closed this?

ved1beta:
I was stuck, examples and tests were failing can you help me with it willing to make it work

djsaunde:
I'd like to have this for Axolotl, I might jump in here and help test @ved1beta

PR #1616, opened 2025-04-28
Support 4bit torch.compile fullgraph with PyTorch nightly

matthewdouglas:
This PR enables use of `torch.compile` without graph breaks for `Linear4bit` modules. It requires nightly PyTorch (2.8+).


matthewdouglas:
@StrongerXi Thanks for the tip! It does work without that `__torch_function__` impl, so I'll remove it!

PR #1626, opened 2025-05-07
Moved int8_mm_dequant from CPU to default backend

Egor-Krivov:
Moved `bitsandbytes::int8_mm_dequant` implementation from CPU to default backend to enable `pytest tests/test_functional.py::TestLLMInt8Functional` on both XPU and CPU. Since it's in default backend, we no longer need it in CPU backend.


matthewdouglas (approved):
LGTM, thank you!

PR #1628, opened 2025-05-08
Enable CPU/XPU native and ipex path

jiqing-feng:
This PR enables ipex and other optimizations including:
1. ipex fused op
2. enable fp4 on cpu
3. enable has_rem on quantize/dequantize 4bit
4. Simple 8bit matmul so can make finetune faster on CPU

Also, it fixed the parameter patch for cpu.

It could pass all [transformers tests](https://github.com/huggingface/transformers/tree/main/tests/quantization/bnb)

After this PR merged, I will update the installation guide.

@matthewdouglas @Titus-von-Koeller


bitsandbytes/functional.py
```diff
@@ -1123,6 +1123,15 @@ def dequantize_4bit(
         if absmax.dtype != torch.float32:
             absmax = absmax.float()
 
+    # IPEX format is different, we need extra process.
+    if getattr(quant_state, "ipex", False) and quant_state.quant_type == "nf4":
+        if A.device.type == "xpu":
+            out = torch.ops.torch_ipex.dequantize_4bit(A, "nf4", quant_state.shape, absmax, None, blocksize).t()
+            return out
+        elif A.device.type == "cpu":
```
Devjiu, line 1131: I think mixing cpu and xpu code is not a good idea, let's just make separate backends for that, what do you think?

bitsandbytes/utils.py
```diff
@@ -38,6 +38,16 @@ def outlier_hook(module, input):
             hook.remove()
 
 
+# convert btw standard 4-bit compression format and ipex compression format
+def reverse_4bit_compress_format(weight):
+    out_1 = torch.empty(weight.size(0), dtype=torch.int32, device=weight.device)
+    out_2 = torch.empty(weight.size(0), dtype=torch.int32, device=weight.device)
```
matthewdouglas, line 44: These two lines seem superfluous; they're reassigned right below.

bitsandbytes/utils.py
```diff
@@ -38,6 +38,16 @@ def outlier_hook(module, input):
             hook.remove()
 
 
+# convert btw standard 4-bit compression format and ipex compression format
+def reverse_4bit_compress_format(weight):
```
matthewdouglas, line 42: I think we should mark this as a private/internal API.

```suggestion
def _reverse_4bit_compress_format(weight: torch.Tensor):
```

bitsandbytes/backends/cpu/ops.py
```diff
@@ -133,32 +164,47 @@ def _(
...
+    # code is fp32, cast to dtype to avoid the mismatch issue
+    code = CODE[quant_type].to(dtype)
+    out_dq = code[out_dq]
+
+    # Apply scales
+    if out_dq.numel() != n:
+        assert out_dq.numel() == n + 1
+        out_dq = torch.narrow(out_dq, 0, 0, n)
+    blocks = n // blocksize
+    blocks += 1 if n % blocksize > 0 else 0
```
matthewdouglas, line 194: Same thought as with the quantize op.

```suggestion
    blocks = -(n // -blocksize)
```

bitsandbytes/functional.py
```diff
@@ -2508,3 +2536,49 @@ def vectorwise_mm_dequant(xq, S1, S2, dtype=torch.half, quant_type="vector"):
         return x.to(dtype)
     else:
         return None
+
+
+def enable_ipex_fusion(linear, x):
```
matthewdouglas, line 2541: This should probably not be a public API.

```suggestion
def _enable_ipex_fusion(linear: torch.nn.Module, x: torch.Tensor):
```

bitsandbytes/nn/modules.py
```diff
@@ -492,8 +521,9 @@ def forward(self, x: torch.Tensor):
             x = x.to(self.compute_dtype)
 
         bias = None if self.bias is None else self.bias.to(self.compute_dtype)
+        weight = self.weight.t() if len(self.weight.shape) == 2 else self.weight
```
matthewdouglas, line 524: It's not clear to me when `len(self.weight.shape) != 2`.

matthewdouglas (changes requested):

bitsandbytes/autograd/_functions.py
```diff
@@ -378,6 +403,16 @@ def matmul_4bit(
...
+    if A.device.type in ("cpu", "xpu") and A.requires_grad == False:
+        if getattr(quant_state, "ipex", False):
+            B = B.t() if B.dim() == 2 else B
+            out = F.gemv_4bit(A, B, out, state=quant_state)
+            if bias is not None:
+                out += bias
+            return out
+        else:
+            return MatMul4Bit.apply(A, B, out, bias, quant_state)
+
```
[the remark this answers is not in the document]
jiqing-feng: 1. A is the input that won't be changed
2. When we enable ipex on CPU, the dim of weight will be 4.

bitsandbytes/autograd/_functions.py
```diff
@@ -298,6 +298,29 @@ def backward(ctx: torch.autograd.function.FunctionCtx, grad_output: torch.Tensor
         return grad_A, grad_B, None, grad_bias, None
 
 
+class MatMul8bitFp(torch.autograd.Function):
+    # For Intel CPU and XPU, the double quant has many unsafe operations which will breaks the finetune.
+    # We'd like to use dequant + matmul to run finetune currently.
```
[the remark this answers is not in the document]
jiqing-feng: The MatMul8bitLt is much slower than MatMul8bitFp in finetune. I have updated the reason in comment.

bitsandbytes/functional.py
jiqing-feng: fixed.

bitsandbytes/autograd/_functions.py
```diff
@@ -298,6 +298,29 @@ def backward(ctx: torch.autograd.function.FunctionCtx, grad_output: torch.Tensor
...
+        output = torch.matmul(A, CB).to(A.dtype)
+        ctx.state = state
+        ctx.dtype_A = A.dtype
+        ctx.grad_shape = A.shape
+        return output
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        state = ctx.state
+        B = state.CxB if state.CxB is not None else state.CB
```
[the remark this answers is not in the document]
jiqing-feng: fixed

bitsandbytes/nn/modules.py
jiqing-feng: already in the comment

bitsandbytes/utils.py
jiqing-feng: fixed

bitsandbytes/backends/cpu/ops.py
```diff
@@ -88,31 +88,62 @@ def _(A: torch.Tensor, absmax: torch.Tensor, code: torch.Tensor, blocksize: int,
...
-    # TODO: Support when weight matrix is not divisible by blocksize
-    torch._check(n % blocksize == 0, lambda: f"n must be divisible by blocksize, got {n} and {blocksize}")
-
-    # Divide into blocks and normalize
-    blocks = A.reshape(-1, blocksize)
-    absmax = blocks.abs().max(dim=1).values.float()
-    scaled = blocks / absmax.unsqueeze(-1)
-
+    blocks = n // blocksize
+    blocks += 1 if n % blocksize > 0 else 0
```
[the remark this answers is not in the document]
jiqing-feng: Actually, I didn't see any complicated logic, I just split the tensor into 2 parts to process. The feature is reported [here](https://github.com/bitsandbytes-foundation/bitsandbytes/discussions/1338#discussioncomment-11991910). The model `google/siglip-so400m-patch14-384` hidden dims cannot be fully divided by blocksize.

bitsandbytes/backends/cpu/ops.py
jiqing-feng: fixed

bitsandbytes/functional.py
```diff
@@ -1123,6 +1123,16 @@ def dequantize_4bit(
...
 
+    # IPEX format is different, we need extra process.
+    if getattr(quant_state, "ipex", False) and quant_state.quant_type == "nf4":
+        return torch.ops.bitsandbytes.dequantize_nf4_ipex(
+            A,
+            absmax,
+            quant_state.blocksize,
+            quant_state.shape,
+            quant_state.dtype,
+        )
```
Devjiu, line 1134: Is there reason why this change can't be in `bitsandbytes/backends/cpu/ops.py`?

bitsandbytes/autograd/_functions.py
```diff
@@ -366,6 +422,8 @@ def matmul(
     state = state or MatmulLtState()
     if threshold > 0.0:
         state.threshold = threshold
+    if A.device.type in ("cpu", "xpu") and state.is_training:
```
Devjiu, line 425: Shouldn't there be an ipex availability check at least?

bitsandbytes/backends/cpu/ops.py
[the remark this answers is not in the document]
matthewdouglas: Thanks for pointing to that model - that helps to see that this is worth it.

I'm not really saying that it's logically complicated, but more that it is hard to read visually. There's the indexing to try to follow, and ambiguous naming like `A_com`. I'm not going to nitpick too much or block on it. I just think it could be easier to read but keep in mind I'm biased here coming from having a much a longer background in C# than in Python or anything ML adjacent. I don't have the same intuitions around abbreviations or blocks of long statements etc. So when I see three simple statements replaced with a dozen that are relatively more complex, I have to question the value-add.

I'm thinking of some minor changes, e.g. just some naming and whitespace mostly. Something like:

```python
    full_blocks = n // blocksize
    rem = n % blocksize
    blocks = full_blocks + 1 if rem else full_blocks
    
    absmax = torch.empty((blocks,), device=A.device, dtype=A.dtype)
    A_flattened = A.reshape(n)

    # Scale full blocks of the tensor to [-1, 1]
    A_full_blocks = A_flattened[: n - rem].reshape(n // blocksize, blocksize)
    absmax[:full_blocks] = torch.abs(A_full_blocks).max(dim=-1)[0]
    scaled = torch.clamp(A_full_blocks * (1 / absmax[:full_blocks].view(-1, 1)), -1, 1).reshape(-1)

    # Scale any partial block
    if rem:
        A_rem = A_flattened[-rem:]
        absmax[-1] = torch.abs(A_rem).max()
        scaled_rem = torch.clamp(A_rem * (1 / absmax[-1]), -1, 1)
        scaled = torch.cat([scaled, scaled_rem], dim=0)

    # Quantize with the lookup table
    quantized = torch.argmin(torch.abs(scaled.view(-1, 1) - CODE[quant_type]), dim=-1, keepdim=True).to(torch.uint8)
```

I also changed `torch.zeros` to `torch.empty` but otherwise expect to keep the logic the exact same.

bitsandbytes/functional.py
matthewdouglas: @Devjiu See here: https://huggingface.slack.com/archives/C07FCTCJKU0/p1746584773504889

tests/test_functional.py
```diff
@@ -94,16 +94,6 @@ class Test8BitBlockwiseQuantizeFunctional:
     @pytest.mark.parametrize("blocksize", [4096, 2048, 1024, 512, 256, 128, 64])
     @pytest.mark.parametrize("signed", TRUE_FALSE, ids=id_formatter("signed"))
     def test_dynamic_blockwise_quantization(self, device, dtype, nested, blocksize, signed):
-        if device == "cpu":
-            # This test is slow on CPU, so avoid atypical use cases.
-            if nested:
-                pytest.skip("Not a typical use case.")
-            if blocksize != 256:
-                pytest.skip("Only blocksize 256 is the typical one supported on CPU.")
```
matthewdouglas, line 102: I think we should still skip these, at least for the fp32 kernel. If the torch implementation for bf16/fp16 is that much faster then it may make sense to keep or put a `slow` marker on it so we could filter it out.

tests/test_functional.py
```diff
@@ -136,10 +126,10 @@ def test_dynamic_blockwise_quantization(self, device, dtype, nested, blocksize,
         abserr = sum(diffs) / len(diffs)
         relerr = sum(reldiffs) / len(reldiffs)
         if signed:
-            assert abserr < 0.0035
+            assert abserr < 0.0036
             assert relerr < 0.015
         else:
-            assert abserr < 0.00175
+            assert abserr < 0.0023
```
matthewdouglas, line 132: It's better if we didn't change this just for CPU/XPU support; if we need to let's adjust based on `device == "cpu"` and maybe in combination with `dtype` and `blocksize` for now.

tests/test_linear4bit.py
```diff
@@ -186,12 +180,6 @@ def test_linear_serialization(device, quant_type, compress_statistics, bias, qua
 @pytest.mark.parametrize("blocksize", [64, 128])
 @pytest.mark.parametrize("compress_statistics", TRUE_FALSE, ids=id_formatter("compress_statistics"))
 def test_copy_param(device, quant_type, blocksize, compress_statistics):
-    if device == "cpu":
-        if compress_statistics:
-            pytest.skip("Currently segfaults on CPU")
```
matthewdouglas, line 191: I'm still getting this segfault with the fp32 C++ implementation.

bitsandbytes/backends/cpu/ops.py
```diff
@@ -88,31 +88,62 @@ def _(A: torch.Tensor, absmax: torch.Tensor, code: torch.Tensor, blocksize: int,
...
+CODE = {"nf4": _NF4_QUANT_TABLE, "fp4": _FP4_QUANT_TABLE}
 
 
 @register_kernel("bitsandbytes::quantize_4bit", "cpu")
 def _(
     A: torch.Tensor, blocksize: int, quant_type: str, quant_storage: torch.dtype
 ) -> tuple[torch.Tensor, torch.Tensor]:
     torch._check_is_size(blocksize)
-    torch._check(quant_type == "nf4", lambda: f"quant_type must be nf4 on CPU, got {quant_type}")
+    torch._check(quant_type in ("nf4", "fp4"), lambda: f"quant_type must be nf4 or fp4 on CPU, got {quant_type}")
```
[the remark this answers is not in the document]
jiqing-feng: done

bitsandbytes/backends/cpu/ops.py
```diff
@@ -170,17 +216,13 @@ def _(
...
-        shape=shapeB,
-        dtype=A.dtype,
-    )
+    # Applied from dequantize_4bit
+    B = B.view(-1, 1)
+    upper = (B >> 4).to(torch.int64)
+    lower = (B & 0x0F).to(torch.int64)
+    blocks = torch.cat((upper, lower), dim=1).reshape(-1, blocksize)
+    B_dq = code[blocks] * absmax[:, None]
+    B_dq = B_dq.reshape(-1, *shapeB[1:]).to(A.dtype)
```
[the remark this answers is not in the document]
jiqing-feng: done

tests/test_linear4bit.py
jiqing-feng: That's because the input shape is incorrect, I've change the input tensor to a more reasonable shape. It could pass now.

tests/test_functional.py
jiqing-feng: fixed

tests/test_functional.py
jiqing-feng: done

bitsandbytes/autograd/_functions.py
jiqing-feng: OK, I checked if ipex enabled before it.

jiqing-feng: Checked ipex before using it.

jiqing-feng:
pytest --ignore test_optim.py --ignore test_triton.py --ignore test_cuda_setup_evaluator.py
 
 
CPU previous: 378 passed, 1537 failed, 1638 skipped, 197 xfailed, 153 warnings in 613.27s
CPU current: 2079 passed, 1498 skipped, 153 deselected, 9 xfailed, 59 warnings in 1192.94s
 
XPU previous: not enabled
XPU current: 2093 passed, 1493 skipped, 153 deselected, 63 warnings in 562.25s


It also could pass all [transformers tests](https://github.com/huggingface/transformers/tree/main/tests/quantization/bnb)

I also updated the installation guide.

Hi @matthewdouglas . Please take the next round review.

bitsandbytes/autograd/_functions.py
Devjiu: thanks

tests/test_linear4bit.py
matthewdouglas: I think the shape was OK but there's a bug that comes up with the specific data; seemed like a division by zero happens. Either way out of scope for this PR.

jiqing-feng: I see, we can have a seperate PR to fix the op issue. But for this test, it enabled double compress, so I don't think quantize a (64,) size tensor is what we want, the current shape (300, 400) is more reasonable. Please let me know if you still want to keep the original (64,) shape.

docs/source/installation.mdx
```diff
@@ -316,15 +316,29 @@ pip install -e .   # `-e` for "editable" install, when developing BNB (otherwise
 > [!TIP]
 > Intel CPU/XPU backend only supports building from source; for now, please follow the instructions below.
 
-It does not need compile CPP codes, all required ops are in [intel_extension_for_pytorch](https://pytorch-extension.intel.com/), please follow the instruction to install ipex.
+It requires [intel_extension_for_pytorch](https://pytorch-extension.intel.com/), please follow the instruction to install ipex.
```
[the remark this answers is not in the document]
jiqing-feng: Yes, I have updated the limitation.

tests/test_functional.py
jiqing-feng: I still want to keep the bf16/fp16 tests because we may need to enable absmax with fp16/bf16 dtype in the future.

jiqing-feng:
Hi @matthewdouglas . Please check if there are anything missed before merging. We can discuss the tests in this issue #1637

tests/test_linear4bit.py
matthewdouglas: Let's keep the new shape and I'll note for myself that the old shape just happened to surface a data-dependent bug in the C++ CPU implementation.

bitsandbytes/cextension.py
```diff
@@ -283,11 +283,28 @@ def get_native_library() -> BNBNativeLibrary:
...
+    ipex_xpu = None
+
+
 try:
     lib = get_native_library()
+    if not ipex_cpu:
+        logger.warning(
+            "The installed version of bitsandbytes was compiled without IPEX support. "
+            "You can install ipex by running `pip install intel_extension_for_pytorch`to get better performance if you use the Intel CPU.",
+        )
```
matthewdouglas, line 303: This seems like extra noise that we'd want to avoid.

Something to point out is that we still plan to ship libbitsandbytes_cpu in our wheels, so for most users, it's going to load a CPU, CUDA, or eventually ROCm or Metal library and we'll hit this logging line. At most we should really only raise this warning when:


1. We're on a platform with IPEX CPU support. My understanding is this is limited to Linux x86-64.
2. We expect the user to be using CPU, i.e. no CUDA, XPU, or MPS accelerators available. 
  On torch >= 2.6 we could just use `torch.accelerator.is_available()` and on older versions I think we can overlook privateuse1 backends like HPU or Ascend NPU.
3. There's some expectation of IPEX being beneficial. We don't want to prompt users to install it if e.g. it needs AVX512 or AMX support to be effective. This is something I can't speak to directly but defer to Intel folks to determine.

Any other thoughts @Titus-von-Koeller ?

bitsandbytes/autograd/_functions.py
```diff
@@ -366,6 +425,10 @@ def matmul(
     state = state or MatmulLtState()
     if threshold > 0.0:
         state.threshold = threshold
+    # MatMul8bitLt is slower because no fast kernel for quant/dequant 8bit in CPU/XPU
+    if state.is_training:
+        if (A.device.type == "cpu" and ipex_cpu) or (A.device.type == "xpu" and ipex_xpu):
```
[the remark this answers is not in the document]
jiqing-feng: Hi @Egor-Krivov . I am not saying it will be broken. It's just very slow without this branch. From my test, the MatMul8bitFp has 2x speed-up on CPU and 3x speed-up on XPU compared to MatMul8bitLt

bitsandbytes/cextension.py
jiqing-feng: You were right. I also agree that the log should only exist if no devices like cuda/xpu are available and the CPU is an Intel product.

jiqing-feng: I have changed it, please review again. Thanks!

jiqing-feng:
Hi @matthewdouglas . You can see some CPU ops are implemented by pure python, which works for all devices. Could we move these pure python implementations to the default folder? As these ops could be reused by XPU.

bitsandbytes/autograd/_functions.py
```diff
@@ -298,6 +299,64 @@ def backward(ctx: torch.autograd.function.FunctionCtx, grad_output: torch.Tensor
         return grad_A, grad_B, None, grad_bias, None
 
 
+class MatMul8bitFp(torch.autograd.Function):
+    # For Intel CPU and XPU, the double quant has many unsafe operations which will breaks the finetune.
```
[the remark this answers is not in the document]
jiqing-feng: Thanks! I've changed it?.

jiqing-feng:
Hi @matthewdouglas . I added xpu folder in the backend, but it seems didn't work. Do you know why?

jiqing-feng:
Hi @matthewdouglas . I removed pure pytorch ops to the default folder and it works well. The CIs are all passed. Please continue to review this PR.

PR #1486, opened 2025-01-23
Exclude tests from distribution

akx:
Fixes #1478.

It could be a good idea to switch to https://hatch.pypa.io/latest/ instead of `setuptools.build_meta` at some point.

PR #1650, opened 2025-05-24
Renamed tests directory — don't do that people.

excitoon:
image


matthewdouglas:
Hi @excitoon,

Thank you for the suggestion. If I am understanding correctly, the concern is regarding the `tests` being distributed and placed in `site-packages/tests`. This was already handled more appropriately by #1486.

matthewdouglas:
@excitoon 

Something worth mentioning is that there was never official macOS support in bitsandbytes. Unfortunately, wheels from 0.42 and earlier were tagged as platform-agnostic, when in fact they included only binaries built for Linux x86-64 systems. 

In 0.43 we added Windows support and proceeded to correct the binary wheels by separately tagging them for `manylinux_2_24_x86_64` and `win_amd64`. 

There was also an sdist in 0.42, but it's not built properly with the Python tooling (e.g. setup.py). It was removed in 0.43 because it does not work properly out of the box on systems that we do support. We have a PR that we'll be looking at to bring proper sdist support: #1512. I think this particular issue can best be addressed there and the related issue: #1511.

This all means that on any platform without support, like macOS, `pip`will find `bitsandbytes-0.42.0-py3-none-any.whl` as the latest "supported" wheel from PyPI. That's also the latest if looking for an sdist. Even so, 0.42 and earlier cannot actually be built for macOS.

We will soon distribute wheels for macOS 13.1+ on arm64 as well, which hopefully improves that situation. Until then I would suggest that macOS users do not install bitsandbytes at all if they can avoid it; upstream projects using it should exclude it from their dependencies on macOS until we do provide support for it.

PR #1629, opened 2025-05-08
[Triton/XPU] Support 4bit dequantization logic on Triton

Devjiu:
This PR adds xpu backend and Triton kernel for dequantization nf4 dtype.
Trtion used as an optional import.

Tests:
- `tests/test_functional.py::TestQuantize4BitFunctional` supported nf4/fp4 cases
- `tests/test_functional.py::Test8BitBlockwiseQuantizeFunctional` implemented quantize_blockwise with binary search that works faster for XPU
- `tests/test_linear4bit.py`

Signed-off-by: Dmitrii Makarenko <dmitrii.makarenko@intel.com>

PR #1661, opened 2025-05-29
update bnb log

jiqing-feng:


bitsandbytes/cextension.py
```diff
@@ -298,7 +298,16 @@ def get_native_library() -> BNBNativeLibrary:
...
 try:
-    lib = get_native_library()
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        if not (ipex_cpu or ipex_xpu):
+            logger.warning(
+                "Detected Intel XPU but no Intel Extension for PyTorch (IPEX) installed. "
+                "IPEX is recommended for Intel XPU support in bitsandbytes to get better performance. "
+                "Please check the installation doc to install `intel_extension_for_pytorch`. "
+            )
+        lib = ErrorHandlerMockBNBNativeLibrary("XPU does not need native library")
```
yao-matrix, line 308: why need `ErrorHandlerMockBNBNativeLibrary`

bitsandbytes/cextension.py
```diff
@@ -298,7 +298,16 @@ def get_native_library() -> BNBNativeLibrary:
 
 
 try:
-    lib = get_native_library()
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        if not ipex_xpu:
+            logger.warning(
+                "Detected Intel XPU but no Intel Extension for PyTorch (IPEX) installed. "
+                "IPEX is recommended for Intel XPU support in bitsandbytes to get better performance. "
```
Devjiu, line 305: Can we please mention Triton here? 
For example: "XPU is detected, but IPEX is missing, will try to use Trtiton implementation. If performance is lower your expectations, consider installing IPEX"

The current wording effectively prohibits the use of XPU without IPEX.

jiqing-feng: fixed

jiqing-feng:
We need to use mock lib because of the [functional variable](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/bitsandbytes/functional.py#L25).

jiqing-feng:
Hi @matthewdouglas . This PR is ready to be reviewed, please let me know your comments. Thanks!

Egor-Krivov:
I don't think we should push user into a specific XPU backend, until we clearly know which one is likely faster. 
I just run simple inference benchmark and couldn't get performance improvements from IPEX: https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1629#issuecomment-2939220472

jiqing-feng:
> I don't think we should push user into a specific XPU backend, until we clearly know which one is likely faster. I just run simple inference benchmark and couldn't get performance improvements from IPEX: [#1629 (comment)](https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1629#issuecomment-2939220472)

You were right, I have changed the log. This PR will avoid xpu through load native library error, as we don't want to load native lib on XPU. Please review my new changes. Without this PR, the error be like:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/bitsandbytes/cextension.py", line 115, in <module>
    lib = get_native_library()
          ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/bitsandbytes/cextension.py", line 86, in get_native_library
    dll = ct.cdll.LoadLibrary(str(binary_path))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ctypes/__init__.py", line 454, in LoadLibrary
    return self._dlltype(name)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/ctypes/__init__.py", line 376, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: /usr/local/lib/python3.11/dist-packages/bitsandbytes/libbitsandbytes_cpu.so: cannot open shared object file: No such file or directory
```

bitsandbytes/cextension.py
```diff
@@ -298,7 +298,15 @@ def get_native_library() -> BNBNativeLibrary:
...
 
 try:
-    lib = get_native_library()
+    if hasattr(torch, "xpu") and torch.xpu.is_available():
+        if not ipex_xpu:
+            logger.warning(
+                "Detected Intel XPU without intel_extension_for_pytorch installed. Triton implementation will be used."
+            )
+        # create a mock with error messaging as fallback
+        lib = ErrorHandlerMockBNBNativeLibrary("XPU does not need library loading")
```
matthewdouglas, line 307: Wouldn't we still want it in case the user has a reason to use the two CPU ops it implements? We don't necessarily know they won't just because an XPU is available.

jiqing-feng:
Close this PR as we may change the XPU behavior in the future.

PR #1705, opened 2025-07-10
[XPU]Add the 32-bit / 8-bit adam optimizer implementation for XPU.

YangKai0616:
We implemented the **32-bit / 8-bit Adam optimizer** on **XPU** using **Triton**. 
The UT files were modified to test the optimizer's performance on XPU. 
Implementing the optimizer with Triton could be a great starting point for compatibility across multiple platforms like XPU and CUDA. 
Moving forward, we will continue to implement more optimizers!


YangKai0616:
@matthewdouglas, please help review, thanks!

yao-matrix:
@Devjiu, could you pls help review, too? Thx

PR #1720, opened 2025-08-01
4bit quantization for arbitrary `nn.Parameter`

matthewdouglas:
This PR is in the same spirit as the recently introduced feature in huggingface/peft#2638.

Several models exist in the Hugging Face ecosystem where there are MoE layers that use `nn.Parameter` and are not compatible with the default quantization approach of replacing `nn.Linear`. Such example models include, but are not limited to:

* [Llama 4](https://huggingface.co/docs/transformers/en/model_doc/llama4)
* [Granite MoE](https://huggingface.co/docs/transformers/en/model_doc/granitemoe)
* [DBRX](https://huggingface.co/docs/transformers/en/model_doc/dbrx)
* [JetMoE](https://huggingface.co/docs/transformers/en/model_doc/jetmoe)

A new utility, `bitsandbytes.nn.parametrize.replace_parameter_4bit()` is introduced. This will quantize and replace an `nn.Parameter` with a parametrization layer which automatically dequantizes the parameter when it is accessed

Additional work will be done on the HF Transformers side to enable integration with options in `BitsAndBytesConfig`.
.

### DIFF
--- a/bitsandbytes/nn/parametrize.py
+++ b/bitsandbytes/nn/parametrize.py
@@ -1,0 +1,149 @@
+from functools import partial
+from typing import Any, Literal, Optional
+
+import torch
+import torch.nn as nn
+import torch.nn.utils.parametrize as P
+
+from .. import functional as F
+
+
+class Bnb4bitParametrization(nn.Module):
+    """
+    A parametrization module that handles dequantization of a 4-bit quantized parameter.
+
+    The parameter data is expected to be already quantized when this parametrization is applied.
+    This module will dequantize the parameter data to its original floating-point representation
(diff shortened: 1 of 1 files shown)


winglian:
Will there need to be any changes in PEFT to apply lora adapters to quantized parameters once this lands?

BenjaminBossan:
We'll have to test, but at the very least, https://github.com/huggingface/peft/pull/2710 needs to be merged in PEFT for this to work properly.

cmp-nct:
> We'll have to test, but at the very least, [huggingface/peft#2710](https://github.com/huggingface/peft/pull/2710) needs to be merged in PEFT for this to work properly.

it's merged

PR #1729, opened 2025-08-18
Fix QuantState and dict conversions

cyyever:
Fix the failure of QuantState.as_dict then followed by QuantState.from_dict.
A reproducing example is 
```
import torch

from bitsandbytes.functional import (
    QuantState,
    dequantize_4bit,
    quantize_4bit,
)


a = torch.rand(2, 3)
quantized, quant_state = quantize_4bit(A=a, quant_type="nf4")
quant_dict = quant_state.as_dict()
quant_state = QuantState.from_dict(quant_dict, device=torch.device("cuda"))
quantized = dequantize_4bit(A=quantized, quant_state=quant_state)
```
Before this PR, it failed with (on bitsandbytes 0.47.0):
```
  File "/home/cyy/a.py", line 13, in <module>
    quant_state = QuantState.from_dict(quant_dict, device=torch.device("cuda"))
  File "/home/cyy/.local/lib/python3.13/site-packages/bitsandbytes/functional.py", line 469, in from_dict
    raise ValueError(
        f"There should be exactly one `quant_state` item with ending from {cls.valid_qs_type_keys}.\nDetected {qs_key}.",
    )
ValueError: There should be exactly one `quant_state` item with ending from ['bitsandbytes__fp4', 'bitsandbytes__nf4'].
Detected [].
```

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -20,7 +20,7 @@ jobs:
         platform: [linux-x64, linux-aarch64, macos, windows]
         # default runners don't have AVX-512 support, but icelake does
         cpu_type: ["", icelake]
-        torch_version: ["2.3.1", "2.8.0", "2.9.1"]
+        torch_version: ["2.3.1", "2.9.1", "2.10.0"]
 
         exclude:
           # aarch64 minimum torch version is 2.5.1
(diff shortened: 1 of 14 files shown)


matthewdouglas:
Hi,
Can you provide more detail on the issue that this fixes? A simple reproducer with example output would be ideal. Thanks!

cyyever:
@matthewdouglas Added

bitsandbytes/functional.py
```diff
@@ -531,7 +532,10 @@ def as_dict(self, packed=False):
         # packed format allows serialization of non-tensor components, critical for saving in safetensors format
         qs_packed_dict = {k: v for k, v in qs_dict.items() if isinstance(v, torch.Tensor)}
         non_tensor_dict = {k: v for k, v in qs_dict.items() if not isinstance(v, torch.Tensor)}
-        qs_packed_dict["quant_state." + "bitsandbytes__" + self.quant_type] = pack_dict_to_tensor(non_tensor_dict)
+        key = "quant_state.bitsandbytes__"
```
TimDettmers: Latent issue (non-blocking): When `quant_type is None` and `as_dict(packed=True)` is called, the key becomes `"quant_state.bitsandbytes__"` (no suffix). If `from_dict` later receives this packed dict, the `qs_key[0].split(".")[-1]` check on line 470 would reject it because `"bitsandbytes__"` is not in `valid_qs_type_keys`. This is not a practical problem today since blockwise QuantState is never serialized via the packed path, but it means the packed round-trip is still incomplete for `quant_type=None`. Consider adding `if self.quant_type is None: return qs_dict` before the packed section (i.e., always return unpacked for blockwise QuantState) to make the intent explicit.

PR #1735, opened 2025-08-23
[NVIDIA] CUDA 13 Support

johnnynunez:
Enable Thor and Spark
Enable CUDA 13 
@matthewdouglas you can clean code. https://pypi.jetson-ai-lab.io/sbsa/cu130/bitsandbytes/0.48.0.dev0

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,8 +115,6 @@ if(BUILD_CUDA)
 
     if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS "11.4")
         message(FATAL_ERROR "CUDA Version < 11.4 is not supported")
-    elseif(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL "13.0")
-        message(FATAL_ERROR "CUDA Version > 12 is not supported")
     endif()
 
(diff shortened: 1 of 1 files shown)


matthewdouglas:
Thanks! I'm going to hold on this for a moment - have to check and see if a nightly torch cu130 build is ready to fully test this. We also need to fix the backwards compatibility issues with CUDA < 12.9, since the included CCCL for those versions looks to be too old for using e.g. `cuda::maximum`. I suppose the options there will be to either guard around this or to pull in the newest CCCL separately.

johnnynunez:
> Thanks! I'm going to hold on this for a moment - have to check and see if a nightly torch cu130 build is ready to fully test this. We also need to fix the backwards compatibility issues with CUDA < 12.9, since the included CCCL for those versions looks to be too old for using e.g. `cuda::maximum`. I suppose the options there will be to either guard around this or to pull in the newest CCCL separately.

for sure, really appreciate it, it is was only the start way. Feel free to change it

matthewdouglas:
Thanks, I'm going to close this in favor of #1761 which takes this further to handle backwards compatibility, packaging, docs, etc.

PR #1679, opened 2025-06-15
Add SYCL Kernels for XPU backend

xiaolil1:
This is the pull request for the SYCL Kernels targeting the XPU backend.

- It features the implementation of the "dequantize_blockwise," "dequantize_4bit," and "dequant & gemv_4bit fusion" kernels.
- The target low-precision quantization datatypes encompass NF4, FP4 and General8bits. 
- This PR aims to eliminate the dependency on IPEX and improve the performance.

PR #1763, opened 2025-09-24
Build/Package Intel XPU binary for Linux

matthewdouglas:
This PR is introduced in order to add a build/packaging workflow for the SYCL kernels that were introduced in #1679.

The build produces `libbitsandbytes_xpu.so` for the Linux x86-64 platform. However, I have not yet been able to verify that this build works as expected.

As a side note, this build has a minimum glibc >= 2.34 dependency, and we currently tag our wheels as `manylinux_2_24_x86_64`. This requirement is reasonable for XPU, but we want to keep the CUDA support as broad as possible, so I may open a separate followup PR to bundle separate manylinux_2_34 and manylinux_2_24 wheels, with and without XPU support respectively. 

@jiqing-feng @xiaolil1 Please review and let me know if there's any issues with this. Thanks!

### DIFF
--- a/.github/scripts/build-xpu.sh
+++ b/.github/scripts/build-xpu.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+declare build_os
+
+set -xeuo pipefail
+
+# We currently only build XPU on Linux.
+if [ "${build_os:0:6}" == ubuntu ]; then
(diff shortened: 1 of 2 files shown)


jiqing-feng:
I've verified on my XPU node. It works well!

Issue #1332, opened 2024-08-24
Linear8bitLt can not be moved back to cpu


Nerogar:
### System Info

During inference of larger models in VRAM constrained environments, offloading unused model layers from VRAM to RAM is an easy method of reducing the overall VRAM usage. Linear8bitLt can not be moved back to cpu memory once forward() is called on them. There are two issues:
1. The internal `state` object is ignored during the `to("cpu")` call. All tensors remain in VRAM
2. Moving the layer back to cuda calls the quantization logic again, breaking the model

### Reproduction

```python
import torch
import bitsandbytes as bnb

if __name__ == '__main__':
    # initialize a simple model
    generator = torch.Generator()
    generator.manual_seed(42)

    linear = bnb.nn.Linear8bitLt(
        input_features=32,
        output_features=32,
        bias=False,
        has_fp16_weights=False,
    )
    x_in = torch.randn(size=(1, 32), generator=generator)
    torch.nn.init.xavier_uniform(linear.weight, generator=generator)

    # move everything to CUDA for the first time
    linear.to("cuda")
    x_in = x_in.to("cuda")

    # call the model once to get a "good" result
    x_out_1 = linear(x_in)
    print(f"output 1: {x_out_1}")

    # move the model to cpu and observe that some tensors are still stored in VRAM
    linear.to("cpu")
    print(f"CxB device after cpu offloading: {linear.state.CxB.device}")
    linear.to("cuda")

    # call the model again after moving it to CUDA
    x_out_2 = linear(x_in)
    print(f"output 1: {x_out_2}")
```

Output:

output 1: tensor([[ 0.0978, -0.4744,  0.0976, -1.7158,  0.3936,  0.7334, -0.6406,  0.5264,
          1.7373, -1.0938, -1.0625, -0.3091,  0.9946,  2.1582,  0.4675,  1.2090,
         -0.7349, -0.2979,  0.6055,  0.1614,  1.0742, -2.6758, -2.2266,  0.7310,
         -1.5635,  0.1646, -0.0526,  0.4590,  1.0068, -1.6650,  0.5469,  0.1232]],
       device='cuda:0')

CxB device after cpu offloading: cuda:0

output 2: tensor([[   41.0000,  -198.3750,    40.5000,  -716.0000,   173.3750,   314.5000,
          -267.0000,   219.5000,   731.0000,  -459.5000,  -444.7500,  -134.5000,
[...]

PR #1765, opened 2025-09-25
fix peft cpu ut failure

yao-matrix:
in `to` only `data` of param moved to the specified device, but `SCB` and `CB` still keep in the before device, this will lead to `peft` below ut fail 
```
pytest -rA tests/test_gpu_examples.py::TestLoftQ::test_bloomz_loftq_8bit[cpu]
```

Since in this code https://github.com/huggingface/peft/blob/main/src/peft/utils/integrations.py#L107, peft will find `weight.data` has different placement from `SCB`, so kind of only partial `to` was done.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -679,6 +679,7 @@ class Int8Params(torch.nn.Parameter):
     def to(self, *args, **kwargs):
         device, dtype, non_blocking, convert_to_format = torch._C._nn._parse_to(*args, **kwargs)
 
+
         if device is not None and device.type != "meta" and self.data.device.type == "cpu":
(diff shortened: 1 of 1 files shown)


yao-matrix:
@matthewdouglas @jiqing-feng , pls help review and comment, thx very much.

jiqing-feng (approved):

matthewdouglas:
Makes sense, thanks! Let's just fix the lint issue and we can merge it.

Related issue: #1332

matthewdouglas (approved):

yao-matrix:
@matthewdouglas , fixed lint, thx very much.

matthewdouglas:
Thanks @yao-matrix. I've taken this a little bit further in #1769 and will merge that instead.

PR #1774, opened 2025-10-01
Add trove-classifiers requirement to pyproject.toml

### LINKED ISSUE #1773: Requires a recent `trove-classifiers`
### System Info

Because of https://github.com/bitsandbytes-foundation/bitsandbytes/blob/00c0a1aab2466d37fd516578c6f19360e8e6aca2/pyproject.toml#L42
```
"Programming Language :: Python :: 3.13",
```

a recent `trove-classifiers` version is needed. To be precise `trove-classifiers>=2025.8.6.13`

This should be added to build requirements as otherwise it fails building.

### Reproduction

`pip wheel .` (with a `trove-classfiers==2025.5.9.12`)

### Expected behavior

Build and install the `trove-classifiers` in order to build correctly

ccoulombe:
Fixes #1773

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["scikit-build-core", "setuptools >= 77.0.3"]
+requires = ["scikit-build-core", "setuptools >= 77.0.3", "trove-classifiers>=2025.8.6.13"]
 build-backend = "scikit_build_core.setuptools.build_meta"
 
 [project]


matthewdouglas:
Thanks for the PR!

My understanding is that this should only be significant when you've already installed an older version of `trove-classifiers` in the build environment, e.g. using `--no-build-isolation`. Happy to merge though!

PR #1781, opened 2025-10-10
[NVIDIA] Add Orin, GB300, Spark Support

johnnynunez:


.github/scripts/build-cuda.sh
```diff
@@ -12,19 +12,19 @@ elif [ "${build_arch}" = "aarch64" ]; then
...
     # CUDA 13.0+: Add sm100/sm110/sm120
-    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;90;100;110;120"
+    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;87;90;100;103;110;120;121"
 else
     # By default, target Pascal through Hopper.
     build_capability="60;70;75;80;86;89;90"
 
     # CUDA 12.8+: Add sm100 and sm120; remove < sm70 to align with PyTorch 2.8+cu128 minimum
-    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="70;75;80;86;89;90;100;120"
+    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="70;75;80;87;86;89;90;100;120;121"
```
matthewdouglas: My understanding is that Orin (sm87) and Spark (GB10, sm121) are only available on aarch64 platforms, so we shouldn't need to do this for x86-64.

.github/scripts/build-cuda.sh
```diff
@@ -12,19 +12,19 @@ elif [ "${build_arch}" = "aarch64" ]; then
     build_capability="75;80;90"
 
     # CUDA 12.8-12.9: Add sm100/sm120
-    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="75;80;90;100;120"
+    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="75;80;87;90;100;120"
 
     # CUDA 13.0+: Add sm100/sm110/sm120
-    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;90;100;110;120"
+    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;87;90;100;103;110;120;121"
```
matthewdouglas: I would have expected building for sm80, sm100, and sm120 to cover this as we don't use any specific features in sm87/sm103/sm121 yet. Can you clarify on the benefit of adding these targets?

.github/scripts/build-cuda.sh
```diff
@@ -12,19 +12,19 @@ elif [ "${build_arch}" = "aarch64" ]; then
...
     # By default, target Pascal through Hopper.
     build_capability="60;70;75;80;86;89;90"
 
     # CUDA 12.8+: Add sm100 and sm120; remove < sm70 to align with PyTorch 2.8+cu128 minimum
-    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="70;75;80;86;89;90;100;120"
+    [[ "${cuda_version}" == 12.8.* || "${cuda_version}" == 12.9.* ]] && build_capability="70;75;80;87;86;89;90;100;120;121"
 
     # CUDA 13.0+: Remove < sm75 to align with PyTorch 2.9+cu130 minimum
-    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;86;89;90;100;120"
+    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;87;90;100;103;110;120;121"
```
matthewdouglas: Same comment on sm87, sm110, sm121 being exclusive to aarch64.

I think we should keep sm86/sm89, and maybe consider adding sm103.

```suggestion
    [[ "${cuda_version}" == 13.*.* ]] && build_capability="75;80;86;89;90;100;103;120"
```

johnnynunez: Right

PR #1790, opened 2025-10-29
mps: add nf4 dequantize/quantize kernel

bghira:
this ports the CUDA NF4 support to Metal.

so far, I've targeted nf4 quant/dequant because it's one of the least-accessible formats for Mac users.

we're using uint8 under the hood. for what it's worth, Metal (and the underlying hardware) lacks fp8/fp4 support.

performance has not been the forefront of this effort, as most of the time was spent determining how to plug metallib into bitsandbytes and correctly build it.

I'd like some feedback on this approach, because due to my inexperience with your build toolchain, it's highly likely I've done things in ways that can be improved.

I'm building on lessons I'd learnt while building a pytorch custom op for universal-metal-flash-attention, namely the way the MTLBuffers are retrieved from torch MPSGraph objects, which required the use of the torch headers.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -103,7 +103,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04, ubuntu-22.04-arm, windows-2025, macos-15]
         # Test with the oldest supported torch version, the newest two stable/RC.
-        torch_version: ["2.3.1", "2.7.1", "2.8.0"]
+        torch_version: ["2.3.1", "2.8.0", "2.9.0"]
         include:
           - os: ubuntu-22.04
             arch: x86_64
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.11.2
+    rev: v0.14.3
     hooks:
       - id: ruff
         args:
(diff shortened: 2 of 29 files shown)


bghira:
@rickardp cc

bghira:
@matthewdouglas if we can get this reviewed and merged, i can continue adding Metal support.

matthewdouglas:
Hi @bghira 

For the time being, I'd like to avoid linking in libtorch. Doing so adds more complexity to our packaging process, and IMO it is not worth it for only this. We aim to support a reasonably broad range of PyTorch versions; with linking to libtorch we'd have to either build for each of those and somehow distribute them all, or pin the version of PyTorch. 

Eventually we'll consider using the newer [LibTorch Stable ABI](https://docs.pytorch.org/docs/stable/notes/libtorch_stable_abi.html), but that would be quite a while away, and we're only going to do it if there's very clear benefit to doing so. 

Instead, I ask that native code is built independently of PyTorch, and ideally exposes a C API that can be used in the same way that the CPU, CUDA, ROCm, and XPU backends work. Ideally we would not build as a Python extension that links to cpython either, but if so, I would also ask that it uses one build for all Python 3.10+ versions, e.g. with the [Stable ABI](https://docs.python.org/3/c-api/stable.html#stable-application-binary-interface).

I'm not super familiar here, but it seems the main things we need out of Torch are:
* `torch::mps::get_dispatch_queue()`
* `torch::mps::get_command_buffer()`
* `torch::mps::commit()`

It would be great to find a way around that. As mentioned on Discord, `torch.mps.compile_shader()` may also be an option. Do note that you can still write the shader in a separate file and read it in. 

I understand that this is a little different then most PyTorch extensions with native code, but it's intentional to help us with distribution and broad compatibility.

bghira:
understood;  Metal support is basically not possible to take any further in this library without substantial performance drawbacks.

PR #1796, opened 2025-11-03
Fix indexing overflow issue for blockwise quantization on AMD

sstamenk:
This PR ports the changes from #1784 to kernels.hip and ops.hip so that the newly added test_dynamic_blockwise_quantization_large test can pass on AMD GPUs.  Without this change the test gets aborted.

Tested this on both W7900 (gfx1100) and R9700 (gfx1201) and all 3 unit tests passed successfully.


matthewdouglas (approved):
Thanks, LGTM!

PR #1804, opened 2025-11-14
Cpu fused kernel

jiqing-feng:
The fused kernel optimized 4bit model inference about 4x speed-up on TPOT compared dequant+matmul. For next optimization of TTFT, we need to import libxsmm.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -515,11 +515,11 @@ class Linear4bit(nn.Linear):
         save weight and bias,
         then fill state_dict with components of quant_state
         """
-        super()._save_to_state_dict(destination, prefix, keep_vars)  # saving weight and bias
         if getattr(self.weight.quant_state, "packing_format_for_cpu", False):
             self.weight.data, self.weight.quant_state = _convert_weight_packed_for_cpu_inverse(
                 self.weight.data, self.weight.quant_state
             )
+        super()._save_to_state_dict(destination, prefix, keep_vars)  # saving weight and bias
         if getattr(self.weight, "quant_state", None) is not None:
             for k, v in self.weight.quant_state.as_dict(packed=True).items():
                 destination[prefix + "weight." + k] = v if keep_vars else v.detach()
--- a/bitsandbytes/functional.py
+++ b/bitsandbytes/functional.py
@@ -2214,6 +2214,7 @@ def _convert_weight_packed_for_cpu_inverse(
         recovered_state.absmax = qabsmax
         recovered_state.offset = offset
         recovered_state.state2 = state2
+        recovered_state.nested = True
 
     recovered_state.dtype = recovered_state.original_dtype
     recovered_state.packing_format_for_cpu = False


jiqing-feng:
Hi @matthewdouglas . All CI passed, I have rebased the PR. Please let me know what need to be changed before merge. Thanks!

CMakeLists.txt
```diff
@@ -280,12 +280,24 @@ if (BUILD_CPU)
         include(CheckCXXCompilerFlag)
         check_cxx_compiler_flag(-mavx512f HAS_AVX512F_FLAG)
         check_cxx_compiler_flag(-mavx512bf16 HAS_AVX512BF16_FLAG)
+        check_cxx_compiler_flag(-mavx512dq HAS_AVX512DQ)
+        check_cxx_compiler_flag(-mavx512bw HAS_AVX512BW)
+        check_cxx_compiler_flag(-mavx512vl HAS_AVX512VL)
```
matthewdouglas: I don't think we need to bother testing for these and can just safely add them when adding `-mavx512f`.

jiqing-feng:
Hi @matthewdouglas . The BNB will only load 1 lib (one from cpu/cuda/xpu). It means we can only build 1 .so file for the bnb. But we cannot build CPU and XPU together, because CPU relies on openMP(libiomp5.so) but XPU relies on GNU OpenMP (libgomp.so), build them together will raise error like: `libbitsandbytes_xpu.so: undefined symbol: __kmpc_for_static_init_8`. I suppose it's same for cuda. But without OMP, the CPU kernel might be even worse than python op, and there might be other incompatible flags across different backends.

In the current stage, we can only consider to build one backend, so the format cpu will not be triggered in other backends. Even though, I added the reverse logic in case we want to support multi-backends in the future. 

cc @SunMarc

csrc/cpu_ops.cpp
```diff
@@ -258,6 +226,241 @@ void quantize_cpu(float* code, float* A, float* absmax, unsigned char* out, long
...
+template <typename scalar_t, int BLOCK_M, int BLOCK_N, int DATA_TYPE> struct tinygemm_kernel_nn {
+    static inline void apply(
+        const scalar_t*, const unsigned char*, scalar_t*, const scalar_t*, int64_t, int, int64_t, int64_t, int64_t,
+        int64_t, int64_t
+    ) {
+        static_assert(sizeof(scalar_t) == 0, "tinygemm_kernel_nn primary template should never be instantiated");
+    }
+};
+
+template <int BLOCK_M, int BLOCK_N, int DATA_TYPE> struct tinygemm_kernel_nn<bf16_t, BLOCK_M, BLOCK_N, DATA_TYPE> {
```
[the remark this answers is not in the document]
jiqing-feng: Appreciate it. Thanks!

matthewdouglas:
Hi @jiqing-feng 

If you rebase it should trigger CI to run the tests on the PR now.

In regard to building the optimized CPU code when you're using an accelerator, you're right, I suppose has_avx512() would return false regardless of your CPU since we're only building that in the CPU-only lib? In the future I might want to change things so we always use a separate CPU library build instead of including CPU code in the accelerator libs. For now, it's a good point, so we shouldn't be rearranging weights if you're using an accelerator at all.

I'll look in more detail next week but another question I have is the naming of the gemv. It seems like this is actually a full GEMM implementation? I realize I never made a custom op for full GEMM so that's something maybe I can do in a followup PR.

jiqing-feng:
Yes, exactly! We should never rearrange weight if we are using an accelerator; it's for CPU-only. 
Yes, the cpu gemv kernel contains a full gemm implementation because it's a fused kernel. We do dequant and gemm in one kernel, that's where the speed-up is from.

OK, would like to see your next round review. It would be better to give me an approximate time when we can merge this PR so I can plan for the next feature. Thanks!

jiqing-feng:
Hi @matthewdouglas . Please trigger the CI. Thanks!

bitsandbytes/nn/modules.py
```diff
@@ -311,9 +316,13 @@ def cpu(self):
         return self.to(device="cpu")
 
     def cuda(self, device: Optional[int | device | str] = None, non_blocking: bool = False):
+        if getattr(self.quant_state, "packing_format_for_cpu", False):
+            self.data, self.quant_state = _convert_weight_packed_for_cpu_inverse(self.data, self.quant_state)
```
matthewdouglas: I think maybe this is something that would be better done in `to()` but since you make the point that we don't do the avx512 CPU build when you're using CUDA/XPU, it won't really matter yet - shouldn't end up repacking. So I'm OK to come back to it later if we separate CPU code out of the accelerator libs.

jiqing-feng:
Hi @matthewdouglas . I've fixed and verified the save and re load. You can check it by the following script.

```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
import shutil
import os

model_id = "hugging-quants/Meta-Llama-3.1-8B-Instruct-BNB-NF4"
save_path = "./local_test_model"
input_text = "What is bitsandbytes?"

if os.path.exists(save_path):
    shutil.rmtree(save_path)

def run_generation(model, tokenizer, prompt, tag):
    inputs = tokenizer(prompt, return_tensors="pt").to(model.device)
    
    print(f"\n[{tag}] Start Generating...")
    with torch.no_grad():
        outputs = model.generate(
            **inputs, 
            max_new_tokens=10, 
            do_sample=False,
            pad_token_id=tokenizer.eos_token_id
        )
    
    output_str = tokenizer.decode(outputs[0], skip_special_tokens=True)
    print(f"[{tag}] Output: {output_str}")
print(">>> Loading original model on CPU...")
tokenizer = AutoTokenizer.from_pretrained(model_id)
    model_id,
    device_map="cpu",
    torch_dtype=torch.bfloat16,
    low_cpu_mem_usage=True
)
run_generation(model, tokenizer, input_text, "Run-1")
print(f"\n>>> Saving model to {save_path}...")
model.save_pretrained(save_path)
tokenizer.save_pretrained(save_path)
del model
del tokenizer
import gc
gc.collect()
print(f"\n>>> Reloading model from {save_path}...")
loaded_model = AutoModelForCausalLM.from_pretrained(
    save_path,
    device_map="cpu",
    torch_dtype=torch.bfloat16,
    low_cpu_mem_usage=True
)
loaded_tokenizer = AutoTokenizer.from_pretrained(save_path)

run_generation(loaded_model, loaded_tokenizer, input_text, "Run-2")
```

Output:
```
>>> Loading original model on CPU...
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00, 35.76it/s]

[Run-1] Start Generating...
The following generation flags are not valid and may be ignored: ['temperature', 'top_p']. Set `TRANSFORMERS_VERBOSITY=info` for more details.
[Run-1] Output: What is bitsandbytes? Bitsandbytes is a free online tool that allows

>>> Saving model to ./local_test_model...

>>> Reloading model from ./local_test_model...
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00, 35.76it/s]

[Run-2] Start Generating...
The following generation flags are not valid and may be ignored: ['temperature', 'top_p']. Set `TRANSFORMERS_VERBOSITY=info` for more details.
[Run-2] Output: What is bitsandbytes? Bitsandbytes is a free online tool that allows
```

matthewdouglas:
Looks good, thanks!

PR #1809, opened 2025-11-14
Replace NULL with nullptr in pythonInterface.cpp

yuguo68:
for better type safety.


matthewdouglas:
This is reasonable, thanks!

PR #1811, opened 2025-11-21
CI: Run tests on PRs, refactor nightly test workflow

matthewdouglas:
This PR creates a new workflow for running some of our tests on PRs. We reuse some of our nightly test workflow, so it is refactored into a callable workflow that we can use for both nightly and PRs.

On PRs we will run a more limited set of tests. E.g. the Windows CPU tests are skipped as they're quite slow. We also will only run two versions of PyTorch (oldest supported and latest stable).

Additionally we will now run Linux x64 CPU tests on both an Ice Lake Xeon (with AVX512 support) and the default GitHub runner (AMD EPYC, Zen3, no AVX512 support).

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -1,8 +1,6 @@
 name: Nightly Tests
 
 on:
-  push:
-    branches: [pr-test-workflow]
   workflow_dispatch:
   schedule:
     # Every day at 02:15 AM UTC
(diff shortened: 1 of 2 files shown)


matthewdouglas:
I should note that on at least one run with the GitHub runner it used an Ice Lake chip - Xeon 8370C, so it's not really guaranteed that we will always get the EPYC. Most of the time we seem to get the EPYC though, and if we need to later set up a runner guaranteed to not have AVX512 support we could look into that.

PR #1829, opened 2025-12-11
Add release for DGX Spark cuda121

mfuntowicz:
DGX Spark requires the CUDA 121 compute capabilities but aarch64 builds don't include this target. 
This PR brings it in so it becomes pip installable


matthewdouglas:
Makes sense to specifically target sm121 on aarch64, thanks!

PR #1836, opened 2025-12-22
Create unistd.h

asmertpc-cloud:
fix for windows

### DIFF
--- a/csrc/unistd.h
+++ b/csrc/unistd.h
@@ -1,0 +1,61 @@
+// Source - https://stackoverflow.com/a
+// Posted by AShelly, modified by community. See post 'Timeline' for change history
+// Retrieved 2025-12-21, License - CC BY-SA 4.0
+
+#ifndef _UNISTD_H
+#define _UNISTD_H    1
+
+/* This is intended as a drop-in replacement for unistd.h on Windows.
+ * Please add functionality as needed.
+ * https://stackoverflow.com/a/826027/1202830
+ */
+
+#include <stdlib.h>
+#include <io.h>
+#include <getopt.h> /* getopt at: https://gist.github.com/ashelly/7776712 */
+#include <process.h> /* for getpid() and the exec..() family */
+#include <direct.h> /* for _getcwd() and _chdir() */
+
+#define srandom srand
+#define random rand
+
+/* Values for the second argument to access.
+   These may be OR'd together.  */
+#define R_OK    4       /* Test for read permission.  */
+#define W_OK    2       /* Test for write permission.  */
+//#define   X_OK    1       /* execute permission - unsupported in windows*/
+#define F_OK    0       /* Test for existence.  */
+
+#define access _access
+#define dup2 _dup2
+#define execve _execve
+#define ftruncate _chsize
+#define unlink _unlink
+#define fileno _fileno
+#define getcwd _getcwd
+#define chdir _chdir
+#define isatty _isatty
+#define lseek _lseek
+/* read, write, and close are NOT being #defined here, because while there are file handle specific versions for Windows, they probably don't work for sockets. You need to look at your app and consider whether to call e.g. closesocket(). */
+
+#ifdef _WIN64
+#define ssize_t __int64
+#else
+#define ssize_t long
+#endif
+
+#define STDIN_FILENO 0
+#define STDOUT_FILENO 1
+#define STDERR_FILENO 2
+/* should be in some equivalent to <sys/types.h> */
+//typedef __int8            int8_t;
+typedef __int16           int16_t; 
+typedef __int32           int32_t;
+typedef __int64           int64_t;
+typedef unsigned __int8   uint8_t;
+typedef unsigned __int16  uint16_t;
+typedef unsigned __int32  uint32_t;
+typedef unsigned __int64  uint64_t;
+
+#endif /* unistd.h  */
+


matthewdouglas:
Hi,

Can you please clarify what it is that is being fixed for Windows? What parts of unistd.h are actually needed? I assume this is for AMD GPUs on Windows, is that right? Preferably dependencies like #1834 and #1835 would be included in the same PR; but ideally I would not want to add any more code than necessary.

asmertpc-cloud:
Hi
Sorry, I'm not quite comfortable with Git yet.
You're right, it's for AMD GPUs running Windows.
I closed #1834 and #1835, but calling unistd.h needs to be done like this.

`
#ifndef` _UNISTD_H
#define _UNISTD_H`

/* Checking that we're on Windows */
#if defined(_WIN32) || defined(_WIN64)

#include <io.h>
#include <process.h>
#include <direct.h>
#include <stdlib.h>

/* Access Constants (POSIX -> Windows) */
#define F_OK 0
#define X_OK 1 /* В Windows реально нет проверки на исполнение через access */
#define W_OK 2
#define R_OK 4

/* Standard Descriptors */
#define STDIN_FILENO  0
#define STDOUT_FILENO 1
#define STDERR_FILENO 2

/* 
   Macros for functions.
   We use __host__ for HIP compatibility,
   so the compiler knows where these functions are available.
*/
#ifdef __cplusplus
extern "C" {
#endif

#define access _access
#define chdir  _chdir
#define close  _close
#define dup    _dup
#define dup2   _dup2
#define getpid _getpid
#define isatty _isatty
#define lseek  _lseek
#define read   _read
#define rmdir  _rmdir
#define unlink _unlink
#define write  _write

#ifdef _MSC_VER
  /* MSVC doesn't have a sleep function; we use the Windows API. */
  #ifndef NOMINMAX
    #define NOMINMAX
  #endif
  #include <windows.h>
  #define sleep(x) Sleep((x) * 1000)
  #define usleep(x) Sleep((x) / 1000)
#endif

#ifdef __cplusplus
}
#endif

#else
/* If we're not on Windows, we just pass in the real unistd.h. */
#include <unistd.h>
#endif

#endif /* _UNISTD_H */

Jason-123-cyber (approved):
Ty

rwfsmith:
Is it even using anything in unistd.h? I created a PR to build for AMD on Windows and I just have it ignore that include, which builds and runs successfully.

TimDettmers (changes requested):
## PR Review: #1836 — Create unistd.h

Adds a `csrc/unistd.h` shim file intended to provide POSIX compatibility for Windows builds (AMD/ROCm on Windows). The file is copied verbatim from a Stack Overflow answer and carries a **CC BY-SA 4.0 license**, which is incompatible with this project's MIT license.

**Verdict: Request changes (3 blocking issues)**

### Blocking issues

**1. License incompatibility (CC BY-SA 4.0 vs MIT) — BLOCKING**

The file header explicitly states:
```
// Source - https://stackoverflow.com/a
// Posted by AShelly, modified by community.
// Retrieved 2025-12-21, License - CC BY-SA 4.0
```

Stack Overflow content posted after 2018-05-02 is licensed under CC BY-SA 4.0. CC BY-SA is a copyleft license that requires derivative works to be distributed under the same (or compatible) license. MIT is not compatible with CC BY-SA 4.0 — merging this file would create a license conflict for the entire bitsandbytes distribution. This is a hard blocker; the file cannot be merged in its current form regardless of any other considerations.

**2. Unnecessary approach — superseded by PR #1846**

PR #1846 ("Adding support for building for AMD on Windows") solves the same `unistd.h` problem with a targeted `#ifdef _WIN32` guard in `csrc/ops_hip.cuh`:
```c
#ifdef _WIN32
#include <io.h>
#include <process.h>
#include <windows.h>
#else
#include <unistd.h>
#endif
```

This approach is strictly superior because:
- It only includes what is actually needed (no `getopt.h`, no `ssize_t` typedef, no `int16_t`/`uint32_t` typedefs that conflict with `<cstdint>`).
- It avoids adding a new file with external licensing concerns.
- It is the standard pattern for platform-conditional includes in C/C++ codebases.
- Maintainer @matthewdouglas and commenter @rwfsmith both questioned whether the full shim is needed; rwfsmith confirmed that simply guarding the include builds and runs successfully.

PR #1843 takes the same guard approach. Both are better solutions than adding a 61-line compatibility header.

**3. Unsafe macro definitions and type redefinitions**

Even setting aside the license issue, the shim's content is problematic for this codebase:

- **`#include <getopt.h>`** — `getopt.h` is not a standard Windows header and is not used anywhere in bitsandbytes. This would cause a compilation failure on most Windows toolchains.
- **`typedef __int16 int16_t;` / `typedef __int32 int32_t;` etc.** — These types are already defined by `<cstdint>` (included on line 11 of `ops_hip.cuh`). The redefinitions would cause compiler errors or undefined behavior from duplicate typedefs.
- **`#define ssize_t __int64` / `#define ssize_t long`** — Using `#define` for a type alias is fragile and can break in template contexts. Not needed: `ssize_t` is not used in the bitsandbytes CUDA/HIP source.
- **`#define srandom srand` / `#define random rand`** — Not used in the codebase. Unnecessary macro pollution.
- **`#define execve _execve` / `#define ftruncate _chsize` / `#define fileno _fileno` / `#define getcwd _getcwd`** — None of these POSIX functions are called in bitsandbytes. These macros add risk (name collisions with standard library or third-party headers) for no benefit.
- **No `_WIN32` guard** — Unlike the author's revised version posted in the PR comments, the actual committed file has no `#if defined(_WIN32)` guard. If this header were included on Linux, it would shadow the real `<unistd.h>` and break the build.

### Security assessment

- **No malicious patterns detected.** The file contains only preprocessor directives, typedefs, and macro definitions. No executable code, no network access, no command execution.
- **Macro namespace pollution is a concern** but is a correctness issue rather than a security issue. The macros (`#define access _access`, `#define write _write`, etc.) could cause surprising behavior if this header leaks into other translation units, but this is a build/correctness risk, not a security exploit.
- **No invisible Unicode characters** detected in the diff.

### Recommendation

Close this PR in favor of #1846, which provides a comprehensive Windows+AMD build solution (CMakeLists.txt changes, GPU detection, and the `unistd.h` guard) without any licensing issues. If the `#ifdef` approach in #1846 is insufficient for some specific compilation scenario, the right path forward is to write a minimal MIT-licensed shim from scratch that includes only the headers actually needed by `ops_hip.cuh` (which appears to be none of the POSIX functions—`unistd.h` is simply not needed on Windows).

- Security: Clear (no malicious patterns)
- Downstream impact: None (build-system-only change, no Python API affected)
- Tests: N/A (no tests added or needed for a header shim, but the shim itself would fail to compile due to `getopt.h` and duplicate typedefs)
- CI: Not triggered (no checks reported on this branch)
- Cross-PR conflicts: Directly superseded by #1846 and #1843, which both solve the same problem with `#ifdef` guards
- License: **BLOCKING — CC BY-SA 4.0 content cannot be merged into an MIT-licensed project**

matthewdouglas:
Thank you for the contribution! Closing in favor of #1846.

PR #1838, opened 2025-12-31
Refactor: Use logging module for diagnostics and warnings instead of print statements

Harshvardhan-To1:
Replace all print() statements with logging calls and add a NullHandler to the top-level logger to allow downstream users to manage and filter library logs.

This change prevents bitsandbytes from writing directly to stdout/stderr from library code paths, making its logging opt-in. The CLI entrypoint (python -m bitsandbytes) now configures logging only for its run, respecting the BNB_LOG_LEVEL environment variable.

With respect to https://github.com/bitsandbytes-foundation/bitsandbytes/issues/919

### DIFF
--- a/bitsandbytes/__init__.py
+++ b/bitsandbytes/__init__.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import logging
 import importlib
 import sys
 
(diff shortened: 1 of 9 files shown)


bitsandbytes/__init__.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 
+import logging
```
TimDettmers: Import ordering: `import logging` should come after `import importlib` (alphabetical stdlib ordering). ruff/isort will flag this.

matthewdouglas:
Hi,

Thank you for the PR. We've decided to merge #1883 for this instead.

PR #1839, opened 2026-01-09
Fix xpu 4bit kernel

jiqing-feng:
Fix XPU 4-bit kernel for odd-sized weights and resolve related test failures.
The original XPU 4-bit kernel did not handle odd weight shapes correctly. This PR fixes the logic and resolves failures in tests such as test_4bit_quant_large.

For `pytest -k "xpu" -ra ./`:
```
2495 passed, 1335 skipped, 3998 deselected, 24 xfailed, 46 warnings
```

### DIFF
--- a/bitsandbytes/backends/xpu/ops.py
+++ b/bitsandbytes/backends/xpu/ops.py
@@ -38,7 +38,7 @@ def _dequantize_4bit_impl(
         get_ptr(absmax),
         get_ptr(out),
         ct.c_int(blocksize),
-        ct.c_int64(out.numel()),
+        ct.c_int(out.numel()),
         _get_tensor_stream(A),
     )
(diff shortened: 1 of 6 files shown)


jiqing-feng:

matthewdouglas:
Hi,

Because this moves to int64 for indexing, I assume direct uses like this one in Unsloth would need to be updated?

https://github.com/unslothai/unsloth/blob/ec1757c1a02175851146ff5f6ab2a26c8c863fc8/unsloth/kernels/utils.py#L438-L452

jiqing-feng:
Hi @matthewdouglas . I have updated the kernels so they can cast the dtype inside the kernel without api change. Please review the new changes. Thanks!

matthewdouglas (approved):

Issue #89, opened 2022-11-05
8-bit optimizers dont work with FSDP


prajdabre:
When I use an 8-bit ADAM with FSDP, I get an error as follows:

`RuntimeError: output tensor must have the same type as input tensor`

If my understanding is correct, there seems to be a casting issue. Is there any workaround this?

TIA.

TimDettmers:
I looked at the deepspeed implementation before, which had a similar issue with shared weights. The problem was that the algorithm splits all tensors found in the optimizer state, which includes the quantization statistics. But this can lead to incorrect behavior. The workaround in deepspeed is to hide the quantization statistics by obscuring their type (putting the tensor into a list/tuple).

I am not sure if the error message that you provided is related to that or not.

It would be nice if we could get 8-bit Adam working for FSDP. Would you be able to provide a simple example for debugging and replication purposes? Since I will be pretty busy the next month, I would also be very happy to guide you on how to fix this if you create a PR and provide me with error messages / stack traces. I think it would be pretty useful since more and more people are using FSDP.

philschmid:
Hey @TimDettmers, 

I created a gist with an example. The gist includes a `process_dataset.py` to prepare a dataset and a `run_clm_bnb8.py` script, which uses the `adamw_bnb_8bit` optimizer and `FSDP`. 

https://gist.github.com/philschmid/99410e8bf66d34e52bb0cd5270b07989

I hope that's enough for you to test it.

philschmid:
I tested the example i shared with the `adamw_bnb_8bit` and `adafactor` it seems that its not working if the training runs. 

AdamWInt8
```bash
{'loss': 2.6643, 'learning_rate': 4.847094801223242e-05, 'epoch': 0.09}
{'loss': 2.752, 'learning_rate': 4.694189602446483e-05, 'epoch': 0.18}
{'loss': 3.1493, 'learning_rate': 4.541284403669725e-05, 'epoch': 0.28}
{'loss': 3.412, 'learning_rate': 4.3883792048929664e-05, 'epoch': 0.37}
{'loss': 3.6722, 'learning_rate': 4.0825688073394495e-05, 'epoch': 0.55}
```

Adafactor
```bash
{'loss': 2.8385, 'learning_rate': 4.847094801223242e-05, 'epoch': 0.09}   
{'loss': 2.6384, 'learning_rate': 4.694189602446483e-05, 'epoch': 0.18}                   
[...]

PR #1840, opened 2026-01-11
bug: fix 8bitoptim support with fsdp

ved1beta:
#89 

The state_dict() method in Optimizer8bit wraps quantization states (like state1, state2, absmax1, etc.) in a nested dict with key __bnb_optimizer_quant_state__
This allows FSDP's full_optim_state_dict to gather optimizer states across ranks without trying to gather the quantization tensors which have different shapes than model parameters
The load_state_dict() method automatically unwraps these states when loading

tested with
```torchrun --nproc_per_node=2 test.py```

```
"""
import os
import torch
import torch.distributed as dist
import torch.nn as nn
from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
from torch.distributed.fsdp import FullOptimStateDictConfig, StateDictType
import bitsandbytes as bnb

# Initialize distributed
# For single GPU testing, set environment variables if not already set
if "RANK" not in os.environ:
    os.environ["RANK"] = "0"
    os.environ["LOCAL_RANK"] = "0"
    os.environ["WORLD_SIZE"] = "1"
    os.environ["MASTER_ADDR"] = "localhost"
    os.environ["MASTER_PORT"] = "12355"

dist.init_process_group("nccl")

rank = dist.get_rank()
local_rank = int(os.environ["LOCAL_RANK"])
device = torch.device(f"cuda:{local_rank}")
torch.cuda.set_device(device)

# Create model and wrap with FSDP
class YourModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(128, 64)
    
    def forward(self, x):
        return self.linear(x)

model = YourModel().to(device)
model = FSDP(model)

# Use 8-bit optimizer
optimizer = bnb.optim.Adam8bit(model.parameters(), lr=1e-4)

# Training loop
model.train()
for step in range(3):
    x = torch.randn(2, 128, device=device)
    y = model(x).sum()
    y.backward()
    optimizer.step()
    optimizer.zero_grad()
    if rank == 0:
        print(f"Step {step} completed")

# This call now works!
# The optimizer's state_dict() method automatically wraps quantization states
# in a nested dict that FSDP can handle during full_optim_state_dict gathering
with FSDP.state_dict_type(
    model,
    StateDictType.FULL_STATE_DICT,
    optim_state_dict_config=FullOptimStateDictConfig(offload_to_cpu=True, rank0_only=True),
):
    optim_state = FSDP.full_optim_state_dict(model, optimizer)
    
    if rank == 0:
        print("\n✓ Successfully retrieved full optimizer state dict!")
        print(f"State dict contains {len(optim_state.get('state', {}))} parameter states")
        
        # Verify quantization states are wrapped (for debugging)
        if optim_state.get("state"):
            first_param_state = next(iter(optim_state["state"].values()))
            wrapped_key = bnb.optim.optimizer.Optimizer8bit._FSDP_WRAPPED_QUANT_STATE_KEY
            if wrapped_key in first_param_state:
                print(f"✓ Quantization states are properly wrapped under '{wrapped_key}'")
                print(f"  Wrapped keys: {list(first_param_state[wrapped_key].keys())}")

dist.destroy_process_group()



```

### DIFF
--- a/bitsandbytes/optim/optimizer.py
+++ b/bitsandbytes/optim/optimizer.py
@@ -111,6 +111,8 @@ class GlobalOptimManager:
 
 
 class Optimizer8bit(torch.optim.Optimizer):
+    _FSDP_WRAPPED_QUANT_STATE_KEY = "__bnb_optimizer_quant_state__"
+
     def __init__(self, params, defaults, optim_bits=32, is_paged=False):
         """
         Base 8-bit optimizer class.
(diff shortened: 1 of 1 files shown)


matthewdouglas:
LGTM, thanks!

PR #1841, opened 2026-01-14
Renovate pre-commit + CI a bit

akx:
Happy 2026!

Just some minor-ish CI pipeline and lint fixes (for now 😇):

* upgrade pre-commit tools, fix typos (new version of crate-ci/typos)
* use `uv` to install Python and tools (faster!)
* use `pre-commit-uv` (faster!) ([`prek`](https://github.com/j178/prek/) is also an option)
* upgrade CI action versions and pin them to SHAs (as is security best practice)

### DIFF
--- a/.github/scripts/build-rocm.sh
+++ b/.github/scripts/build-rocm.sh
@@ -18,7 +18,7 @@ if [ "${build_os:0:6}" == ubuntu ]; then
     docker run --rm --platform "linux/$build_arch" -i \
         -w /src -v "$PWD:/src" "$image" sh -c \
         "apt-get update \
-      && pip install cmake==3.31.6 \
+      && uv tool install cmake==3.31.6 \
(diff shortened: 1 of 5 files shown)


.github/scripts/build-rocm.sh
```diff
@@ -18,7 +18,7 @@ if [ "${build_os:0:6}" == ubuntu ]; then
     docker run --rm --platform "linux/$build_arch" -i \
         -w /src -v "$PWD:/src" "$image" sh -c \
         "apt-get update \
-      && pip install cmake==3.31.6 \
+      && uv tool install cmake==3.31.6 \
```
matthewdouglas: The ROCm docker images likely won't have `uv` installed, and that's not covered by a setup-uv action running on the host. I recognize the advantage of using uv but in this case it's probably not worth changing since installing cmake is < 1% of the build runtime.

matthewdouglas:
@akx Thanks! These are welcome changes! Left some comments, but will be happy to merge when it's ready!

PR #1843, opened 2026-01-20
fix ROCm GPU architecture detection failed on windows

guinmoon:
On Windows, we use `hipinfo` instead of `rocminfo`.

```log
ROCm GPU architecture detection failed despite ROCm being available.

Could not detect ROCm warp size: [WinError 2] The system cannot find the file specified. Defaulting to 64. (some 4-bit functions may not work!)

ROCm warp size detection failed despite ROCm being available.
```


matthewdouglas:
Thank you for the contribution! We've decided to instead merge #1846 which should address the same issues.

PR #1855, opened 2026-02-06
Add LARS to str2optimizer32bit dictionary

### LINKED ISSUE #1810: LARS missing in str2optimizer32bit
### System Info

I initially posted this as a [discussion forum question](https://github.com/bitsandbytes-foundation/bitsandbytes/discussions/1801), but I think it is more appropriate as an issue.

I am using bitsandbytes (version 0.48.2) and I am trying to run use LARS in my torch code. Strangely I can get other bitsandbytes optimizers including LAMB running, but LARS wont run.

Error message:
> <PATH2PROJECT>.venv/lib/python3.12/site-packages/bitsandbytes/backends/cuda/ops.py", line 635, in optimizer_update_32bit_impl
>     raise ValueError(
> ValueError: Unsupported optimizer name: lars. Supported optimizers: ['adam', 'momentum', 'rmsprop', 'lion', 'adagrad', 'ademamix']
> 

Looking at `bitsandbytes/backends/cuda/ops.py` it seems to be because
`optim_fns = str2optimizer32bit.get(optimizer_name, None)` has value `None`, because `str2optimizer32bit` does not have a case for lars. Here is a [link to the relevant lines?](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/63f538a4e492017dd4a52edfde142c4ca818d401/bitsandbytes/backends/cuda/ops.py#L543-L577). The error message is also incorrect as it lists the supported 8bit optimizers instead of the supported 32bit optimizers. 

**TLDR:**
- `"lars"` is not handled by `str2optimizer32bit `in `bitsandbytes/backends/cuda/ops.py`?
- The error message displays the keys of `str2optimizer8bit_blockwise `instead of the keys of `str2optimizer32bit `as the list of supported optimizers.

**Sidenote**: It looks like LAMB maps to cadam kernels. Is this because the same kernels can be used for both methods?

**Related issues:**
- https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1118
- https://github.com/Nerogar/OneTrainer/issues/390

### Reproduction
```
import torch
import torch.nn as nn
import bitsandbytes as bnb

device = "cuda"

# Dummy model
class SimpleModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.lin = nn.Linear(10, 2)

    def forward(self, x):
        return self.lin(x)

model = SimpleModel().to(device)

# Fake batch on GPU
x = torch.randn(4, 10, device=device)
y = torch.randint(0, 2, (4,), device=device)

criterion = nn.CrossEntropyLoss().to(device)

# bitsandbytes LARS optimizer on GPU
optimizer = bnb.optim.LARS(
    model.parameters(),
    lr=1e-3,
    momentum=0.9,
    weight_decay=1e-4,
)

# Training step
optimizer.zero_grad()
pred = model(x)
loss = criterion(pred, y)
loss.backward()
optimizer.step()
```

### Expected behavior

Running the script will lead to a ValueError.   

>   File "<PATH2PROJECT>/.venv/lib/python3.12/site-packages/bitsandbytes/backends/cuda/ops.py", line 635, in _optimizer_update_32bit_impl
>     raise ValueError(
> ValueError: Unsupported optimizer name: lars. Supported optimizers: ['adam', 'momentum', 'rmsprop', 'lion', 'adagrad', 'ademamix']

Mr-Neutr0n:
## Summary
- Adds "lars" entry to `str2optimizer32bit` dictionary mapping to momentum kernels
- LARS uses momentum-based kernels since it's essentially SGD with momentum plus layerwise adaptive learning rates
- LARS8bit and LARS32bit classes set `block_wise=False`, so no changes needed to `str2optimizer8bit_blockwise`

## Related Issue
Fixes #1810

## Test plan
- [ ] Verify LARS32bit optimizer no longer throws KeyError
- [ ] Run existing LARS optimizer tests

### DIFF
--- a/bitsandbytes/backends/cuda/ops.py
+++ b/bitsandbytes/backends/cuda/ops.py
@@ -637,7 +637,7 @@ def _optimizer_update_32bit_impl(
     optim_fns = str2optimizer32bit.get(optimizer_name, None)
     if optim_fns is None:
         raise ValueError(
-            f"Unsupported optimizer name: {optimizer_name}. Supported optimizers: {list(str2optimizer8bit_blockwise.keys())}"
+            f"Unsupported optimizer name: {optimizer_name}. Supported optimizers: {list(str2optimizer32bit.keys())}"
         )
     if g.dtype == torch.float32:
         optim_func = optim_fns[0]
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -102,6 +102,11 @@ str2optimizers["momentum8bit_blockwise"] = (
     lambda pxx: bnb.optim.SGD8bit(pxx, 0.01, 0.9, block_wise=True),
 )
 
+str2optimizers["lars"] = (
+    lambda pxx: bnb.optim.PytorchLARS(pxx, 0.01, 0.9),
+    lambda pxx: bnb.optim.LARS(pxx, 0.01, 0.9),
+)
+
 str2optimizers["rmsprop"] = (
     lambda pxx: torch.optim.RMSprop(pxx, 0.01, 0.9),
     lambda pxx: bnb.optim.RMSprop(pxx, 0.01, 0.9, block_wise=False),
@@ -118,6 +123,7 @@ str2statenames["paged_adam"] = [("exp_avg", "state1"), ("exp_avg_sq", "state2")]
 str2statenames["lion"] = [("exp_avg", "state1")]
 str2statenames["paged_lion"] = [("exp_avg", "state1")]
 str2statenames["momentum"] = [("momentum_buffer", "state1")]
+str2statenames["lars"] = [("momentum_buffer", "state1")]
 str2statenames["lamb"] = [("exp_avg", "state1"), ("exp_avg_sq", "state2")]
 str2statenames["rmsprop"] = [("square_avg", "state1")]
 
@@ -155,6 +161,7 @@ optimizer_names_32bit = [
     "paged_adamw",
     "paged_adam",
     "momentum",
+    "lars",
     "rmsprop",
     "lion",
     "paged_lion",
@@ -181,7 +188,7 @@ def test_optimizer32bit(dim1, dim2, gtype, optim_name, device):
     if optim_name.startswith("paged_") and device == "xpu":
         pytest.skip("Paged optimizers are not supported on XPU currently.")
 
-    if gtype == torch.bfloat16 and optim_name in ["momentum", "rmsprop"]:
+    if gtype == torch.bfloat16 and optim_name in ["momentum", "lars", "rmsprop"]:
         pytest.skip()
     if dim1 == 1 and dim2 == 1:
         return


Mr-Neutr0n:
hey, just wanted to check if this is something you'd be interested in merging. adds LARS to the optimizer mapping so it doesn't error out. small addition

Mr-Neutr0n:
Friendly bump! Let me know if there's anything I should update or improve to help move this forward.

TimDettmers:
Pushed two additional changes on top of the original LARS dict entry fix:

1. **Fixed wrong error message** in `_optimizer_update_32bit_impl` — was displaying `str2optimizer8bit_blockwise.keys()` instead of `str2optimizer32bit.keys()` (the second bug reported in #1810).
2. **Added LARS to the parametrized 32-bit optimizer tests** — uses `PytorchLARS` as the reference implementation, with bf16 skip since momentum kernels lack a bf16 variant (matching the existing `momentum`/`rmsprop` skip).

Also rebased onto current `main`. Tests: 8 passed, 4 bf16 skipped.

TimDettmers (commented):
## PR Review: #1855 — Add LARS to str2optimizer32bit dictionary

Bug fix: adds the missing `"lars"` entry to the `str2optimizer32bit` dict in the CUDA backend, fixing `KeyError` when using `LARS32bit`. Also fixes an incorrect error message in `_optimizer_update_32bit_impl` that was displaying `str2optimizer8bit_blockwise` keys instead of `str2optimizer32bit` keys (second bug from #1810). Adds test coverage for LARS in the 32-bit optimizer parametrized tests.

**No blocking issues.**

The fix is correct:
- All three LARS classes (`LARS`, `LARS8bit`, `LARS32bit`) pass `optimizer_name="lars"` and `block_wise=False` to `Optimizer1State.__init__`, so only `str2optimizer32bit` (for 32-bit state) and `str2optimizer8bit` in `functional.py` (for non-blockwise 8-bit state) are needed. The `str2optimizer8bit` dict in `functional.py` already had a `"lars"` entry (mapping to momentum kernels), so only the `str2optimizer32bit` dict was missing it.
- The kernel mapping to `cmomentum32bit_grad_32` / `cmomentum32bit_grad_16` is correct — LARS uses momentum-based updates (the layerwise adaptive learning rate scaling is handled in the Python optimizer layer, not the kernel).
- The 2-tuple (fp32 + fp16, no bf16) matches the existing `"momentum"` entry. The dispatch code correctly handles the missing bf16 variant by checking `len(optim_fns) == 3`.
- The bf16 skip in `test_optimizer32bit` correctly adds `"lars"` alongside `"momentum"` and `"rmsprop"`.
- The test uses `PytorchLARS` as the reference implementation, which is the existing pure-PyTorch reference class in `bitsandbytes/optim/lars.py`.
- The error message fix (line 640: `str2optimizer8bit_blockwise.keys()` → `str2optimizer32bit.keys()`) is a clear correctness fix — the function is validating against `str2optimizer32bit`, so the error should display those keys.

- Security: Clear (no network access, no command execution, no new imports, no obfuscation)
- Downstream impact: None (additive fix — restores intended functionality, does not change any public API)
- Tests: Adequate (LARS added to parametrized 32-bit tests with bf16 skip matching existing patterns)
- CI: Not triggered (fork PR — maintainer needs to approve workflow run before merge)
- Cross-PR conflicts: PR #1871 (deprecation removal) touches `tests/test_optim.py` and `bitsandbytes/functional.py`. It removes the non-blockwise `str2optimizer8bit` dict and `percentile_clipping` from LARS. No direct file overlap with this PR’s `backends/cuda/ops.py` changes, but the `tests/test_optim.py` changes will need a straightforward merge. Recommend merging this PR first since it is smaller and fixes a reported bug.

PR #1859, opened 2026-02-14
Handle non-contiguous tensors in quantize/dequantize ops

### LINKED ISSUE #1342: quantize_4bit/dequantize_4bit gives wrong output on in-contiguous tensor
### System Info

Linux

### Reproduction

I hit this bug during my coding work. I verified it with the following script. With bitsandbytes 0.43.3, it shows that half of the elements from the quantize/dequantize from the in-contiguous tensor are different from the counterpart of contiguous tensor. 

```
import torch
from bitsandbytes.functional import quantize_4bit, dequantize_4bit

# Create a contiguous tensor
original_tensor = torch.randn(3, 4, 6).to('cuda')


# Create a non-contiguous tensor by slicing (skipping elements)
non_contiguous_tensor = original_tensor[:, ::2, :]
compressed_non_contiguous_tensor, quant_state = quantize_4bit(non_contiguous_tensor)
uncompressed_non_contiguous_tensor = dequantize_4bit(compressed_non_contiguous_tensor, quant_state)

# Create a contiguous tensor by calling contiguous()
contiguous_tensor = non_contiguous_tensor.contiguous()
compressed_contiguous_tensor, quant_state = quantize_4bit(contiguous_tensor)
uncompressed_contiguous_tensor = dequantize_4bit(compressed_contiguous_tensor, quant_state)

# it is expected to be the same, but half of the elements are different
num_different_elements = (uncompressed_contiguous_tensor != uncompressed_non_contiguous_tensor).sum().item()
print(f"The number of different elements is: {num_different_elements}")
```

### Expected behavior

Incontinguous tensors are expected to give the same quant/dequant results with the contiguous one, as long as the elements are the same.

### LINKED ISSUE #1690: `bitsandbytes.functional.quantize_4bit`, `bitsandbytes.functional.quantize_blockwise` require contiguous input, fail silently if not
### System Info

Python 3.12.3, on an AWS P4 instance.

### Reproduction

Say I call `quantize_4bit` with `x` of shape `(a, b, c, blocksize)`. Call `num_ch = a * b * c`. I am getting `qx`, `state`, so that:

* `state.absmax.shape = (num_ch,)`
* `qx.shape = (num_ch * blocksize // 2,)`, `qx.dtype = torch.uint8`

My best guess was the memory layout of `qx` and `state.absmax` allows me to do `qx.view(a, b, c, blocksize // 2)` and `state.absmax.view(a, b, c)` and then work with that. But this does not seem to work.

```python
# x.shape = (4, 3, 38, 256)
qx, state = quantize_4bit(x, blocksize=256)
absmax_cmp = x.view(-1, 256).abs().max(dim=-1)
torch.testing.assert_close(state.absmax, absmax_cmp)

start, end = 10, 15
partx = x[:, :, start:end, :]
qx_part, state_part = quantize_4bit(partx, blocksize=256)
full = state.absmax.view(4, 3, 38)
part = state_part.absmax(4, 3, 5)
torch.testing.assert_close(full[:, :, start:end], part)

full = qx.view(4, 3, 38, -1)
part = qx_part.view(4, 3, 5, -1)
torch.testing.assert_close(full[:, :, start:end], part)
```

As it stands, this fails with the 2nd assert. But if I replace:

```
partx = x[:, :, start:end, :].contiguous()
```

Then, it all works. This means `quantize_4bit` does not work properly if its input is not contiguous. But it also does not check that, so things just fail silently.

### Expected behavior

I'd expect `quantize_4bit` and also `quantize_blockwise` either to check whether the input is contiguous, or to convert it to contiguous before quantization.

If I find some time, I submit a PR, which should also include a test that the example above works.

TimDettmers:
## Summary

- Auto-convert non-contiguous input tensors to contiguous in the CUDA backend for `quantize_blockwise`, `dequantize_blockwise`, `quantize_4bit`, and `dequantize_4bit`
- The CUDA kernels use raw pointers (`get_ptr()`) which assume contiguous memory layout; non-contiguous inputs (e.g. strided slices) produced silently incorrect results
- Added regression tests verifying non-contiguous tensors produce identical results to their contiguous equivalents

Fixes #1342
Fixes #1690

## Test plan

- [x] New `TestNonContiguousInputs` class in `tests/test_ops.py` with 4 test methods (54 CUDA parametrizations)
- [x] Tests create non-contiguous tensors via strided slicing and compare quantize/dequantize output against contiguous baseline
- [x] Includes end-to-end roundtrip test (quantize → dequantize) for 4bit
- [x] All existing tests pass (371 passed, 73 skipped, 60 xfailed; 6 pre-existing OOM failures in `test_4bit_quant_large` unrelated to this change)

### DIFF
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -276,9 +276,7 @@ def test_quant_storage_shard_roundtrip(device, quant_type, quant_storage):
     reassembled = torch.cat(shards).reshape(qB.shape)
 
     assert reassembled.dtype == qB.dtype
-    assert torch.equal(
-        reassembled.view(torch.uint8), qB.view(torch.uint8)
-    ), "Bytes changed after shard roundtrip"
+    assert torch.equal(reassembled.view(torch.uint8), qB.view(torch.uint8)), "Bytes changed after shard roundtrip"
 
     out = bnb.functional.gemv_4bit(A, reassembled.t(), state=state)
     torch.testing.assert_close(out, ref)


matthewdouglas (approved):
LGTM!

PR #1659, opened 2025-05-28
# Fix for matmul_4bit out Parameter Issue

ved1beta:
fix: matmul_4bit out parameter handling for 1D and 2D tensors

- Fix incorrect handling of out parameter in matmul_4bit for 1D tensors
- Improve shape validation in gemv_4bit.out to handle 1D tensor cases
- Ensure proper tensor reshaping and memory handling in CUDA kernels
- Update conditional logic in matmul_4bit to consistently handle OUT parameter

This fixes GitHub issue #1235 where the out parameter was not working correctly
for matrix multiplication operations, particularly with 1D input tensors.

PR #1860, opened 2026-02-14
Fix matmul_4bit out parameter not writing to output tensor

### LINKED ISSUE #1235: out kwarg in matmul_4bit() is not working
### System Info

Linux 20.04

### Reproduction

output is defined as a tensor. The following works as expected:
`output = matmul_4bit(a, b)`

but the following does not, the elements in output are not changed. 
`matmul_4bit(a, b, out=output)`

### Expected behavior

`matmul_4bit(a, b, out=output)` is expected to have the output values set. It is a preferred way as it saves extra memory copy.

TimDettmers:
## Summary

Fixes #1235

The `out` kwarg in `matmul_4bit()` was accepted but silently ignored in the `MatMul4Bit.forward()` code path (2D+ inputs). Calling `matmul_4bit(a, b, out=output)` would return the correct result but never write it into the `output` tensor.

**Fix:** After computing the result via `torch.nn.functional.linear()`, copy it into `out` if provided. This is a 3-line change in `MatMul4Bit.forward()`.

**Note:** The 1D (gemv) path already handled `out` correctly via `F.gemv_4bit()`. Only the matrix multiplication path was broken.

**Related:** PR #1659 proposed a similar fix. This PR takes a simpler approach — no `resize_()` call (which could surprise callers expecting a fixed-size buffer), and no special-casing for 1D inputs in `matmul_4bit()` since the gemv path already works.

## Test plan

- [x] Added `test_matmul_4bit_out_parameter` covering both 2D (matmul) and 1D (gemv) paths
- [x] Tests pass for nf4/fp4, fp16/fp32, with and without bias, on CPU and CUDA
- [x] Verified `out` tensor contains correct values and returned tensor shares storage with `out`
- [ ] Full test suite (to be run separately)

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -322,7 +322,12 @@ class MatMul4Bit(torch.autograd.Function):
             out.copy_(output)
             output = out
 
-        # 3. Save state
+        # 3. Write to out tensor if provided
+        if out is not None:
+            out.copy_(output)
+            output = out
+
+        # 4. Save state
         ctx.state = quant_state
         ctx.dtype_A, ctx.dtype_B, ctx.dtype_bias = A.dtype, B.dtype, None if bias is None else bias.dtype
 
--- a/tests/test_autograd.py
+++ b/tests/test_autograd.py
@@ -258,3 +258,45 @@ def test_matmul_4bit(
 
                 if req_grad[2]:
                     torch.testing.assert_close(gradBias1, gradBias2)
+
+
+@pytest.mark.parametrize("device", get_available_devices())
+@pytest.mark.parametrize("quant_type", ["nf4", "fp4"], ids=id_formatter("quant_type"))
+@pytest.mark.parametrize("dtype", [torch.float16, torch.float32], ids=describe_dtype)
+@pytest.mark.parametrize("has_bias", TRUE_FALSE, ids=id_formatter("has_bias"))
+def test_matmul_4bit_out_parameter(device, quant_type, dtype, has_bias):
+    """Test that matmul_4bit(A, B, out=output) writes the result into output (issue #1235)."""
+    M, K, N = 32, 64, 48
+
+    # Create weight matrix (K, N) and quantize — matmul_4bit computes A @ dequant(B)
+    W = torch.randn(K, N, device=device, dtype=dtype)
+    torch.nn.init.xavier_uniform_(W)
+    B_quant, quant_state = bnb.functional.quantize_4bit(W, quant_type=quant_type)
+
+    bias = None
+    if has_bias:
+        bias = torch.randn(N, device=device, dtype=dtype)
+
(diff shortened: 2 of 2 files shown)


TimDettmers (commented):
## PR Review: #1860 — Fix matmul_4bit out parameter not writing to output tensor

`[bug-fix]` `[test]` — Fixes #1235. The `out` kwarg in `matmul_4bit()` was accepted but silently ignored in the `MatMul4Bit.forward()` code path (2D+ inputs). The fix adds `out.copy_(output)` after computing the linear result, then reassigns `output = out` so the returned tensor is the caller's pre-allocated buffer. Clean, minimal, correct.

**No blocking issues.**

The fix follows the standard PyTorch `out` parameter convention: (1) copy the result into the provided tensor via `.copy_()`, (2) return that same tensor. The 1D gemv path already handled `out` correctly, so only the matrix multiplication path needed this fix.

The test covers both the 2D (matmul) and 1D (gemv) code paths, parametrized across nf4/fp4, fp16/fp32, with and without bias, and across available devices. It verifies both that the `out` tensor contains the correct values and that the returned tensor shares storage with `out` (via `data_ptr()` comparison).

**Cross-PR note:** PR #1659 (by ved1beta) targets the same issue with a more invasive approach: it adds `resize_()` on the `out` tensor (which could surprise callers expecting a fixed-size buffer), uncomments and re-enables type checks in `_gemv_4bit_impl`, and changes the 1D dispatch logic in `matmul_4bit()`. This PR's approach is strictly simpler — a 3-line change in `MatMul4Bit.forward()` with no behavioral changes to the gemv path (which already works). PR #1659 will conflict on `bitsandbytes/autograd/_functions.py` and should be closed or superseded once this merges.

- Security: Clear
- Downstream impact: LOW (bug fix makes behavior match documented API; no downstream project passes `out` to `matmul_4bit` — TGI and vLLM call it without `out`; vLLM comments reference #1235 as a known bug)
- Tests: Adequate (2D + 1D paths, nf4/fp4, fp16/fp32, bias/no-bias, CPU/CUDA)
- CI: All checks pass
- Performance: Negligible (`.copy_()` only runs when `out` is provided)
- Cross-PR conflicts: PR #1659 overlaps on `bitsandbytes/autograd/_functions.py` (same bug, different fix; recommend closing #1659 after this merges)
- Serialization: Not affected
- torch.compile: Not affected (no op registration changes)

matthewdouglas:
This makes sense, but the goal with issue #1235 was to save an extra memory copy. Since we're copying anyway I'm not sure we get the perf advantage that is being asked for, but at the very least the out kwarg API will behave as expected.

matthewdouglas:
The `out` parameter is now handled in `MatMul4Bit.forward()` as of #1949 (merged via the new `gemm_4bit` kernel PR). Specifically, lines 349-351 in `bitsandbytes/autograd/_functions.py` now do `out.copy_(output); output = out` when `out` is provided. This PR's fix is superseded — closing.

PR #1865, opened 2026-02-14
Fix: Linear8bitLt crash with tied weights (skip_modules)

### LINKED ISSUE #1634: Can't get llm_int8_skip_modules to work: 'Parameter' object has no attribute 'SCB'
I'm working with a RecurrentGemma model (alpindale/recurrentgemma-9b-it).
My desired result is quantizing the model to int8, but keeping the recurrent blocks (model.layers.X.temporal_block) unquantized.
Unfortunately setting `llm_int8_skip_modules=["temporal_block"]` doesn't work.
On loading the model, I get `AttributeError: 'Parameter' object has no attribute 'SCB'`.
You can reproduce it with this Colab: https://colab.research.google.com/drive/1NkgcXmuYJg0XqFWMuZA-ZwWDjgdsuCjr
I've also tried `llm_int8_skip_modules=["recurrent"]`, this yields the same error.
Also the library isn't broken; `llm_int8_skip_modules=["lm_head"]` works fine.
My initial thought is that I'm setting the option incorrectly, but I'm not even sure how to find the right setting.
Anyone have any thoughts on how to fix this?

TimDettmers:
## Summary
- Fixes `AttributeError: 'Parameter' object has no attribute 'SCB'` when using `llm_int8_skip_modules` without `lm_head` on models with tied weights (e.g. OPT)
- Adds `isinstance(self.weight, Int8Params)` guard in `Linear8bitLt.forward()` to fall back to `F.linear` when weight is a plain `Parameter` (e.g. due to weight tying with embedding layer)
- Makes `_save_to_state_dict` and `_load_from_state_dict` robust to non-`Int8Params` weights via `getattr` defaults

## Test plan
- [ ] New test `test_linear8bitlt_tied_weights_no_crash` passes on CPU and CUDA
- [ ] Existing `test_linear8bitlt.py` tests remain unaffected
- [ ] Manual verification with OPT-125m reproduction from issue

Fixes #1634

### DIFF
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -276,9 +276,7 @@ def test_quant_storage_shard_roundtrip(device, quant_type, quant_storage):
     reassembled = torch.cat(shards).reshape(qB.shape)
 
     assert reassembled.dtype == qB.dtype
-    assert torch.equal(
-        reassembled.view(torch.uint8), qB.view(torch.uint8)
-    ), "Bytes changed after shard roundtrip"
+    assert torch.equal(reassembled.view(torch.uint8), qB.view(torch.uint8)), "Bytes changed after shard roundtrip"
 
     out = bnb.functional.gemv_4bit(A, reassembled.t(), state=state)
     torch.testing.assert_close(out, ref)


TimDettmers (commented):
## PR Review: #1865 — Fix: Linear8bitLt crash with tied weights (skip_modules)

**Classification:** `[bug-fix]`

Fixes #1634 where `Linear8bitLt` raises `AttributeError: 'Parameter' object has no attribute 'SCB'` when `llm_int8_skip_modules` is used without including `lm_head` on models with tied weights (e.g. OPT, RecurrentGemma).

The fix adds an `isinstance(self.weight, Int8Params)` guard in `forward()` to fall back to `F.linear`, and makes `_save_to_state_dict` / `_load_from_state_dict` robust to non-`Int8Params` weights via `getattr` defaults.

**Blocking issue (1):**

1. **Regression: 64 existing `test_linear_serialization` tests fail.** The `isinstance(self.weight, Int8Params)` guard in `forward()` triggers incorrectly when a `Linear8bitLt` module is deserialized via `torch.save`/`torch.load` (pickle). After pickle round-trip, the weight loses its `Int8Params` subclass type and becomes a plain `nn.Parameter`. The guard then falls through to `F.linear(x, self.weight, self.bias)` which crashes with dtype mismatches (`Half vs Float` for `has_fp16_weights=False`, `Float vs Char` for `has_fp16_weights=True`). All 64 parameterizations of `test_linear_serialization` fail on CUDA. This is confirmed not pre-existing: nightly tests on `main` pass, and other recent PR branches (`fix/issue-1269`, `fix/issue-1810`, `deprecation`) also pass the same test suite.

   The root cause identification is correct (tied weights replace `Int8Params` with a plain `Parameter`), but the guard is too broad — it also catches legitimate deserialization paths where the parameter type is similarly lost.

   **Suggested fix:** Instead of checking `isinstance(self.weight, Int8Params)`, consider checking for the presence of the `CB` attribute: `if not hasattr(self.weight, 'CB')`. This would distinguish tied-weight Parameters (which never have `CB`) from deserialized Parameters (which do have `CB` via pickle). Alternatively, add an explicit attribute flag (e.g., `self._is_quantized_module = True`) that survives serialization, and check for attributes like `CB`/`SCB` rather than the class type.

**Additional observations:**

- The `_save_to_state_dict` and `_load_from_state_dict` changes (using `getattr(..., None)`) are correct and necessary defensive improvements.
- The new test `test_linear8bitlt_tied_weights_no_crash` is well-structured and covers the exact scenario from issue #1634. It does not test the `state_dict()` round-trip for the tied-weight case, but that is a minor gap.
- The `test_linear4bit.py` change is purely cosmetic (reformatting an assertion to one line).
- The `coordinator_guide.md` change is unrelated to the bug fix (updates agent workflow instructions). Benign but ideally would be in a separate commit.

## Downstream Impact

Risk level: LOW (once the regression is fixed)

Affected APIs:
- `Linear8bitLt.forward()` — adds graceful fallback for non-Int8Params weights
- `Linear8bitLt._save_to_state_dict()` — uses getattr defaults (no behavior change for normal case)
- `Linear8bitLt._load_from_state_dict()` — uses getattr defaults (no behavior change for normal case)

Affected projects:
- Transformers: Positive impact — this fixes the `llm_int8_skip_modules` crash users report. No API change.
- PEFT: Not affected
- Accelerate: Not affected
- TGI: Not affected
- vLLM: Not affected

## Cross-PR Conflicts

- PR #1871 (Remove deprecated APIs): Overlaps on `bitsandbytes/nn/modules.py`. No semantic conflict with the Linear8bitLt changes.
- PR #1866 (Fix Params4bit attribute access for FSDP): Overlaps on `bitsandbytes/nn/modules.py`. No semantic conflict.

- Security: Clear
- Downstream impact: LOW (positive — fixes a user-facing crash)
- Tests: New regression test is good; but the fix causes 64 existing serialization tests to fail
- CI: 64 test failures in `test_linear_serialization` caused by this PR
- Serialization: `_save_to_state_dict` / `_load_from_state_dict` changes are compatible; the `forward()` guard breaks deserialized modules
- Commit hygiene: 3 commits, one unrelated. Suggest squash merge after fixes.

Issue #1118, opened 2024-03-09
LAMB missing in str2optimizer32bit?


Xynonners:
### System Info

Python 3.11 venv, Arch Linux

### Reproduction

```
import torch
import bitsandbytes as bnb

a = [torch.rand(1), torch.rand(1)]
opt = bnb.optim.LAMB8bit(a)
a[0].grad = torch.rand(1)
opt.step()
```

venv/lib/python3.11/site-packages/bitsandbytes/functional.py", line 1211, in optimizer_update_32bit
    optim_func = str2optimizer32bit[optimizer_name][0]
                 ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
KeyError: 'lamb'


### Expected behavior

it steps the optimizer?

EtienneDosSantos:
Ran into this KeyError as well and you're absolutely correct with your assumption that it's missing in `str2optimizer32bit = {},`. I've opened a pull request #1222 to address this issue.

Titus-von-Koeller:
Thanks for raising this and helping with the fix ❤️ Situation is fixed in the upcoming patch release.

PR #1870, opened 2026-02-16
Add LARS to str2optimizer32bit and fix error message

### LINKED ISSUE #1810: LARS missing in str2optimizer32bit

TimDettmers:
## Summary
- Add `"lars"` entry to `str2optimizer32bit` dictionary, mapping to momentum kernels (`cmomentum32bit_grad_32`, `cmomentum32bit_grad_16`). This follows the same pattern as LAMB→adam kernels (fixed in #1118).
- Fix incorrect error message in `_optimizer_update_32bit_impl` that displayed `str2optimizer8bit_blockwise.keys()` instead of `str2optimizer32bit.keys()`.
- Add LARS to the parametrized 32-bit optimizer tests in `test_optim.py`, using `PytorchLARS` as the reference implementation, with bf16 skip (matching momentum/rmsprop behavior since the momentum kernels lack a bf16 variant).

This builds on the approach from PR #1855 by @Mr-Neutr0n but also addresses the wrong error message, which was the second bug reported in the issue.

Fixes #1810

## Test plan
- [x] `pytest tests/test_optim.py -v --tb=short -k "lars"` — 8 passed, 4 skipped (bf16, expected)
- [x] `ruff check` and `ruff format` pass cleanly

### DIFF
--- a/bitsandbytes/backends/cuda/ops.py
+++ b/bitsandbytes/backends/cuda/ops.py
@@ -574,6 +574,10 @@ str2optimizer32bit = {
         lib.cademamix32bit_grad_fp16,
         lib.cademamix32bit_grad_bf16,
     ),
+    "lars": (
+        lib.cmomentum32bit_grad_32,
+        lib.cmomentum32bit_grad_16,
+    ),
 }
 
 str2optimizer8bit_blockwise = {
@@ -633,7 +637,7 @@ def _optimizer_update_32bit_impl(
     optim_fns = str2optimizer32bit.get(optimizer_name, None)
     if optim_fns is None:
         raise ValueError(
-            f"Unsupported optimizer name: {optimizer_name}. Supported optimizers: {list(str2optimizer8bit_blockwise.keys())}"
+            f"Unsupported optimizer name: {optimizer_name}. Supported optimizers: {list(str2optimizer32bit.keys())}"
         )
     if g.dtype == torch.float32:
         optim_func = optim_fns[0]
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -102,6 +102,11 @@ str2optimizers["momentum8bit_blockwise"] = (
     lambda pxx: bnb.optim.SGD8bit(pxx, 0.01, 0.9, block_wise=True),
 )
 
+str2optimizers["lars"] = (
+    lambda pxx: bnb.optim.PytorchLARS(pxx, 0.01, 0.9),
+    lambda pxx: bnb.optim.LARS(pxx, 0.01, 0.9),
+)
+
 str2optimizers["rmsprop"] = (
     lambda pxx: torch.optim.RMSprop(pxx, 0.01, 0.9),
     lambda pxx: bnb.optim.RMSprop(pxx, 0.01, 0.9, block_wise=False),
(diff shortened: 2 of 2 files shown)


TimDettmers:
Closing in favor of pushing the additional fixes (error message + tests) to the existing PR #1855.

PR #1876, opened 2026-02-17
[CI] Test against PyTorch 2.10

matthewdouglas:


matthewdouglas:
There are some test regressions in torch 2.10 that were noted on Slack. For the time being these are being xfailed.

The igemm test failures are specifically on 2.10+cu130. It's also likely this functionality will be deprecated/removed.

PR #1877, opened 2026-02-18
Unify CUDA and HIP kernel sources via compat.cuh portability layer

Abdennacer-Badaoui:
## Summary                           

  Merge the duplicated CUDA and HIP kernel sources into a single codebase. Instead of maintaining near-identical file pairs that inevitably drift apart, all GPU code now lives in shared files that compile for both platforms.                                                                    
   
  ## Problem                                                                                                                                                                                                                                                                                        
                                                         
  We maintained near-identical copies of every GPU kernel:

  | CUDA | HIP | ~LOC each |
  |------|-----|-----------|
  | `kernels.cu` | `kernels.hip` | 2600+ |
  | `kernels.cuh` | `kernels_hip.cuh` | 130 |
  | `ops.cu` | `ops.hip` | 650+ |
  | `ops.cuh` | `ops_hip.cuh` | 190 |
  | `common.cuh` | `common_hip.cuh` | 45 / 11 |

  The HIP files were originally auto-generated by hipify and manually patched. Every bug fix or new feature had to be applied to both copies.

  ## Approach

  Two portability headers abstract all CUDA/HIP differences:

  - **`compat.cuh`** — Host-safe types and macros (safe to include from `.cpp` files)
  - **`compat_device.cuh`** — Device-only layer: CUB/hipCUB, reduction ops, MMA (include from `.cu` files only)

  These provide macros, type aliases, and namespace aliases:

  - `bnb_cub::` → `cub::` on CUDA, `hipcub::` on HIP
  - `bnb_bfloat16` → `__nv_bfloat16` on CUDA, `hip_bfloat16` on HIP
  - `bnb_stream_t` → `cudaStream_t` / `hipStream_t`
  - `BNB_MAX_OP` → `cub::Max()` / `hipcub::Max()`
  - `BNB_CHECK_RETURN()` → error checking
  - `bnb_blasLt*`, `bnb_sparse*` → cuBLAS/hipBLAS and cuSPARSE/hipSPARSE

  The `<<<grid, block>>>` launch syntax works natively on HIP, so no `hipLaunchKernelGGL` wrappers are needed. For HIP builds, CMake sets `LANGUAGE HIP` on the `.cu` files.

  `#if BNB_HIP` guards are only needed for genuinely divergent code (~10% of changes):
  - `atomicMax` (CUDA needs CAS loop, HIP has native)
  - `Context` class (cuBLAS vs rocBLAS handle creation)
  - `igemmlt` (hipBLAS requires explicit heuristic algo selection)
  - Warp-size-dependent kernels (via `BNB_WARP_SIZE` compile-time constants)

  The two-header split is necessary because `.cpp` files (like `pythonInterface.cpp`) are compiled by the host compiler (gcc/g++), which cannot parse CUB/device headers.

  ## Changes

  - **Deleted 5 files:** `common_hip.cuh`, `kernels.hip`, `kernels_hip.cuh`, `ops.hip`, `ops_hip.cuh`
  - **Unified 5 files:** `common.cuh`, `kernels.cu`, `kernels.cuh`, `ops.cu`, `ops.cuh`
  - **Added 2 files:** `compat.cuh`, `compat_device.cuh`
  - **Updated:** `CMakeLists.txt`, `pythonInterface.cpp`
  - **Net result:** 10 files → 7 files


matthewdouglas:
@Abdennacer-Badaoui Thanks! This is essentially what I was expecting we could do. I think this is a good way forward. Most of the review comments above make sense as well!

TimDettmers:
## MI300X Runtime Validation of Unified CUDA/HIP Kernels

I validated this PR's unified kernel approach on actual AMD MI300X hardware. The unified files were activated from `csrc/examples/`, several build issues were fixed, and the full test suite was run.

### Test Environment

- **GPU**: AMD Instinct MI300X OAM (192 GB HBM3)
- **ROCm**: 6.1.0
- **PyTorch**: 2.6.0+rocm6.1
- **Host**: RunPod (EU-RO-1), Ubuntu 22.04, kernel 6.5.0
- **Build**: `cmake -DCOMPUTE_BACKEND=hip -DBNB_ROCM_ARCH="gfx942"`

### Test Results

```
3136 passed, 321 skipped, 0 failed, 24 xfailed, 738 warnings in 206.20s
```

**Full test suite passes with zero failures.**

### Build Fixes Required

The unified files from `csrc/examples/` needed several fixes to compile on HIP/gfx942. These are on branch `unified-hip-validation`:

1. **`BNB_WARP_SIZE` detection** (`common.cuh`): The original code used `#ifdef __GFX9__` to set warp size to 64 for CDNA. However, `__GFX9__` is only defined during the device compilation pass — the host compilation pass (which also instantiates templates) doesn't see it. Fixed by using `__AMDGCN_WAVEFRONT_SIZE` instead, which the HIP compiler correctly defines in both passes.

2. **`BLOCK_LOAD_WARP_TRANSPOSE` with small blocks** (`kernels.cu`): `kQuantizeBlockwise` with `BLOCK_SIZE=64` uses only 32 threads (`64/NUM_PER_TH=2`). On CDNA (warp=64), `BLOCK_LOAD_WARP_TRANSPOSE` requires `block_dim >= warp_size`, so 32 threads fails. Fixed by conditionally selecting `BLOCK_LOAD_DIRECT` when `THREADS < BNB_WARP_SIZE`:
   ```cpp
   static constexpr auto LOAD_ALGO =
       (THREADS >= BNB_WARP_SIZE) ? bnb_cub::BLOCK_LOAD_WARP_TRANSPOSE : bnb_cub::BLOCK_LOAD_DIRECT;
   ```

3. **Missing `common.h` include** (`ops.cuh`): The unified `ops.cuh` dropped the `#include "common.h"` needed for `DataType_t` enum resolution.

### Branch

All fixes are on [`unified-hip-validation`](https://github.com/bitsandbytes-foundation/bitsandbytes/tree/unified-hip-validation) — the commits on top of `merge-cuda-hip` show the full diff of what was needed to go from the example files to a working build + passing tests.

### Summary

The unified kernel approach works well. The compat headers successfully abstract CUDA/HIP differences, and the conditional compilation produces correct code for MI300X. The three issues found above are all fixable within the compat header pattern — no fundamental problems with the unification design.

Abdennacer-Badaoui:
I have merged @TimDettmers’ changes into my branch. The PR is ready to be merged if you’re okay with the current state of the unification. If you have any suggestions or improvements, please let me know.

PR #1879, opened 2026-02-20
Fix ruff lint and format issues in agents/ scripts

Abdennacer-Badaoui:
Description: Rename ambiguous variable l to label/lbl (E741) and apply ruff formatting in`agents/fetch_issues.py` and `agents/query_issues.py`.


Abdennacer-Badaoui:
I'm bypassing the rules on this one and merging it, as it only involves a linting/formatting fix.

PR #1886, opened 2026-03-04
[ROCm] Enable more targets

sstamenk:
## Summary

Add missing RDNA GPU targets to ROCm builds:
- **RDNA 3** (all ROCm versions): gfx1102, gfx1103
- **RDNA 3.5 / RDNA 4** (ROCm 6.4+): gfx1152, gfx1153

Updated in CI build script, CMake defaults, README, and installation docs.

### DIFF
--- a/bitsandbytes/backends/default/ops.py
+++ b/bitsandbytes/backends/default/ops.py
@@ -248,6 +248,9 @@ def _(
 
     # Quantize with the lookup table
     code = CODE[quant_type].to(scaled.device).to(scaled.dtype)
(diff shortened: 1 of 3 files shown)


sstamenk:
I haven't run the unit tests for these targets since I don't have access to the hardware.

matthewdouglas:
> I haven't run the unit tests for these targets since I don't have access to the hardware.

This is OK. We don't always run the tests on all of the NVIDIA GPUs that we build for either. Right now gfx1153 seems like it would be particularly difficult to acquire :)

PR #1899, opened 2026-03-13
Update CUDA/ROCm setup tests

sstamenk:
Updated backend-specific setup handling and tests for CUDA/ROCm library selection. The change makes `get_cuda_bnb_library_path()` fail fast when the wrong override variable is used (`BNB_ROCM_VERSION` on CUDA or `BNB_CUDA_VERSION` on ROCm), improves the warning/error guidance, and refreshes the evaluator tests to use `BNB_BACKEND` with clearer CUDA/ROCm coverage.
Removed the `test_get_rocm_bnb_library_path_rocm_override_takes_priority` unit-test as it doesn't provide much utility.

This PR is a subset of the changes originally proposed in https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1888

### DIFF
--- a/bitsandbytes/cextension.py
+++ b/bitsandbytes/cextension.py
@@ -33,32 +33,49 @@ def get_cuda_bnb_library_path(cuda_specs: CUDASpecs) -> Path:
     cuda_override_value = os.environ.get("BNB_CUDA_VERSION")
     rocm_override_value = os.environ.get("BNB_ROCM_VERSION")
 
-    if rocm_override_value:
-        library_name = re.sub(r"rocm\d+", f"rocm{rocm_override_value}", library_name, count=1)
-        if torch.version.cuda:
-            raise RuntimeError(
-                f"BNB_ROCM_VERSION={rocm_override_value} detected for CUDA!\n"
-                "Use BNB_CUDA_VERSION instead: export BNB_CUDA_VERSION=<version>\n"
-                "Clear the variable and retry: unset BNB_ROCM_VERSION\n"
+    if torch.version.hip:
+        if cuda_override_value:
+            if not rocm_override_value:
+                raise RuntimeError(
+                    f"BNB_CUDA_VERSION={cuda_override_value} detected but this is not a CUDA build!\n"
+                    "Use BNB_ROCM_VERSION instead: export BNB_ROCM_VERSION=<version>\n"
+                    "Clear the variable and retry: unset BNB_CUDA_VERSION\n"
+                )
+            logger.warning(
+                f"WARNING: BNB_CUDA_VERSION={cuda_override_value} is set but ignored on this ROCm build. "
+                "Clear the variable: unset BNB_CUDA_VERSION",
             )
-        logger.warning(
-            f"WARNING: BNB_ROCM_VERSION={rocm_override_value} environment variable detected; loading {library_name}.\n"
-            "This can be used to load a bitsandbytes version built with a ROCm version that is different from the PyTorch ROCm version.\n"
-            "If this was unintended clear the variable and retry: unset BNB_ROCM_VERSION\n"
-        )
-    elif cuda_override_value:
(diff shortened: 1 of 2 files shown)


matthewdouglas (changes requested):
## Review: Update CUDA/ROCm setup tests

(Note: this review was created with the assistance of an LLM.)

Thanks for the PR! The fail-fast behavior for wrong override variables and the `HIP_ENVIRONMENT` → `BNB_BACKEND` migration are good improvements. A couple of issues to address before merging:

### 1. Missing path assertion in `test_get_rocm_bnb_library_path_override`

The diff removes the functional assertion that validates the override actually produces the correct library path:

```python
assert get_cuda_bnb_library_path(rocm70_spec).stem == "libbitsandbytes_rocm72"
```

The test now only checks `assert "BNB_ROCM_VERSION" in caplog.text`, which verifies the warning was emitted but not that the correct library was selected. This looks like an accidental omission — please restore it.

### 2. `test_get_cuda_bnb_library_path_override` is fragile

This test sets `BNB_CUDA_VERSION` but doesn't clear `BNB_ROCM_VERSION`. With the new logic, `rocm_override_value` is checked first — so if `BNB_ROCM_VERSION` happens to be set in the environment, the test will raise `RuntimeError` instead of exercising the CUDA override path. Please add:

```python
monkeypatch.delenv("BNB_ROCM_VERSION", raising=False)
```

(The base test `test_get_cuda_bnb_library_path` already does this correctly.)

### Minor notes

- The dead `re.sub(r"rocm\d+", ...)` call in the CUDA+`BNB_ROCM_VERSION` error path is harmless (regex doesn't match `libbitsandbytes_cuda*`) but could be avoided by moving the substitution after the `torch.version.cuda` check.
- Consider keeping an adapted version of `test_get_rocm_bnb_library_path_rocm_override_takes_priority` — the "both overrides set" edge case is still reachable and worth covering, especially now that the branching logic has changed. Not blocking though.

sstamenk:
@matthewdouglas everything should be good now, here is an overview of the changes:

### Changes addressing review feedback on `get_cuda_bnb_library_path` and its tests

#### `bitsandbytes/cextension.py`

**Restructured override logic to branch on backend first.** The previous `if rocm_override_value` / `elif cuda_override_value` structure branched on which env var was set, causing two issues:
- The `re.sub` ran before the wrong-backend error check (harmless but dead code on the wrong backend)
- When both `BNB_ROCM_VERSION` and `BNB_CUDA_VERSION` were set, behavior was asymmetric: error on CUDA, silent ignore on ROCm

The new structure dispatches on `torch.version.hip` / `torch.version.cuda` first, then handles overrides within each backend branch:

| Scenario | Behavior |
|---|---|
| Only wrong variable set | `RuntimeError` (unchanged) |
| Both set | Applies the correct one, warns about the ignored wrong one **(new)** |
| Only correct variable set | Applies override with existing warning (unchanged) |
| Neither set | Uses detected version (unchanged) |
| Other backends (e.g. XPU) | Rejects both overrides with `RuntimeError` **(new)** |

#### `tests/test_cuda_setup_evaluator.py`

- **Fixed `test_get_cuda_bnb_library_path_override`** — added `monkeypatch.delenv("BNB_ROCM_VERSION", raising=False)` so a stray env var doesn't cause the test to hit the wrong code path
- **Mirrored in `test_get_rocm_bnb_library_path_override`** — added `monkeypatch.delenv("BNB_CUDA_VERSION", raising=False)` for symmetry
- **Added `test_get_cuda_bnb_library_path_cuda_override_takes_priority`** — both overrides set on CUDA: CUDA wins, ROCm override warned about
- **Added `test_get_rocm_bnb_library_path_rocm_override_takes_priority`** — both overrides set on ROCm: ROCm wins, CUDA override warned about
- **Updated error match strings** — reflect new messages (`not a ROCm/CUDA build` instead of `detected for CUDA/ROCm`)

#### Tested

ROCm tests verified passing on `torch 2.9.1+rocm7.2.0` (4 passed, 4 CUDA-only skipped).

PR #1907, opened 2026-03-29
feat: Implement 4-bit and 8-bit Quantized Convolutional Layers

### LINKED ISSUE #1414: Support for quantization of convolutional layers
### Feature request

Would be great if the library could support more types of layers for quantization like with `torch.ao`, seems like it is already available with [conv2d](https://pytorch.org/docs/stable/generated/torch.ao.nn.quantized.Conv2d.html). Sadly `torch.ao` does not seem to support CUDA as a backend right now. Would it be possible to implement the 8-bit and 4-bit kernels in Triton or CUDA to allow for the quantization of convolutional layers? A similar issue has been raised [here](https://github.com/bitsandbytes-foundation/bitsandbytes/issues/29) earlier.

### Motivation

Make modules that use convolutional layers use less memory through quantization.

### Your contribution

Yes, I am willing to work on implementing convolutional kernels if it is possible to integrate with this library.

OnePunchMonk:
This PR introduces support for 4-bit and 8-bit quantized convolutional layers as requested in #1414.

Included:
- \Conv1d4bit\, \Conv2d4bit\, \Conv3d4bit\ with FP4 and NF4 types
- \Conv1d8bitLt\, \Conv2d8bitLt\, \Conv3d8bitLt\
- Complete test suite matching bitsandbytes standards
- Resolves the missing support for vision and audio models relying natively on convolutions.

Fixes #1414


matthewdouglas:
Thanks for the contribution and the effort that went into this — the code is well-structured and the test coverage is thorough.

After reviewing, I don't think this is something we should add to bitsandbytes right now, and I want to explain why rather than just closing it.

Both the 4-bit and 8-bit conv classes here do dequantize-then-compute: weights are stored in reduced precision but fully dequantized back to fp16/fp32 before the forward pass, delegating to standard `F.conv*d`. For the 8-bit case in particular this creates a misleading class name — unlike `Linear8bitLt` which actually runs int8 tensor core arithmetic, `Conv8bitLt` doesn't do any int8 compute. The memory savings are real but achievable today with `Params4bit` and a custom forward.

The broader question is who this is for. The dominant use case for bitsandbytes is transformer models with large linear layers. The one place transformer architectures use something conv-shaped — GPT-2's `Conv1D` — is already handled by transposing the weight to make it compatible with the standard linear quantization path. The audience this PR targets is conv-heavy models like vision and audio backbones, which is a genuinely different audience with different memory profiles and tooling expectations. We haven't seen meaningful demand signal from that audience — the linked issue has a single reaction and no comments — and adding 12 new public classes is a long-term maintenance commitment we want to be confident about before taking on.

I'm going to close this PR for now. Thanks again for putting the work in.

PR #1909, opened 2026-04-02
update legend

jiqing-feng:
Hi @matthewdouglas . Since we have enabled CPU optimizer and XPU paged optimizer. The legend should be updated to full support for CPU and XPU. Please review this PR. Thanks!


matthewdouglas:
Thanks!

PR #1914, opened 2026-04-03
[ROCm] Bump ROCm version to 7.2.1

sstamenk:
ROCm 7.2.1 released last week. Updating the GitHub workflow and documentation to reflect this.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -136,7 +136,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         arch: [x86_64]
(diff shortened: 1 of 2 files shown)


sstamenk:
Wasn't sure if I should add a new entry so I just replaced the existing 7.2 one.

matthewdouglas:
> Wasn't sure if I should add a new entry so I just replaced the existing 7.2 one.

LGTM. This is generally what we've been doing, i.e. just bumping to newer patch releases. When we build and ship these they're only named by major/minor version number.

PR #1916, opened 2026-04-07
Fix torch.compile graph breaks from Params4bit __getattr__ (#1904, #1917)

### LINKED ISSUE #1904: Params4bit.__getattr__ breaks torch.compile - use @property instead
### System Info

PR #1866 added `__getattr__` to `Params4bit` for FSDP state_dict support. This works fine for FSDP, but it breaks `torch.compile`.

`Params4bit` is a `torch.Tensor` subclass. PyTorch's Dynamo (the compiler frontend) doesn't know how to trace tensor subclasses that define `__getattr__`, so it creates graph breaks whenever it encounters attribute access on such objects. With activation checkpointing, these graph breaks multiply across layers, resulting in many more subgraphs than necessary and significant compilation overhead.

We noticed this when running `torch.compile` on a QLoRA fine-tuning workload (LLaMA 70B, HuggingFace, activation checkpointing). With `__getattr__` present, we saw significant performance degradation caused by graph breaks. Removing `__getattr__` (or replacing with `@property` as proposed below) restores expected performance.

### Reproduction

I put together a minimal repro using `torch._dynamo.explain()` on a small model with `Linear4bit` and activation checkpointing. With `__getattr__` present it showed graph breaks; after removing it, they were gone. The script is rough, so I'd appreciate it if you could verify this with a more representative test — or suggest one if you have something better suited.

### Expected behavior

Replace `__getattr__` + `_QUANT_STATE_ATTR_MAP` with `@property` descriptors. Properties are resolved at the class level through Python's descriptor protocol — Dynamo handles them fine, no graph breaks. FSDP still works because `getattr(weight, "absmax")` resolves the same way. Example:

```python
    @property
    def absmax(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.absmax
        raise AttributeError("'Params4bit' object has no attribute 'absmax'")

    @property
    def code(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.code
        raise AttributeError("'Params4bit' object has no attribute 'code'")

    @property
    def quant_map(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.code
        raise AttributeError("'Params4bit' object has no attribute 'quant_map'")

    @property
    def offset(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.offset
        raise AttributeError("'Params4bit' object has no attribute 'offset'")

    @property
    def state2(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.state2
        raise AttributeError("'Params4bit' object has no attribute 'state2'")

    @property
    def nested_absmax(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None and qs.state2 is not None:
            return qs.state2.absmax
        raise AttributeError("'Params4bit' object has no attribute 'nested_absmax'")

    @property
    def nested_blocksize(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None and qs.state2 is not None:
            return qs.state2.blocksize
        raise AttributeError("'Params4bit' object has no attribute 'nested_blocksize'")

    @property
    def nested_quant_map(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None and qs.state2 is not None:
            return qs.state2.code
        raise AttributeError("'Params4bit' object has no attribute 'nested_quant_map'")

    @property
    def nested_dtype(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None and qs.state2 is not None:
            return qs.state2.dtype
        raise AttributeError("'Params4bit' object has no attribute 'nested_dtype'")

    @property
    def nested_offset(self):
        qs = self.__dict__.get("quant_state")
        if qs is not None:
            return qs.offset
        raise AttributeError("'Params4bit' object has no attribute 'nested_offset'")
```

### LINKED ISSUE #1917: torch.compile fullgraph breaks on CPU: matmul_4bit mutates quant_state.dtype
## Summary

`matmul_4bit()` mutates `quant_state.dtype` in-place on CPU ([_functions.py:387](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/bitsandbytes/autograd/_functions.py#L387)):

```python
if A.device.type == "cpu":
    quant_state.dtype = A.dtype
```

Under `torch.compile(fullgraph=True)` with activation checkpointing, Dynamo flags this as:

```
torch._dynamo.exc.Unsupported: HigherOrderOperator: Mutating a variable
not in the current scope (SideEffects)
```

This only affects CPU — on CUDA the branch is not taken, so GPU compilation works fine.

## How discovered

The regression test added in #1916 (`test_linear4bit_torch_compile_activation_checkpointing`) exercises `fullgraph=True` + `torch.utils.checkpoint`. It passes on all GPU backends but fails on CPU with torch 2.10.0 due to this pre-existing mutation. The test now skips on CPU to keep #1916 focused on the `__getattr__` → `@property` fix.

## Suggested fix

Refactor `matmul_4bit()` to avoid mutating `quant_state.dtype` in-place. For example, pass the input dtype as a parameter to the downstream functions instead of writing it onto the shared `QuantState` object. This would make `matmul_4bit` safe for Dynamo tracing on CPU.

## Reproducer

```python
import torch
import bitsandbytes as bnb

class CheckpointedNet(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.layers = torch.nn.ModuleList([
            bnb.nn.Linear4bit(256, 256, bias=False, compute_dtype=torch.bfloat16, quant_type="nf4")
            for _ in range(4)
        ])

    def forward(self, x):
        for layer in self.layers:
            x = torch.utils.checkpoint.checkpoint(layer, x, use_reentrant=False)
        return x

net = CheckpointedNet().to("cpu")
compiled = torch.compile(net, fullgraph=True, backend="inductor")
x = torch.randn(16, 256, dtype=torch.bfloat16, requires_grad=True)
compiled(x).sum().backward()  # raises Unsupported: Mutating a variable not in the current scope
```

## Environment

- torch 2.10.0
- CPU only (CUDA is unaffected)
- Observed on linux-x64, linux-aarch64, linux-x64-icelake, macOS

Titus-von-Koeller:
## Summary

- Replace `__getattr__` + `_QUANT_STATE_ATTR_MAP` on `Params4bit` with `@property` descriptors to eliminate `torch.compile` graph breaks under activation checkpointing
- Remove dead `quant_state.dtype` mutation in `matmul_4bit` CPU path that caused a separate `torch.compile` graph break under activation checkpointing
- Add regression test `test_linear4bit_torch_compile_activation_checkpointing` that compiles with `fullgraph=True` + `torch.utils.checkpoint` to catch this class of issue

## Context

PR #1866 added `__getattr__` to `Params4bit` for FSDP `state_dict` traversal. Since `Params4bit` is a `torch.Tensor` subclass, Dynamo cannot trace through `__getattr__`, creating graph breaks on every attribute access. With activation checkpointing these multiply across layers, causing significant compilation overhead (#1904).

`@property` descriptors use Python's descriptor protocol (resolved at class level), which Dynamo handles correctly — no graph breaks. FSDP still works because `getattr(weight, "absmax")` resolves the same way.

Attributes that collide with `Params4bit` instance attrs (`blocksize`, `quant_type`) or `torch.Tensor` attrs (`dtype`, `shape`) are intentionally omitted — they're packed into the `bitsandbytes__*` blob and never traversed by FSDP as separate keys.

`QuantState.__getattr__` is left unchanged since `QuantState` is not a `Tensor` subclass.

Additionally, `matmul_4bit()` mutated `quant_state.dtype = A.dtype` on the CPU path (#1917). This in-place mutation is unnecessary — `MatMul4Bit.forward` already casts via `.to(A.dtype)`, and `gemv_4bit` doesn't read `state.dtype`. Removing it eliminates the Dynamo graph break on CPU under activation checkpointing.

## Validated against three code states

| State | Regression test | Existing tests |
|---|---|---|
| Before #1866 (no `__getattr__`) | 4 passed | N/A |
| At #1866 (with `__getattr__`) | **4 failed** (graph break) | no existing tests failed |
| After this fix (`@property`) | 4 passed | 2765 passed, none failed |

## Test plan

- [x] `test_linear4bit_torch_compile_activation_checkpointing` — 4 variants pass (nf4/fp4 × compress_statistics), including CPU
- [x] `test_linear4bit_torch_compile` — all 64 variants pass (no regressions)
- [x] `test_params4bit_quant_state_attr_access` — all 4 variants pass (FSDP traversal still works)
- [x] ran complete test suite — clean
- [x] `pre-commit run --all-files` — clean

Fixes #1904
Fixes #1917

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -382,10 +382,7 @@ def matmul_4bit(
     bias: Optional[torch.Tensor] = None,
 ):
     assert quant_state is not None
-    # Change dtype to input dtype on CPU
     if A.device.type == "cpu":
-        quant_state.dtype = A.dtype
-
         if getattr(quant_state, "packing_format_for_cpu", False):
             out = F.gemv_4bit(A, B, out, state=quant_state)
             if bias is not None:
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -453,9 +453,6 @@ def test_linear4bit_torch_compile_activation_checkpointing(device, quant_type, c
         pytest.skip("This configuration is not supported on HPU.")
     if device == "cuda" and platform.system() == "Windows":
         pytest.skip("Triton is not officially supported on Windows")
-    if device == "cpu":
-        pytest.skip("matmul_4bit mutates quant_state.dtype on CPU, causing a separate graph break (#1917)")
-
     dim = 256
     batch_size = 16
     compute_dtype = torch.bfloat16


Titus-von-Koeller:
## PR Review: #1916 -- Fix torch.compile graph breaks from Params4bit \_\_getattr\_\_ (#1904)

**Classification:** `[bug-fix]`
**Size:** Small-medium (150 added / 37 deleted, 2 files)
**Author:** Titus-von-Koeller (maintainer)
**Verdict:** Approve

Bug fix replacing `__getattr__` + `_QUANT_STATE_ATTR_MAP` on `Params4bit` with `@property` descriptors to eliminate `torch.compile` graph breaks under activation checkpointing. Root cause is well-understood: Dynamo cannot trace through `__getattr__` on `torch.Tensor` subclasses, so every attribute access creates a graph break. With activation checkpointing these multiply across layers. The descriptor protocol (resolved at class level) is Dynamo-compatible.

**No blocking issues.**

The approach is correct. Properties are resolved via Python's descriptor protocol at the class MRO level, which Dynamo handles natively. The behavioral semantics are preserved: each property delegates to `self.__dict__.get("quant_state")` and raises `AttributeError` with the same message format when the attribute is unavailable. The four attributes that collided with Params4bit instance attrs (`blocksize`, `quant_type`) or Tensor attrs (`dtype`, `shape`) are correctly omitted -- these were dead code in `__getattr__` since regular attribute lookup would find them first, and as properties they would incorrectly shadow the instance/tensor attributes.

**One observation (non-blocking):**

The `nested_offset` property checks `qs is not None` but does *not* check `qs.state2 is not None`, unlike the other `nested_*` properties. This matches the old `_QUANT_STATE_ATTR_MAP` behavior exactly (`"nested_offset": lambda qs: qs.offset`), and it matches `as_dict()` which serializes `nested_offset` as `self.offset.item()` (not `self.state2.offset`). So this is correct as-is -- just noting it since the asymmetry is easy to misread.

**Regression test is well-designed.** Uses `fullgraph=True` which causes `torch.compile` to error on graph breaks rather than silently falling back to eager. Covers forward + backward with activation checkpointing, parametrized across devices, quant types, and compress_statistics. Existing `test_params4bit_quant_state_attr_access` continues to verify FSDP traversal of all state_dict keys.

**CI status:** All CUDA/GPU tests pass across A10, L40S, T4 (CUDA 11.8/12.8/13.0). All builds pass. Lint passes. 4 CPU test failures on torch 2.10.0 appear pre-existing/infra -- PR only touches Python code.

- **Security:** Clear
- **Downstream impact:** None -- behavioral contract preserved. `getattr(weight, "absmax")` still resolves. PEFT/Transformers `__dict__` round-trips unaffected (properties don't appear in `__dict__`). FSDP traversal works identically.
- **Serialization:** No change -- `as_dict()`, `from_dict()`, state_dict keys all unchanged
- **Cross-PR conflicts:** PR #1863 also touches `tests/test_linear4bit.py` but in a different test function; should merge cleanly in either order

PR #1919, opened 2026-04-09
Slightly relax `maxerr1` threshold for fp32 in `test_gemv_4bit`

jiqing-feng:
## Summary

Relax the `maxerr1` threshold for `torch.float32` with `dim > 512` from `4e-8` to `5e-8` to accommodate minor precision differences on XPU.

## Issue

The test `test_gemv_4bit[dim=1024-fp32-fc2-fp4-DQ_True-xpu]` fails with:

```
AssertionError: maxerr1=0.00000011 exceeds 0.00000004 + 7*0.00000001 = 0.00000011
assert 1.1177733540534973e-07 < 1.1e-07
```

The observed value (`1.1178e-07`) is only marginally above the threshold (`1.1e-07`), indicating minor numerical variance rather than a functional issue.

## Fix

Adjust the baseline mean for `maxerr1` from `4e-8` to `5e-8`, resulting in a new threshold of `1.2e-07` (with 7σ), which safely accommodates XPU's precision characteristics while remaining strict enough to catch real regressions.

### DIFF
--- a/tests/test_functional.py
+++ b/tests/test_functional.py
@@ -1140,7 +1140,7 @@ class TestQuantize4BitFunctional:
             },
             torch.float32: {
                 "le512": {"err1": (2e-8, 2e-9), "relerr1": (8e-7, 1.2e-6), "maxerr1": (6e-8, 2e-8)},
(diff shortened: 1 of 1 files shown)


matthewdouglas (approved):

matthewdouglas:
Looks reasonable to me!

PR #1921, opened 2026-04-13
Enable XPU for blockwise quantization and FSDP tests

jiqing-feng:
## Summary

Replace CUDA-only guards with device-agnostic `torch.accelerator` APIs so that XPU (and other accelerators) can run the blockwise quantization and FSDP state_dict tests.

## Changes

**tests/test_functional.py**
- Allow XPU to run the full `test_dynamic_blockwise_quantization` parameter matrix (nested, non-256 blocksizes, fp16/bf16) instead of skipping them as "non-CUDA".

**tests/test_linear4bit.py**
- Replace `torch.cuda.is_available()` and NCCL-specific skip guards on `test_fsdp_state_dict_save_4bit` with a single `torch.accelerator.is_available()` check.

**tests/fsdp_state_dict_save.py**
- Auto-detect the accelerator type via `torch.accelerator.current_accelerator()` and map it to the correct distributed backend (CUDA → nccl, XPU → xccl).
- Replace `torch.cuda.set_device()` with `torch.accelerator.set_device_index()` and use the detected device type for `model.to()`.

### DIFF
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -569,7 +569,14 @@ def test_params4bit_quant_state_attr_access(device, quant_type, compress_statist
     assert w.bnb_quantized is True
 
 
-@pytest.mark.skipif(not torch.accelerator.is_available(), reason="FSDP requires an accelerator device")
+@pytest.mark.skipif(
+    not (
+        (hasattr(torch, "accelerator") and torch.accelerator.is_available())
+        or torch.cuda.is_available()
+        or (hasattr(torch, "xpu") and torch.xpu.is_available())
+    ),
+    reason="FSDP requires an accelerator device",
+)
 def test_fsdp_state_dict_save_4bit():
     """Integration test: FSDP get_model_state_dict with cpu_offload on a 4-bit model (#1405).
 
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -451,7 +451,7 @@ def test_linear4bit_torch_compile_activation_checkpointing(device, quant_type, c
     """
     if device == "hpu" and not is_supported_on_hpu(quant_type):
         pytest.skip("This configuration is not supported on HPU.")
-    if device == "cuda" and platform.system() == "Windows":
+    if platform.system() == "Windows":
         pytest.skip("Triton is not officially supported on Windows")
     dim = 256
     batch_size = 16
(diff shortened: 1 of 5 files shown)


matthewdouglas:
Hi @jiqing-feng,

Using `torch.accelerator` is likely going to break this on older torch versions. I think it was introduced in PyTorch 2.6. This is probably a minimum requirement for XPU (if not even higher), but right now we support PyTorch 2.3+ for CUDA hardware. I will probably soon bump to 2.4, so we would want some backwards compatibility for that.

Otherwise, looks good!

jiqing-feng:
Hi @matthewdouglas . I have added backwards compatibility for torch versions without torch.accelerator by falling back to torch.cuda/torch.xpu checks and device setup, while keeping the torch.accelerator path for newer versions.

jiqing-feng:
The failed tests seems like an env issue. Is this because the change of `torch.cuda` to `torch.accelerator`? @matthewdouglas , do you have any suggestions about it?

matthewdouglas:
@jiqing-feng I think we just need to skip that test on Windows. Otherwise to have a chance at working we'd probably need to switch `nccl` to `gloo` and set `USE_LIBUV=0`. Even then I don't know that it would work, and it's not a use case that's particularly important. We have other tests being skipped when `platform.system() == "Windows"` too, so it's fine to just do that.

jiqing-feng:
Hi @matthewdouglas . I have skipped fsdp on Windows for all devices, please check it. Thanks!

tests/test_linear4bit.py
```diff
@@ -451,7 +451,7 @@ def test_linear4bit_torch_compile_activation_checkpointing(device, quant_type, c
     """
     if device == "hpu" and not is_supported_on_hpu(quant_type):
         pytest.skip("This configuration is not supported on HPU.")
-    if device == "cuda" and platform.system() == "Windows":
+    if platform.system() == "Windows":
```
matthewdouglas: This should be as it was, and what we needed to skip on windows was below at `test_fsdp_state_dict_save_4bit`

jiqing-feng:
Hi @matthewdouglas . I've reverted the wrong change and skip fsdp on Windows for all devices. Please review it. Thanks!

jiqing-feng:
Hi @matthewdouglas . Would you please rerun the CI and review it? Thanks!

jiqing-feng:
Hi @matthewdouglas . The failed CI is not related to my changes. It also happens in the main branch.

PR #1922, opened 2026-04-13
remove deprecated tests

jiqing-feng:
This PR removes deprecated test coverage to reduce noise and avoid generating cases that are intentionally unsupported.

### Changes

- `tests/test_autograd.py`
  - Stop generating `req_grad[1] == True` for `test_matmullt`.
  - Remove unreachable/dead assertions tied to that deprecated path.
- `tests/test_linear8bitlt.py`
  - Restrict `test_linear_serialization` to non-deprecated `has_fp16_weights=False`.
  - Remove skip-based handling for deprecated `has_fp16_weights=True` combinations.
- `tests/test_modules.py`
  - Remove deprecated test cases.
- `tests/test_functional.py`
  - Remove deprecated IGEMM functional test block.

### Impact

- Smaller and cleaner test matrix.
- No behavior change in supported runtime paths.
- CI signal improves by removing deprecated/skip-heavy combinations.


tests/test_functional.py
```diff
@@ -723,40 +446,6 @@ def test_dequant_mm(self, device, dim1, dim4, dims, has_bias):
             n = C5.numel()
             assert_all_approx_close(C1, C4, atol=0.015, rtol=0.1, count=int(0.01 * n))
 
-    @pytest.mark.parametrize("dim1", [2048, 4096], ids=id_formatter("dim1"))
-    @pytest.mark.parametrize("dim2", [512, 1024], ids=id_formatter("dim2"))
-    def test_int8_double_quant(self, dim1, dim2):
```
matthewdouglas, line 728: This is the main one where I hesitate because we may still have external users calling `int8_double_quant` - even if they're not actually training int8 weights.

There's a tip in the docs about this but this is one where we haven't emitted deprecation warnings on either. 

```
<Tip>
    This function is useful for training, but for inference it is advised to use [`int8_vectorwise_quant`] instead.
    This implementation performs additional column-wise transposed calculations which are not optimized.
</Tip>
```

My ask here is that for now we change this one to only test on CUDA and give this one a short deprecation cycle before removing. This test doesn't necessarily assume int8 training and I believe is our only coverage on the op.

matthewdouglas (dismissed):

matthewdouglas:
I'm not fundamentally opposed to removing these tests as we generally either don't run them, or would be unlikely to take action on them if they did fail. The main hesitation is that we haven't gone through a deprecation cycle on some of these. Particularly `int8_double_quant` as mentioned.

As for the rest, considering I was planning to start emitting deprecation warnings for all of those features specifically before we push out v0.50.0, I think they are OK.

jiqing-feng:
Hi @matthewdouglas . I have added `test_int8_double_quant` back, and only limit it on cuda. Please review it. Thanks!

PR #1928, opened 2026-04-20
feat: add anonymous feature-usage telemetry

Titus-von-Koeller:
## Summary

Adds lightweight, opt-out, anonymous feature-usage telemetry via `huggingface_hub.utils.send_telemetry()`, mirroring the pattern Transformers already uses for its `quant` user-agent field. Data lands in the HF Hub telemetry index under `path_prefix=/api/telemetry/bitsandbytes/` and answers two concrete questions: which features deserve continued maintenance and which are safe to retire.

Key design points:

- **Namespaced**: all metadata keys under `bitsandbytes.*` so they don't collide in the shared telemetry index.
- **Session fingerprint + per-feature events**: the fingerprint (bnb version, OS, arch, libc, Python/torch versions, accelerator vendor/name/arch/count) is sent with each feature event; each distinct feature fires at most once per process.
- **Wired at first-use, not import time**: passive installs send nothing.
- **Opt-out**: `BNB_DISABLE_TELEMETRY`, `HF_HUB_DISABLE_TELEMETRY`, or `HF_HUB_OFFLINE`.
- **Auto-disabled under pytest** so CI and local test runs don't pollute the real-usage stream.
- **Silent no-op** when `huggingface_hub` is not installed.
- **No new runtime dependency**: `huggingface_hub` stays optional.

### Features tracked

| topic | call site | rationale |
|-------|-----------|-----------|
| `linear_4bit` | `Linear4bit.forward` | QLoRA usage shape (nf4/fp4, blocksize, compute_dtype) |
| `linear_8bit` | `Linear8bitLt.forward` | LLM.int8() usage (threshold, has_fp16_weights) |
| `params_4bit` | `Params4bit.__new__` | catches direct use outside `Linear4bit` (PEFT, vLLM, custom) |
| `int8_params` | `Int8Params.__new__` | same, for 8-bit |
| `embedding` | all 4 embedding variants | `variant=stable\|standard\|8bit\|4bit` |
| `optimizer` | `Optimizer8bit.step` | optimizer name + `is_paged` |
| `optim_override_config` | `GlobalOptimManager.override_config` | which keys users actually override |
| `optim_register_module_override` | `GlobalOptimManager.register_module_override` | existence signal for the module-level override mechanism |
| `outlier_aware_linear` | `OutlierAwareLinear.__init__` | deprecation candidate |
| `int8_double_quant` | `int8_double_quant()` | deprecation candidate |

### What is never collected

Model names, file paths, tensor shapes, parameter values, user identifiers, or anything derived from user input. The `send_telemetry()` call explicitly passes `token=False`, so no auth / identity info is attached.

### End-to-end verification

`scripts/verify_telemetry.py` exercises every wired-up feature once, tagging every event with a unique `BNB_TELEMETRY_TAG` so a single run's events can be correlated in `ds-hub-telemetry`:

```bash
python scripts/verify_telemetry.py
# prints run_id = verify-XXXXXXXX
# then, after ~30s:
es-cli -H esql 'FROM ds-hub-telemetry
  | WHERE metadata.bitsandbytes.tag == "verify-XXXXXXXX"
  | STATS count = COUNT(*) BY metadata.bitsandbytes.feature
  | SORT count DESC'
```

## Test plan

- [x] 22 new unit tests in `tests/test_telemetry.py` pass (dedup, namespacing, stringification, fingerprint fields, all three opt-out env vars, truthy-value parsing, tag attachment, graceful fallback when `huggingface_hub` is missing, exception swallowing, pytest auto-detection).
- [x] Existing `tests/test_linear4bit.py` and `tests/test_linear8bitlt.py` pass — no regressions (322 tests).
- [x] Existing `tests/test_modules.py` and `tests/test_optim.py` pass — no regressions (509 tests).
- [x] Full `pre-commit run --all-files` passes.
- [x] Manual end-to-end verification: `scripts/verify_telemetry.py` fired with `run_id=verify-4738a43c` — to be confirmed in `ds-hub-telemetry`.
- [ ] Review by @matthewdouglas who shaped the feature list and namespacing decisions.

## Documentation

README `## Telemetry` section explains what is collected, what is not, and the three opt-out env vars. The module docstring in `bitsandbytes/_telemetry.py` is the authoritative reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,43 @@ bitsandbytes has the following minimum requirements for all platforms:
 * 🤗 [Diffusers](https://huggingface.co/docs/diffusers/quantization/bitsandbytes)
 * 🤗 [PEFT](https://huggingface.co/docs/peft/developer_guides/quantization#quantize-a-model)
 
+## Telemetry
+
+`bitsandbytes` sends anonymous, aggregate feature-usage telemetry to the
+Hugging Face Hub. This data is used to prioritize maintenance (which quantization
+methods and optimizers are actually in use?) and to safely retire features that
+are no longer called by anyone.
+
+### What is collected
+
+* A session fingerprint sent once per process: `bitsandbytes` version, OS
+  name/version, CPU architecture, Python/PyTorch versions, accelerator
+  vendor/name/arch/count (e.g. `nvidia`, `NVIDIA H100`, `sm_90`, `1`).
+* One event per distinct feature used, with feature-specific flags. For
+  example: using `Linear4bit` sends `quant_type=nf4`, `blocksize=64`; using
+  `AdamW8bit.step()` sends `name=AdamW8bit`, `is_paged=false`.
+
+### What is never collected
+
+Model names, file paths, tensor shapes, parameter values, user identifiers, or
+anything derived from user input.
+
+### How to opt out
+
+Set any one of these environment variables:
+
+| Variable                     | Scope                        |
+| ---------------------------- | ---------------------------- |
+| `BNB_DISABLE_TELEMETRY=1`    | `bitsandbytes` only          |
+| `HF_HUB_DISABLE_TELEMETRY=1` | all Hugging Face libraries   |
+| `HF_HUB_OFFLINE=1`           | all Hugging Face libraries   |
+
+Telemetry is also automatically suppressed while running under `pytest` (so
+CI and local test runs don't pollute the stream) and a silent no-op when
+`huggingface_hub` is not installed. The implementation lives in
+[`bitsandbytes/_telemetry.py`](bitsandbytes/_telemetry.py) and each event
+fires at most once per process.
+
 ## :heart: Sponsors
 The continued maintenance and development of `bitsandbytes` is made possible thanks to the generous support of our sponsors. Their contributions help ensure that we can keep improving the project and delivering valuable updates to the community.
 
--- a/bitsandbytes/_telemetry.py
+++ b/bitsandbytes/_telemetry.py
@@ -1,0 +1,231 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Anonymous feature-usage telemetry for bitsandbytes.
+
+Sends one HEAD request per distinct feature per process via
+`huggingface_hub.utils.send_telemetry()`. Data lands in the Hugging Face
+Hub telemetry index under `path_prefix == "/api/telemetry/bitsandbytes/"`
+and informs maintenance and deprecation decisions.
+
+What is collected
+    - Session fingerprint (once per process, first feature use):
+      bnb version, OS name/version, CPU arch, glibc version, Python/torch
+      versions, accelerator vendor/name/arch/count.
+    - Per-feature events: feature name plus feature-specific metadata
+      (e.g. `quant_type="nf4"`, `bits="8"`, `paged="true"`).
+
+What is NOT collected
+    Model names, file paths, parameter shapes, user identifiers, training
+    data, gradient values, or any value derived from user input.
+
+Automatically disabled when running under pytest (detected via
+`pytest` in `sys.modules` or `PYTEST_CURRENT_TEST` env var) so that test
+runs in CI and locally do not pollute the real-usage stream.
+
+Opt-out (any of the following env vars disables all telemetry):
+    - BNB_DISABLE_TELEMETRY=1           (bitsandbytes only)
+    - HF_HUB_DISABLE_TELEMETRY=1        (all HF libraries)
+    - HF_HUB_OFFLINE=1                  (all HF libraries)
+
+End-to-end verification:
+    Set `BNB_TELEMETRY_TAG=<some-id>` before importing bitsandbytes and the
+    value is attached as `bitsandbytes.tag` on every event. Use this to
+    correlate a single run's events in ES.
+
+No-ops silently if `huggingface_hub` is not installed, and never raises.
+
+Keys are namespaced under `bitsandbytes.*` in the resulting
+`metadata.bitsandbytes.*` fields so they do not collide with fields logged
+by other libraries in the shared telemetry index.
+"""
+
+from __future__ import annotations
+
+import logging
+import os
+import platform
+import sys
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+_REPORTED: set[str] = set()
+_FINGERPRINT: Optional[dict[str, str]] = None
+
+_TRUTHY = frozenset({"1", "true", "yes", "on"})
+
+
+def _is_pytest() -> bool:
+    """Detect whether we are running inside a pytest process.
+
+    Telemetry is suppressed during test runs so that CI and local test
+    invocations don't pollute the real-usage stream. Tests that want to
+    assert on telemetry behavior monkey-patch this function to return False.
+    """
+    return "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
+
+
+def _is_disabled() -> bool:
+    for var in ("BNB_DISABLE_TELEMETRY", "HF_HUB_DISABLE_TELEMETRY", "HF_HUB_OFFLINE"):
+        if os.environ.get(var, "").strip().lower() in _TRUTHY:
+            return True
+    if _is_pytest():
+        return True
+    return False
+
+
+def _os_info() -> tuple[str, str]:
+    os_name = platform.system()
+    os_name = {"Darwin": "macOS"}.get(os_name, os_name)
+    if os_name == "Windows":
+        try:
+            build = sys.getwindowsversion().build
+            os_version = f"11 (build {build})" if build >= 22000 else f"10 (build {build})"
+        except Exception:
+            os_version = platform.release()
+    elif os_name == "macOS":
+        os_version = platform.mac_ver()[0] or platform.release()
+    else:
+        os_version = platform.release()
+    return os_name, os_version
+
+
+def _accel_info() -> dict[str, str]:
+    info: dict[str, str] = {}
+    try:
+        import torch
+    except ImportError:
+        info["bitsandbytes.accel"] = "unknown"
+        return info
+
+    try:
+        if torch.cuda.is_available():
+            vendor = "amd" if getattr(torch.version, "hip", None) else "nvidia"
+            info["bitsandbytes.accel"] = vendor
+            info["bitsandbytes.accel_count"] = str(torch.cuda.device_count())
+            props = torch.cuda.get_device_properties(0)
+            info["bitsandbytes.accel_name"] = props.name
+            if vendor == "nvidia":
+                info["bitsandbytes.accel_arch"] = f"sm_{props.major}{props.minor}"
+            else:
+                info["bitsandbytes.accel_arch"] = getattr(props, "gcnArchName", "unknown")
+            return info
+
+        if hasattr(torch, "xpu") and torch.xpu.is_available():
+            info["bitsandbytes.accel"] = "xpu"
+            info["bitsandbytes.accel_count"] = str(torch.xpu.device_count())
+            try:
+                info["bitsandbytes.accel_name"] = torch.xpu.get_device_properties(0).name
+            except Exception:
+                pass
+            return info
+
+        if hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
+            info["bitsandbytes.accel"] = "mps"
+            return info
+
+        if hasattr(torch, "hpu") and torch.hpu.is_available():
+            info["bitsandbytes.accel"] = "hpu"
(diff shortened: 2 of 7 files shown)


bitsandbytes/nn/modules.py
```diff
@@ -607,6 +619,16 @@ def _save_to_state_dict(self, destination, prefix, keep_vars):
...
+        report_feature(
+            "linear_4bit",
+            {
+                "quant_type": getattr(self.weight, "quant_type", "unknown"),
+                "blocksize": getattr(self.weight, "blocksize", 0),
+                "compress_statistics": getattr(self.weight, "compress_statistics", False),
+                "input_dtype": str(x.dtype).replace("torch.", ""),
+                "compute_dtype": (str(self.compute_dtype).replace("torch.", "") if self.compute_dtype else "auto"),
+            },
+        )
```
matthewdouglas, line 631: I would prefer we do this in `__init__` rather than add unnecessary overhead in the `forward()` hot path. Plus most of this is in `__init__` - you don't need to use all these `getattr` calls.

bitsandbytes/nn/modules.py
```diff
@@ -1178,6 +1203,14 @@ def to(self, *args, **kwargs):
...
 
     def forward(self, x: torch.Tensor):
+        report_feature(
+            "linear_8bit",
+            {
+                "has_fp16_weights": self.state.has_fp16_weights,
+                "threshold": self.state.threshold,
+                "input_dtype": str(x.dtype).replace("torch.", ""),
+            },
+        )
```
matthewdouglas, line 1213: Same comment as with `Linear4bit`, this is best in `__init__` and not in `forward`.

bitsandbytes/optim/optimizer.py
```diff
@@ -310,6 +319,13 @@ def step(self, closure=None):
             closure (`Callable`, *optional*, defaults to `None`):
                 A closure that reevaluates the model and returns the loss.
         """
+        report_feature(
+            "optimizer",
+            {
+                "name": type(self).__name__,
+                "is_paged": self.is_paged,
+            },
+        )
```
matthewdouglas, line 328: Would prefer this to be in init of optimizers as well rather than in `step()`. Also maybe would be nice to see `optim_bits`.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
matthewdouglas, line 1: minor nit: copyright here is wrong, let's take this out or replace with more appropriate

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+What is NOT collected
+    Model names, file paths, parameter shapes, user identifiers, training
+    data, gradient values, or any value derived from user input.
+
+Automatically disabled when running under pytest (detected via
+`pytest` in `sys.modules` or `PYTEST_CURRENT_TEST` env var) so that test
+runs in CI and locally do not pollute the real-usage stream.
+
+Opt-out (any of the following env vars disables all telemetry):
+    - BNB_DISABLE_TELEMETRY=1           (bitsandbytes only)
```
matthewdouglas, line 28: I am not sure we need to roll our own; it's cleaner to just reuse the existing? I don't see a use case for e.g. opting out of HF Hub telemetry but still opting in for BNB?

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+    - HF_HUB_DISABLE_TELEMETRY=1        (all HF libraries)
+    - HF_HUB_OFFLINE=1                  (all HF libraries)
+
+End-to-end verification:
+    Set `BNB_TELEMETRY_TAG=<some-id>` before importing bitsandbytes and the
+    value is attached as `bitsandbytes.tag` on every event. Use this to
+    correlate a single run's events in ES.
```
matthewdouglas, line 35: Seems unnecessary/overkill?

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def _is_pytest() -> bool:
+    """Detect whether we are running inside a pytest process.
+
+    Telemetry is suppressed during test runs so that CI and local test
+    invocations don't pollute the real-usage stream. Tests that want to
+    assert on telemetry behavior monkey-patch this function to return False.
+    """
+    return "pytest" in sys.modules or "PYTEST_CURRENT_TEST" in os.environ
```
matthewdouglas, line 67: I would consider looking at other env variables and not bother with the `"pytest" in sys.modules` condition.

Most CI platforms will have an env var like `CI` for this.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+    return False
+
+
+def _os_info() -> tuple[str, str]:
+    os_name = platform.system()
+    os_name = {"Darwin": "macOS"}.get(os_name, os_name)
+    if os_name == "Windows":
+        try:
+            build = sys.getwindowsversion().build
+            os_version = f"11 (build {build})" if build >= 22000 else f"10 (build {build})"
```
matthewdouglas, line 85: This seems fragile and also ignores Windows Server etc.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+            if libc_name:
+                info["bitsandbytes.libc"] = f"{libc_name}-{libc_ver}"
+        except Exception:
+            pass
+    try:
+        import torch
+
+        info["bitsandbytes.torch"] = torch.__version__
+    except ImportError:
+        pass
```
matthewdouglas, line 171: I think this is redundant, does hf hub automatically collect this?

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+            `path_filename` in ES queries).
+        details: Optional feature-specific key/value metadata. Keys without a
+            `bitsandbytes.` prefix are prefixed automatically.
+    """
+    if feature in _REPORTED:
+        return
+    _REPORTED.add(feature)
+
+    if _is_disabled():
+        return
```
matthewdouglas, line 197: We may want to to de-duping more granular than just the "feature" name as it is. But maybe we just name the features differently in that case. So that's more of a minor nit.

Should we add to _REPORTED even when disabled? Seems to me we should just exit right away.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+    user_agent = dict(fingerprint)
+    user_agent["bitsandbytes.feature"] = feature
+    if details:
+        for k, v in details.items():
+            key = k if k.startswith("bitsandbytes.") else f"bitsandbytes.{k}"
+            user_agent[key] = str(v)
+
+    tag = os.environ.get("BNB_TELEMETRY_TAG", "").strip()
+    if tag:
+        user_agent["bitsandbytes.tag"] = tag
```
matthewdouglas, line 214: Same as comment earlier, seems unnecessary.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+            info["bitsandbytes.accel"] = vendor
+            info["bitsandbytes.accel_count"] = str(torch.cuda.device_count())
+            props = torch.cuda.get_device_properties(0)
+            info["bitsandbytes.accel_name"] = props.name
+            if vendor == "nvidia":
+                info["bitsandbytes.accel_arch"] = f"sm_{props.major}{props.minor}"
+            else:
+                info["bitsandbytes.accel_arch"] = getattr(props, "gcnArchName", "unknown")
+            return info
+
```
matthewdouglas, line 115: This only looks at the first device; I'm not sure but we may be interested when there's multiple devices and they're different. I'm wondering if for that maybe we just add some sort of flag to tell us whether it is a heterogeneous system or not. Likely it is, but may be valuable to find out otherwise.

Let's grab device 0's SM count and memory. We don't really need the name. So this should be for both AMD and NVIDIA the `multi_processor_count` and `total_memory` properties. Keep gcnArchName and major/minor.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+    else:
+        os_version = platform.release()
+    return os_name, os_version
+
+
+def _accel_info() -> dict[str, str]:
+    info: dict[str, str] = {}
+    try:
+        import torch
+    except ImportError:
```
matthewdouglas, line 99: torch is already a pretty hard dep, this shouldnt need to be caught

bitsandbytes/nn/modules.py
```diff
@@ -239,6 +242,15 @@ def __new__(
...
         self.data = data
         self.module = module
+        report_feature(
+            "params_4bit",
+            {
+                "quant_type": quant_type,
+                "blocksize": blocksize,
+                "compress_statistics": compress_statistics,
+                "quant_storage": str(quant_storage).replace("torch.", ""),
+            },
```
matthewdouglas, line 252: Starting to think we don't need this here, would prefer we just keep Linear4bit and Linear8bitLt but remove this on Params4bit/Int8Params.

bitsandbytes/optim/optimizer.py
```diff
@@ -104,13 +105,21 @@ def override_config(self, parameters, key=None, value=None, key_value_dict=None)
             key_value_dict = {key: value}
 
         if key_value_dict is not None:
+            report_feature(
+                "optim_override_config",
+                {"keys": ",".join(sorted(key_value_dict.keys()))},
+            )
```
matthewdouglas, line 111: Not particularly interested in tracking this

bitsandbytes/optim/optimizer.py
```diff
@@ -104,13 +105,21 @@ def override_config(self, parameters, key=None, value=None, key_value_dict=None)
...
                 if id(p) in self.pid2config:
                     self.pid2config[id(p)].update(key_value_dict)
                 else:
                     self.pid2config[id(p)] = key_value_dict
 
     def register_module_override(self, module, param_name, config):
+        report_feature(
+            "optim_register_module_override",
+            {"keys": ",".join(sorted(config.keys())) if isinstance(config, dict) else "unknown"},
+        )
```
matthewdouglas, line 122: Likewise not really interested in tracking this either

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+        pass
+
+    info.update(_accel_info())
+
+    _FINGERPRINT = info
+    return info
+
+
+def report_feature(feature: str, details: Optional[dict[str, object]] = None) -> None:
```
matthewdouglas, line 179: I think for more clarity we should just name this `_report_feature` as well.

bitsandbytes/_telemetry.py
```diff
@@ -0,0 +1,231 @@
...
+    except Exception:
+        version = "unknown"
+
+    os_name, os_version = _os_info()
+    info = {
+        "bitsandbytes.version": version,
+        "bitsandbytes.os": os_name,
+        "bitsandbytes.os_version": os_version,
+        "bitsandbytes.arch": platform.machine(),
+        "bitsandbytes.python": platform.python_version(),
```
matthewdouglas, line 157: I think this is redundant too, huggingface_hub likely includes Python version already

README.md
```diff
@@ -183,6 +183,43 @@ bitsandbytes has the following minimum requirements for all platforms:
...
+| ---------------------------- | ---------------------------- |
+| `BNB_DISABLE_TELEMETRY=1`    | `bitsandbytes` only          |
+| `HF_HUB_DISABLE_TELEMETRY=1` | all Hugging Face libraries   |
+| `HF_HUB_OFFLINE=1`           | all Hugging Face libraries   |
+
+Telemetry is also automatically suppressed while running under `pytest` (so
+CI and local test runs don't pollute the stream) and a silent no-op when
+`huggingface_hub` is not installed. The implementation lives in
+[`bitsandbytes/_telemetry.py`](bitsandbytes/_telemetry.py) and each event
+fires at most once per process.
```
matthewdouglas, line 221: Let's simplify this:

```suggestion
## Telemetry

bitsandbytes collects anonymous feature-usage data using the same telemetry
mechanism as other Hugging Face libraries (Transformers, Gradio, etc.). This
helps us understand which features are actively used so we can prioritize
maintenance and make informed decisions about deprecation.

### What is collected

Hardware and version info sent once per process (bitsandbytes version, OS, CPU
architecture, accelerator type and compute capability) plus one event per
distinct feature used per process.

### How to opt out

Set any of the following environment variables:

| Variable                     | Effect                              |
| ---------------------------- | ----------------------------------- |
| `HF_HUB_DISABLE_TELEMETRY=1` | Disables telemetry in all HF libs   |
| `HF_HUB_OFFLINE=1`           | Disables all outbound HF Hub calls  |
| `DO_NOT_TRACK=1`             | Standard cross-tool opt-out signal  |

```

PR #1949, opened 2026-05-15
[CUDA] New 4bit GEMM kernels for inference

matthewdouglas:
## TL;DR
4-bit inference is up to 4x faster at batch serving sizes (i.e. 2-64) across GPUs from Turing through Blackwell. Gains are largest at smaller batches. In many cases the new kernel is faster at batch size of 1 also. We also see improvements when using nested quantization, where the fusion in the new kernels provides an advantage.

## Summary
This PR adds new custom fused 4-bit dequantize+GEMM kernels. It is intended to replace both the existing batch size 1 path (GEMV kernel) and the`dequantize + F.linear` path for small-to-medium batch sizes. The new kernels additionally fuse the nested absmax dequantization and optional bias addition.

### New kernels
There are three new kernel tiers, selected automatically at runtime via heuristics calibrated with benchmarks on T4, A10, A100, RTX 4090, L4, L40S, H100, H200, B200, and RTX PRO 6000. The heuristics are expected to generalize well across most hardware on sm75+. Note that while the SIMT kernel does compile and should work on Pascal and Volta, we did not perform benchmarks there and will dispatch to it more conservatively.

Claude Code was used to create a simulation tool to help calibrate these heuristics. We lean towards conservatively dispatching the new kernel to avoid any performance regressions, so across many shapes we may miss some wins by falling back earlier than strictly necessary. Across all of the GPU testing on over 40 shapes and batch sizes from 1 up through 2048, we make a regressive decision less than 1% of the time, and make the most ideal decision over 80% of the time. For the remaining decisions, we simply left performance on the table by either falling back too early, or selecting a kernel where we had a faster alternative available.

| Kernel | Target | dtypes |
|--------|--------|--------|
| SIMT | sm60+ (all GPUs) | bf16, fp16, fp32 |
| MMA `m16n8k8` | sm75 (Turing, e.g. T4) | fp16 only |
| MMA `m16n8k16` | sm80+ (Ampere, Ada, Hopper, Blackwell) | bf16, fp16 |

The dispatch system considers problem dimensions, GPU architecture, and SM count when factoring in the decisions on which kernel to launch. Within each of the MMA kernels, there are several tiling configurations for the dispatcher to choose from. For larger batch sizes, we will automatically fall back to the existing `dequantize + F.linear` path.

### New custom op layer
Added a new custom operator `bitsandbytes::gemm_4bit` to abstract this new operator. It is backwards compatible with other backends, i.e. they will continue to launch their existing GEMV implementations. Note that this operator is not intended as a public API and can change in the future. It serves as an extension point to add implementations for additional hardware.

### Additional cleanup
`bitsandbytes.matmul_4bit` now normalizes packed weights to a canonical shape internally, so callers no longer need to pass the quantized weight tensor in any particular orientation. For weights quantized in `[out_features, in_features]` orientation passing a `.t()` of the quantized weight tensor is fine, although no longer required, and will produce correct results. However, weights that were quantized in transposed `[in_features, out_features]` orientation will now emit a `DeprecationWarning`. Support for this is likely to be dropped in the future. This is not an expected or typical use case.
  
## End-to-End Benchmark Results

All runs: NF4, input_len=128, output_len=128, bitsandbytes 0.49.2 (stable) vs this PR (new).
PyTorch: 2.10.0+cu130 in eager mode. Transformers: 5.7.0.
All hardware except RTX 4090 was hosted on Modal.com.

TPOT = time per output token (decode step), lower is better.

tpot_4090_qwen3_8b

tpot_4090_qwen36_27b

tpot_a100_qwen36_27b

tpot_rtxpro6000_qwen36_27b

tpot_t4_qwen35_9b


matthewdouglas:
The new kernels will not be typically dispatched for QLoRA traing, with exception of small token batches on smaller shapes. Any call with > 1536 tokens is always dispatched to the same code path as before, and for most GPUs and shapes > 128 tokens will be enough to fall back.

In test QLoRA runs, there was identical peak memory usage, training loss, and evaluation loss. There was however a slight decrease in training time noted as a result of a reduction of CPU host overhead. This reduction will vary largely on CPU, GPU, problem sizes, etc. There's another follow-up PR planned separately for further CPU overhead reduction.

pjordanandrsn:
Hi @matthewdouglas — building on the `gemm_4bit` kernel family from #1949, I've
been prototyping an **expert-grouped** variant for fused MoE expert stacks and
wanted to coordinate before proposing anything, in case it overlaps your
roadmap.

The shape: a single-launch grouped W4A16 GEMM that takes the experts as one
`[E, N, K/2]` NF4 stack plus `(sizes, expert_ids)` after the token→expert
sort, and decodes NF4 in the mainloop (LUT→fp32 in registers, blockwise fp32
absmax, fp32 accumulate) — same codebook and absmax conventions as #1949, same
"sort/scatter live outside the op" contract as any grouped GEMM. On sm_86 at
batch-1 decode it runs ~1.2–2.7× (median) a dequantize-then-matmul baseline
across the OLMoE / Qwen3-30B / Gemma / GPT-OSS expert shapes, with J/token
below the baseline on the bandwidth-bound cells and output fidelity ordered
above the materialize-to-bf16 path (fp32 accumulation). Caveats stated
up-front: one census shape is instance-sensitive, top_k=1 and tiny shapes
lose (a dispatch floor routes those back to the dequant path), and the
prefill/compute-bound tile isn't at cuBLAS parity (down-projections ~1.9×
after a config pass; gate_up still below). Everything — including three
pre-registered blind confirmatory runs that each caught real problems — is
receipted in the repo:
https://github.com/pjordanandrsn/grouped-nf4-gemm

Two coordination questions:

1. Is a grouped/MoE variant something you'd want in `bitsandbytes` proper (as
   an extension of the #1949 op family), or is it better as a downstream
   kernel that just consumes the packed layout and conventions?
2. If in-tree is interesting, what would you want to see — a specific op
   signature, the tier structure matching #1949 (SIMT + MMA), a particular
   fidelity/perf bar? Happy to write it to your conventions rather than guess.

No rush and no expectation — I'd rather build to your design than surprise you
with a PR. Also independent of the parametrize-cache fix I filed as #1999 (a
different subsystem).

Thanks for #1949 — it's a clean base to extend.

PR #864, opened 2023-11-08
fixes to recent PR #753

poedator:
Save and load in NF4 / FP4 formats 
This fixes problems arising from my recent pull request #753 , mainly coming from [this commit](https://github.com/TimDettmers/bitsandbytes/pull/753/commits/76b40a5c9ae708db98e8b4a13249b2806601a387): 

1) The idea to override `_load_from_state_dict` for `Linear4bit` looked very elegant, with a view to simplify Transformers integration. However it has more sensitivities than I expected including one with `peft`.

2) replaced `|=` operator with `.update()` to keep compatibility with Python 3.8

this should fix https://github.com/huggingface/peft/issues/1095

@TimDettmers @Titus-von-Koeller please review.

PR #1958, opened 2026-05-28
fix(nn/Linear4bit): consume QuantState keys in _load_from_state_dict (#1946)

### LINKED ISSUE #1946: Linear4bit._save_to_state_dict writes QuantState keys but no _load_from_state_dict consumes them (asymmetric serialization)
### Summary

`bitsandbytes.nn.Linear4bit` overrides `_save_to_state_dict` (`bitsandbytes/nn/modules.py:593`) to write QuantState components alongside the packed weight:

```python
def _save_to_state_dict(self, destination, prefix, keep_vars):
    super()._save_to_state_dict(destination, prefix, keep_vars)  # weight + bias
    if getattr(self.weight, "quant_state", None) is not None:
        for k, v in self.weight.quant_state.as_dict(packed=True).items():
            destination[prefix + "weight." + k] = v if keep_vars else v.detach()
```

Resulting state_dict keys for a Linear4bit (with `compress_statistics=True`):
- `weight` (packed 4-bit tensor)
- `weight.absmax`
- `weight.quant_map`
- `weight.nested_absmax`
- `weight.nested_quant_map`
- `weight.quant_state.bitsandbytes__nf4` (or `__fp4`)

`Linear4bit` does **not** define a corresponding `_load_from_state_dict`. It inherits `nn.Linear._load_from_state_dict`, which only consumes `weight` and `bias`. The QuantState keys land in `unexpected_keys` during load. With `strict=True` (the `model.load_state_dict()` default) this raises `RuntimeError`.

This is asymmetric relative to `Linear8bitLt`, which defines both `_save_to_state_dict` (line 1095) **and** `_load_from_state_dict` (line 1119), with the latter explicitly walking `unexpected_keys` to consume the SCB tensor and remove it from the unexpected list.

### Reproducer

```python
import torch
import bitsandbytes as bnb

src = bnb.nn.Linear4bit(64, 64, bias=False, quant_type='nf4',
                       compute_dtype=torch.bfloat16,
                       compress_statistics=True)
src = src.to('cuda')  # triggers quantize, populates quant_state

sd = src.state_dict()
print("state_dict keys:", list(sd.keys()))
# ['weight', 'weight.absmax', 'weight.quant_map', 'weight.nested_absmax',
#  'weight.nested_quant_map', 'weight.quant_state.bitsandbytes__nf4']

dst = bnb.nn.Linear4bit(64, 64, bias=False, quant_type='nf4',
                       compute_dtype=torch.bfloat16,
                       compress_statistics=True)
dst = dst.to('cuda')

# Fails with strict=True (the default):
dst.load_state_dict(sd)
# RuntimeError: Error(s) in loading state_dict for Linear4bit:
#     Unexpected key(s) in state_dict: "weight.absmax", "weight.quant_map",
#     "weight.nested_absmax", "weight.nested_quant_map",
#     "weight.quant_state.bitsandbytes__nf4"
```

With `strict=False`, the QuantState components are silently ignored: `dst.weight.data` becomes the packed bytes from `src`, but `dst.weight.quant_state` is the one populated during `dst.to('cuda')` — not the one from `sd`. If `src` and `dst` quantized with different inputs (very common, e.g. loading a checkpoint into a freshly-initialized module), `dst` is left in a silently-corrupt state: packed bytes from one model, QuantState scalars from another.

### Why this hasn't been reported as a user-facing bug

Standard workflows bypass `model.load_state_dict()` for bnb-quantized checkpoints:

1. **HF Transformers `from_pretrained` for pre-quantized bnb checkpoints:** uses `bnb.nn.Params4bit.from_prequantized(data, quantized_stats, device=...)` directly via `transformers/quantizers/quantizer_bnb_4bit.py::create_quantized_param`. The `quantized_stats` dict is built by walking state_dict keys with prefix `param_name + "."`. Bypasses `_load_from_state_dict` entirely.
2. **PEFT save/load:** `get_peft_model_state_dict` filters to LoRA-only keys (`"lora_" in k`); QuantState keys never enter the PEFT round-trip.
3. **Accelerate's `fsdp2_load_full_state_dict`:** has its own broadcast + `assign=True` path; PR #3982 added key-based matching that filters non-Params keys.

Paths that **do** hit the asymmetry:
- `torch.distributed.checkpoint` (DCP) when the loaded state_dict contains QuantState keys
- Custom training loops that save via `model.state_dict()` and resume via `model.load_state_dict()`
- The docstring example at `nn/modules.py:522` (`quantized_model.load_state_dict(fp16_model.state_dict())`) works only because the source is **fp16** (no QuantState keys to be unexpected) — masking the bnb→bnb round-trip case

### Discovery context

Surfaced while characterizing FSDP2 + `Params4bit` forward correctness on Jetson Orin Nano Super (sm_87) — the same investigation that led to #1945. The investigation needed to round-trip QuantState through `state_dict()` for a pre-shard-broadcast pattern; that worked on the save side and failed on the load side, leading to this finding. The two issues are unrelated mechanically — this one is a missing override, #1945 is a PyTorch FSDP2 NaN-canonicalization bug — but they share discovery provenance.

### Suggested fix

Implement `_load_from_state_dict` paralleling `Linear8bitLt`'s pattern. Sketch:

```python
def _load_from_state_dict(
    self, state_dict, prefix, local_metadata, strict,
    missing_keys, unexpected_keys, error_msgs,
):
    # Collect QuantState components in state_dict for this prefix
    qs_keys_to_consume = []
    quantized_stats = {}
    weight_dot_prefix = prefix + "weight."
    for k in list(state_dict.keys()):
        if k.startswith(weight_dot_prefix):
            qs_keys_to_consume.append(k)
            # store as `weight.<subkey>` for from_prequantized
            quantized_stats[k[len(prefix):]] = state_dict[k]

    # Standard nn.Linear path consumes 'weight' and 'bias'
    super()._load_from_state_dict(
        state_dict, prefix, local_metadata, strict,
        missing_keys, unexpected_keys, error_msgs,
    )

    # If QuantState components are present, reconstruct via from_prequantized
    if quantized_stats:
        for k in qs_keys_to_consume:
            if k in unexpected_keys:
                unexpected_keys.remove(k)

        weight_data = self.weight.data  # already loaded by super()
        self.weight = Params4bit.from_prequantized(
            data=weight_data,
            quantized_stats=quantized_stats,
            requires_grad=False,
            device=weight_data.device,
            module=self,
        )
```

Edge cases:
- If QuantState keys are absent (e.g., the existing fp16→bnb docstring example), `quantized_stats` is empty and the function falls through to standard `nn.Linear` behavior — no regression.
- The `weight_dot_prefix` match cleanly delimits per-Linear keys; the state_dict passed to `_load_from_state_dict` is already filtered to the current module's prefix by `nn.Module.load_state_dict()`'s recursion.

### Test plan (for the fix)

- [ ] Bnb→bnb round-trip: `m1.state_dict() → m2.load_state_dict(strict=True)` on CPU then CUDA Linear4bit; assert `bnb.matmul_4bit` produces identical output on synthetic input.
- [ ] Existing fp16→bnb docstring example continues to pass (no regression).
- [ ] Variant matrix: NF4 / FP4, `compress_statistics` on/off, `quant_storage` ∈ {uint8, bf16, fp16}, with/without bias.
- [ ] `strict=False` behavior with partial QuantState keys (fall back to existing `_quantize` flow on `.to(device)` if applicable).

### Severity

**Latent — currently masked.** Does not surface in standard HF + PEFT workflows because they bypass `load_state_dict`. But it's a contract violation between `_save_to_state_dict` and the standard `nn.Module` load mechanism, and surfaces for:
- `torch.distributed.checkpoint` (DCP) usage
- Custom training-loop checkpoint code
- Any tooling that round-trips through `model.state_dict()` + `model.load_state_dict()`

### Adjacent PRs (for context, do not address this)

- PR #1866 — `Params4bit.__getattr__` proxy for FSDP `_get_fqns()` traversal (state_dict *traversal*, not load).
- PR #1916 — replaces #1866's `__getattr__` with `@property` descriptors (motivated by `torch.compile` graph breaks, preserved FSDP traversal as a side effect).

Neither addresses the missing `_load_from_state_dict`.

Happy to file a PR with the fix sketch + tests if useful.

Anai-Guo:
Fixes #1946.

### Problem

`Linear4bit._save_to_state_dict` writes the packed `weight` **and** the QuantState components:

- `weight.absmax`, `weight.quant_map`
- `weight.nested_absmax`, `weight.nested_quant_map` (when `compress_statistics=True`)
- `weight.quant_state.bitsandbytes__nf4` / `__fp4`

But `Linear4bit` does not override `_load_from_state_dict`; it inherits `nn.Linear._load_from_state_dict`, which only consumes `weight` and `bias`. So:

- `load_state_dict(strict=True)` raises `RuntimeError: Unexpected key(s) in state_dict` for every QuantState entry.
- `load_state_dict(strict=False)` silently drops them. The destination layer keeps the freshly-quantized `quant_state` from the prior `.to('cuda')` call, which does not match the packed bytes that were just loaded — so forward passes return garbage. This is asymmetric vs `Linear8bitLt`, which already implements both halves (save at `modules.py:1095`, load at `modules.py:1119`).

### Fix

Add `Linear4bit._load_from_state_dict`. After delegating to `super()`, walk `unexpected_keys` for entries under `<prefix>weight.`, collect them into a `qs_dict`, rebuild via `QuantState.from_dict(...)`, install on `self.weight`, and remove the consumed keys from `unexpected_keys`. Mirrors the existing `Linear8bitLt` pattern.

### Reproducer (from the issue, abbreviated)

```python
import bitsandbytes as bnb

src = bnb.nn.Linear4bit(64, 64, bias=False, quant_type='nf4', compute_dtype=torch.bfloat16, compress_statistics=True).to('cuda')
dst = bnb.nn.Linear4bit(64, 64, bias=False, quant_type='nf4', compute_dtype=torch.bfloat16, compress_statistics=True).to('cuda')

dst.load_state_dict(src.state_dict())
# before: RuntimeError: Unexpected key(s) ... weight.absmax, weight.quant_map, ...
# after:  dst.weight.quant_state matches src.weight.quant_state
```

### Notes

- Behavior is unchanged for state_dicts that only contain `weight` (and `bias`): when no `qs_prefix` keys remain in `unexpected_keys`, the new code is a no-op early-return.
- No public API is changed; this only adds a private hook that PyTorch's `load_state_dict` machinery already expects modules to implement when they extend `_save_to_state_dict`.
- `Embedding4bit` has the same issue but a separate `_save_to_state_dict` shape; happy to extend the fix there in a follow-up if maintainers prefer one PR per class.

### DIFF
--- a/bitsandbytes/nn/modules.py
+++ b/bitsandbytes/nn/modules.py
@@ -606,6 +606,66 @@ class Linear4bit(nn.Linear):
             for k, v in self.weight.quant_state.as_dict(packed=True).items():
                 destination[prefix + "weight." + k] = v if keep_vars else v.detach()
 
+    def _load_from_state_dict(
+        self,
+        state_dict,
+        prefix,
+        local_metadata,
+        strict,
+        missing_keys,
+        unexpected_keys,
+        error_msgs,
+    ):
+        """
+        Consume the QuantState components written by ``_save_to_state_dict``.
+
+        Without this, the ``weight.absmax`` / ``weight.quant_map`` / ... keys
+        land in ``unexpected_keys`` and ``load_state_dict(strict=True)`` raises;
+        with ``strict=False`` they are silently dropped and the loaded layer
+        keeps a freshly-quantized ``quant_state`` that does not match the
+        packed bytes that were just loaded.
+        """
+        super()._load_from_state_dict(
+            state_dict,
+            prefix,
+            local_metadata,
+            strict,
+            missing_keys,
+            unexpected_keys,
+            error_msgs,
+        )
+
+        qs_prefix = prefix + "weight."
+        qs_dict: dict[str, torch.Tensor] = {}
+        consumed_keys: list[str] = []
+        for key in list(unexpected_keys):
+            if not key.startswith(qs_prefix):
+                continue
+            qs_dict[key[len(qs_prefix):]] = state_dict[key]
+            consumed_keys.append(key)
+
+        if not qs_dict:
+            return
+
+        try:
+            quant_state = QuantState.from_dict(qs_dict=qs_dict, device=self.weight.device)
+        except Exception as exc:
+            error_msgs.append(
+                f"Linear4bit: failed to reconstruct QuantState from state_dict "
+                f"with prefix '{prefix}': {exc}"
+            )
+            return
+
+        self.weight.quant_state = quant_state
+        self.weight.bnb_quantized = True
+        self.weight.blocksize = quant_state.blocksize
+        self.weight.compress_statistics = quant_state.nested
+        self.weight.quant_type = quant_state.quant_type
+        self.quant_state = quant_state
+
+        for key in consumed_keys:
+            unexpected_keys.remove(key)
+
     def forward(self, x: torch.Tensor):
         fix_4bit_weight_quant_state_from_module(self)
         quant_state = self.weight.quant_state
--- a/tests/test_linear4bit.py
+++ b/tests/test_linear4bit.py
@@ -615,3 +615,54 @@ def test_fsdp_state_dict_save_4bit():
             f"stdout: {result.stdout}\n"
             f"stderr: {result.stderr}"
         )
+
+
+class _Linear4bitWrapper(torch.nn.Module):
+    """Nest Linear4bit under a prefix so load_state_dict exercises the real path."""
+
+    def __init__(self, quant_type, compress_statistics):
+        super().__init__()
+        self.fc = bnb.nn.Linear4bit(
+            64,
+            64,
+            bias=False,
+            compute_dtype=torch.float32,
+            compress_statistics=compress_statistics,
+            quant_type=quant_type,
+        )
+
+
+@pytest.mark.parametrize("device", get_available_devices())
+@pytest.mark.parametrize("quant_type", ["nf4", "fp4"])
+@pytest.mark.parametrize("compress_statistics", TRUE_FALSE, ids=id_formatter("compress_statistics"))
+@pytest.mark.parametrize("strict", TRUE_FALSE, ids=id_formatter("strict"))
+def test_linear4bit_load_from_state_dict(device, quant_type, compress_statistics, strict):
+    """Regression test for load_state_dict silently dropping QuantState keys.
+
+    The existing serialization tests restore via ``from_prequantized`` and never
+    call ``load_state_dict``, so the ``_load_from_state_dict`` override was
+    uncovered. Without the fix, the packed ``weight`` bytes load but the stale
+    ``quant_state`` is kept: ``strict=True`` raises on unexpected keys and
+    ``strict=False`` silently produces wrong outputs. Fails on main, passes here.
+    """
+
+    def build(seed):
+        torch.manual_seed(seed)
+        net = _Linear4bitWrapper(quant_type, compress_statistics)
+        with torch.no_grad():
+            net.fc.weight.data = torch.randn(64, 64) * 0.1
+        return net.to(device)  # triggers 4-bit quantization
+
+    src = build(0)
+    dst = build(999)  # different weights, so a no-op load would be caught
+
+    result = dst.load_state_dict(src.state_dict(), strict=strict)
+
+    assert result.missing_keys == []
+    assert result.unexpected_keys == []
+    assert dst.fc.weight.quant_state is not None
+    assert torch.equal(src.fc.weight.quant_state.absmax, dst.fc.weight.quant_state.absmax)
+
+    x = torch.randn(8, 64, device=device)
+    with torch.no_grad():
+        assert torch.equal(src.fc(x), dst.fc(x))


matthewdouglas:
@SunMarc Could you help me understand if this is reasonable from the transformers/accelerate integration side?

egeozkoc:
I spent some time testing this against the current HuggingFace stack to help answer the transformers/accelerate question, since that (plus the peft sensitivity) is what led to the original 2023 version being walked back in #864. Posting the results in case they help.

**Environment:** `transformers==4.57.6`, `peft==0.19.1`, `accelerate==1.14.0`, torch 2.12, CPU backend. Tiny model: `hf-internal-testing/tiny-random-LlamaForCausalLM` (14 `Linear4bit` layers).

### What the bug actually costs

One thing worth flagging: the impact is broader than the top-level `Linear4bit` in the issue repro. It hits nested modules too, so basically any real model. With a `Linear4bit` nested under a prefix and no override:

| load | result |
| --- | --- |
| `strict=True` | raises `RuntimeError` (5 QuantState keys unexpected), and forward output is garbage |
| `strict=False` | loads "successfully", silently drops the 5 keys, forward output is garbage |

The `strict=False` case is the nasty one. The packed `weight` bytes get loaded but `quant_state` stays the stale one, so there's no error and the outputs are just wrong.

### Integration paths (with this PR applied)

| Scenario | `_load_from_state_dict` invoked? | Result |
| --- | --- | --- |
| `from_pretrained(load_in_4bit=True)` | no (goes through the accelerate meta path) | works |
| `save_pretrained` then `from_pretrained` (serialized 4-bit) | no | logits match exactly |
| PEFT LoRA on 4-bit base, `save_pretrained` then `PeftModel.from_pretrained` | yes (per layer), 0 errors | logits match exactly |
| plain `nn.Module.load_state_dict` (nested), `strict` True/False | yes | logits match exactly |
| `tests/test_linear4bit.py -k "serialization or copy or deepcopy"` | n/a | 164 passed |

So on current versions the transformers and PEFT load paths route 4-bit weights through their own mechanisms, and they either skip this override entirely or hit it as a safe no-op. I couldn't get it to error or change behavior in any of those flows, and the peft#1095-style breakage doesn't seem to reproduce anymore. The override only does real work on the plain `load_state_dict` path, which is where the bug is.

One thing I couldn't rule out: a manual `strict` load onto a meta-device 4-bit model would run `QuantState.from_dict(device=self.weight.device)` against a meta device. I couldn't reach that through any normal transformers/PEFT flow (0 override calls there), so it's just untested rather than known good. Might be worth a guard or at least a note.

### Suggested regression test

The existing `test_linear_serialization` restores via `from_prequantized`, so it never actually calls `load_state_dict`, which is probably why this went unnoticed. Here's a small test that covers the broken path. It fails on `main` and passes with this PR (I checked, 8/8 both ways):

```python
class _Wrapper(torch.nn.Module):
    def __init__(self, quant_type, compress_statistics):
        super().__init__()
        self.fc = bnb.nn.Linear4bit(
            64, 64, bias=False, compute_dtype=torch.float32,
            compress_statistics=compress_statistics, quant_type=quant_type,
        )

@pytest.mark.parametrize("device", get_available_devices())
@pytest.mark.parametrize("quant_type", ["nf4", "fp4"])
@pytest.mark.parametrize("compress_statistics", TRUE_FALSE, ids=id_formatter("compress_statistics"))
@pytest.mark.parametrize("strict", TRUE_FALSE, ids=id_formatter("strict"))
def test_linear4bit_load_from_state_dict(device, quant_type, compress_statistics, strict):
    def build(seed):
        torch.manual_seed(seed)
        net = _Wrapper(quant_type, compress_statistics)
        with torch.no_grad():
            net.fc.weight.data = torch.randn(64, 64) * 0.1
        return net.to(device)  # triggers 4-bit quantization

    src = build(0)
    dst = build(999)  # different weights, so a no-op load is caught
    result = dst.load_state_dict(src.state_dict(), strict=strict)

    assert result.missing_keys == []
    assert result.unexpected_keys == []
    assert dst.fc.weight.quant_state is not None
    assert torch.equal(src.fc.weight.quant_state.absmax, dst.fc.weight.quant_state.absmax)

    x = torch.randn(8, 64, device=device)
    with torch.no_grad():
        assert torch.equal(src.fc(x), dst.fc(x))
```

Happy to open a follow-up PR with this test (and a meta-device check) if that would help.

Anai-Guo:
Thanks @egeozkoc, this is really helpful — especially confirming that the transformers/PEFT/accelerate load paths route 4-bit weights through their own mechanisms and hit this override either not at all or as a safe no-op, so the peft#1095-style breakage from #864 doesn't reproduce on current versions. That was exactly the open question.

I've added your regression test to `tests/test_linear4bit.py` (credited to you in the commit message). It nests `Linear4bit` under a prefix so `load_state_dict` actually runs the override, uses different source/destination weights so a no-op load is caught, and checks both `strict` values — fails on `main`, passes here.

Good call on the meta-device `strict` load; I left it out for now since it isn't reachable through any normal transformers/PEFT flow (0 override calls there), so it'd be testing an untested-but-unused path. Happy to add a guard + note in a follow-up if maintainers prefer.

PR #1965, opened 2026-06-05
Add Experts4bit for 4-bit quantization of fused MoE experts

### LINKED ISSUE #1849: Failed to quant MoE models with fused expert weights in transformers v5
### Description
At transformer v4, experts were implemented using nn.ModuleList of nn.Linear layers. However, starting from transformer v5, newer efficient implementations (likely optimizing for Grouped GEMM or CUDA kernels) are moving towards storing all expert weights in a single fused nn.Parameter tensor (e.g., with shape [num_experts, in_features, out_features]) instead of separate nn.Linear modules.

One typical model is Qwen3MoeForCausalLM, with expert implementations shown as below:
```python
@use_experts_implementation
class Qwen3MoeExperts(nn.Module):
    """Collection of expert weights stored as 3D tensors."""

    def __init__(self, config):
        super().__init__()
        self.num_experts = config.num_experts
        self.hidden_dim = config.hidden_size
        self.intermediate_dim = config.moe_intermediate_size
        self.gate_up_proj = nn.Parameter(torch.empty(self.num_experts, 2 * self.intermediate_dim, self.hidden_dim))
        self.down_proj = nn.Parameter(torch.empty(self.num_experts, self.hidden_dim, self.intermediate_dim))
        self.act_fn = ACT2FN[config.hidden_act]
```

It's worth noticing `use_experts_implementation` allow switching MOE implementation to the more efficent group_gemm version. Bitsandbytes library may also consider implementing a grouped forward version to improve speed on Moe models.

### Reproduction
transformers 5.0.0rc3 should be used.

```python
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer, BitsAndBytesConfig

bnb_config = BitsAndBytesConfig(
    load_in_4bit=True,
    bnb_4bit_compute_dtype=torch.float16,
    bnb_4bit_quant_type="nf4",
    bnb_4bit_use_double_quant=True
)

model_path = "Qwen/Qwen3-30B-A3B"

    model_path,
    quantization_config=bnb_config,
    device_map="auto",
    trust_remote_code=True,
)

print(f"Memory: {model.get_memory_footprint() / 1024**3:.2f} GB")
```
It prints Memory: 55.60 GB, showing quantizing is not working correctly.

### Expected behavior
As in transformer v4, it prints Memory: 15.09 GB, which is properly quantized. 

###  Fix proposal
I've opened a feature request in transformers to address the root cause of this issue: https://github.com/huggingface/transformers/issues/43472.

Instead of handling raw parameters, I proposed introducing a standardized BatchLinear module. If adopted, downstream libraries would only need to support replacing this specific module type, ensuring compatibility for Qwen3-MoE, DeepSeek, and future MoE architectures without model-specific hacks.

Upvoting or commenting on that RFC would help prioritize a unified solution and significantly reduce the maintenance burden for bitsandbytes.

pjordanandrsn:
### Problem
transformers v5 stores MoE expert weights as a single fused 3-D `nn.Parameter`
(`[num_experts, out_features, in_features]`, e.g. `OlmoeExperts`, `Qwen3MoeExperts`) instead of an
`nn.ModuleList` of `nn.Linear`. bitsandbytes' 4-bit conversion replaces `nn.Linear` modules, so these
fused experts — the dominant share of an MoE's weights — are silently skipped and stay full precision
(#1849).

### What this adds
`bitsandbytes.nn.Experts4bit`: a small `nn.Module` holding the two fused expert projections
(`gate_up_proj`, `down_proj`) in 4-bit NF4/FP4.

- **Plain `nn.Parameter` (uint8) + `absmax` buffers, not `Params4bit`.** A 3-D per-expert stack doesn't
  fit `Params4bit`'s tensor-subclass / device-movement machinery cleanly; keeping packed bytes +
  per-expert `absmax` as ordinary tensors means the module serializes through the standard `state_dict`
  with **no custom save/load hooks** (`safetensors.torch.save_model`/`load_model` work out of the box —
  pinned by a test).
- **Per-expert quantization** (`from_float` loops over experts, `quantize_4bit` each). `in_features` must
  be a multiple of `blocksize` so blocks never straddle an expert boundary (validated).
- **Forward** dequantizes one expert at a time and runs the SwiGLU expert MLP, mirroring the reference
  fused-experts forward. It restores the `[packed, 1]` layout `quantize_4bit` emits so
  `dequantize_4bit`'s transpose back-compat shim (keyed on `A.shape[0] == 1`) doesn't fire.
- **Training memory — re-dequantize in backward instead of saving.** Every expert projection routes
  through `_FrozenLinearRecomputeBackward`: forward *is* dequantize + `F.linear`; backward re-dequantizes
  the frozen weight to form `grad_output @ W`. The dequantized `[out, in]` weight never enters autograd's
  saved-tensor storage, so training activation memory stays independent of the number of experts held
  between forward and backward. Bit-exact by construction — recomputation changes what is *saved*, never
  what is *computed* (pinned at `rtol=0/atol=0`, in and out of grad mode). Cost: one extra dequantize per
  projection in backward; subsumed by full gradient checkpointing when enabled.
- **Frozen base:** packed weights are `requires_grad=False`; the forward is differentiable w.r.t. the
  input activations, so it drops in as the frozen base of a QLoRA setup (a per-expert LoRA example lives
  in `examples/experts4bit_qlora_demo.py`).
- **Dtype-cast safety:** `.to(dtype)` / `.half()` / `.bfloat16()` retarget `compute_dtype`; the fp32
  `absmax`/codebook state is shielded from the cast (a stock `nn.Module` cast would silently degrade
  every subsequent dequantization — pinned by a test).

### Correctness & coverage — `tests/test_experts4bit.py`, 53 tests, CPU + CUDA (RTX A2000)
- **Forward** matches a full-precision fused-experts reference built from the same dequantized weights
  (float32, isolating routing from quant error, `1e-4`), and is **bit-exact** vs plain
  dequantize-then-`linear` (`rtol=0/atol=0`).
- **Backward** matches the reference; the frozen base receives **no** gradient; `saved_tensors_hooks`
  proves no weight-shaped tensor (either orientation) is saved while a plain dequantize+linear control
  does save one, with gradients matching the control exactly.
- **Device movement:** `.to()` carries packed weights, `absmax`, and codebook together; cpu→cuda→cpu
  round-trips bit-exact (bytes, scales, and forward).
- **Serialization:** `state_dict` round-trip bit-exact, strict and non-strict into a ctor-built module;
  `safetensors` `save_model`/`load_model` round-trip with bit-exact forward parity.
- **CUDA numerics:** per-expert nf4 dequant mean-abs error ≤ 0.008 at the test file's weight scale
  (measures ~0.0073 on an RTX A2000); forward rel err ≤ 0.2 vs the true-weight reference.
- **#1849 regression:** a fused 3-D `nn.Parameter` expert module (no `nn.Linear` for the walker to catch)
  is actually 4-bit-quantized by `from_float` — uint8-packed, >3× smaller than fp16.
- **End-to-end:** a frozen `Experts4bit` base + trainable per-expert LoRA reduces MSE over 30 steps while
  the 4-bit base stays bit-identical.

### Non-goals (kept out to keep the diff strictly 4-bit)
- **N-bit storage generalization** (int8/fp8/bf16/fp16 under the same layout) — implemented and tested on
  the fork's [`experts-nbit-followup`](https://github.com/pjordanandrsn/bitsandbytes/tree/experts-nbit-followup)
  branch; happy to open it as a follow-up once this lands.
- **Grouped-GEMM** — the forward is a per-expert loop; a fused grouped kernel is future work.
- **`compress_statistics` / double-quant** — deliberately not used; `absmax` stays a plain fp32 buffer
  (no nested state). Straightforward to add later.
- **CPU-offload / streaming loader** — a separate capability that lives out-of-tree (in the
  `experts4bit-qlora` package); this PR is the quantization primitive only.

### Notes
- Pure Python over the stable `bitsandbytes.functional` APIs
  (`quantize_4bit` / `dequantize_4bit` / `get_4bit_type` / `QuantState`) — no kernel changes, so the
  tests pass against released bitsandbytes as well as source; nothing here depends on unreleased kernels.
- It's a **memory** optimization (experts 16-bit → 4-bit), not a throughput one — the per-expert dequant
  loop trades compute for memory, consistent with how bitsandbytes frames 4-bit generally.

Closes #1849. cc @matthewdouglas @SunMarc

### DIFF
--- a/bitsandbytes/nn/__init__.py
+++ b/bitsandbytes/nn/__init__.py
@@ -2,6 +2,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
+from .experts import Experts4bit
 from .modules import (
     Embedding,
     Embedding4bit,
--- a/bitsandbytes/nn/experts.py
+++ b/bitsandbytes/nn/experts.py
@@ -212,6 +212,27 @@ class Experts4bit(nn.Module):
         module.down_absmax = down_absmax
         return module
 
+    def _apply(self, fn, recurse=True):
+        """Shield the quantization state from floating-point dtype casts.
+
+        ``nn.Module`` dtype casts (``.to(dtype)``, ``.half()``, ``.bfloat16()``, …) convert
+        every floating-point tensor. The packed ``uint8`` weights are naturally untouched,
+        but the fp32 ``absmax``/``code`` state would be silently cast, degrading every
+        subsequent dequantization. Instead, a float dtype cast retargets ``compute_dtype``
+        while the quantization state stays fp32; device movement is unaffected.
+        """
+        requested_dtype = fn(torch.empty(0, dtype=torch.float32)).dtype
+        preserved = {name: getattr(self, name) for name in ("gate_up_absmax", "down_absmax", "code")}
+        module = super()._apply(fn, recurse)
+        if requested_dtype != torch.float32 and requested_dtype.is_floating_point:
+            module.compute_dtype = requested_dtype
+            for name, old in preserved.items():
+                new = getattr(module, name)
+                if new is not None and new.dtype != old.dtype:
+                    # Re-derive from the pre-cast tensor — a cast round-trip would be lossy.
+                    setattr(module, name, old.to(new.device))
+        return module
+
     def _quantize_stack(self, weights: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor]:
         """Quantize a ``[num_experts, out, in]`` stack to packed bytes + per-expert absmax."""
         packed = []
--- a/bitsandbytes/nn/experts.py
+++ b/bitsandbytes/nn/experts.py
@@ -1,0 +1,258 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from collections.abc import Callable
+from typing import Optional
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F_nn
+
+import bitsandbytes.functional as F
+from bitsandbytes.functional import QuantState
+
+
+class Experts4bit(nn.Module):
+    """4-bit quantized storage for fused Mixture-of-Experts expert weights.
+
+    A growing number of models in the Hugging Face ecosystem store their MoE expert
+    weights as a single 3D ``nn.Parameter`` of shape ``[num_experts, out_features,
+    in_features]`` (e.g. ``OlmoeExperts``, ``Qwen3MoeExperts``) rather than as a
+    collection of ``nn.Linear`` layers. The default 4-bit quantization walker only
+    replaces ``nn.Linear`` modules, so these fused experts are silently skipped and
+    stay in full precision — the dominant contribution to the model's memory footprint
+    (see https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1849).
+
+    ``Experts4bit`` holds the two expert projections (``gate_up_proj`` and ``down_proj``)
+    in 4-bit NF4/FP4 precision. Unlike :class:`Linear4bit`, the packed weights are kept
+    as plain ``nn.Parameter`` buffers and the per-expert quantization statistics
+    (``absmax``) live on the module as ordinary buffers. This avoids bending
+    :class:`Params4bit`'s tensor-subclass and device-movement machinery around a 3D
+    stack, and it means the module serializes through the standard ``state_dict``
+    mechanism with no custom save/load hooks.
+
+    The forward pass dequantizes a single expert at a time (a per-expert loop), mirroring
+    the reference fused-experts forward. Grouped-GEMM is intentionally left for future
+    work.
+
+    <Tip warning={true}>This feature is experimental and may change in future releases.</Tip>
+
+    Args:
+        num_experts (`int`): Number of experts in the layer.
+        hidden_dim (`int`): Model hidden size (the ``in_features`` of ``gate_up_proj`` and
+            the ``out_features`` of ``down_proj``).
+        intermediate_dim (`int`): Expert intermediate size (the ``in_features`` of
+            ``down_proj``).
+        has_gate (`bool`, *optional*, defaults to `True`): Whether ``gate_up_proj`` packs a
+            gate and an up projection (SwiGLU-style). When `False`, the projection is a
+            plain up projection of size ``intermediate_dim``.
+        activation (`Callable`, *optional*): The activation applied to the gate. Defaults
+            to ``torch.nn.functional.silu`` (SwiGLU), matching OLMoE / Qwen3-MoE.
+        compute_dtype (`torch.dtype`, *optional*): The dtype expert weights are
+            dequantized to for the matmul. When `None`, the input's dtype is used.
+        quant_type (`str`, *optional*, defaults to `"nf4"`): The 4-bit data type, ``nf4``
+            or ``fp4``.
+        blocksize (`int`, *optional*, defaults to `64`): The quantization block size.
+        device (*optional*): The device for the (empty) packed buffers.
+
+    Raises:
+        ValueError: If ``quant_type`` is invalid, or if ``hidden_dim`` / ``intermediate_dim``
+            is not divisible by ``blocksize`` (required so per-expert quantization blocks
+            never straddle an expert boundary).
+    """
+
+    def __init__(
+        self,
+        num_experts: int,
+        hidden_dim: int,
+        intermediate_dim: int,
+        has_gate: bool = True,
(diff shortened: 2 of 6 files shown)


matthewdouglas:
Hi, thanks for the PR. I am a little concerned with how quickly it was opened after discussion. With that said I'll follow up soon, but likely we won't merge something for this until after v0.50.0 release.

pjordanandrsn:
Thanks @matthewdouglas — fair concern. The asking-first part was real: nothing was written until the shape was pinned down, and the PR follows it — plain `nn.Parameter` + per-expert `absmax` buffers on the module, `compress_statistics` deferred, `in_features % blocksize == 0` enforced, per-expert dequant loop for the first cut. The footprint/VRAM numbers in the description are measured on my own A2000 12 GB; fitting fused-experts MoE models on a 12 GB card is the itch that started this.

No rush from me; post-v0.50.0 was always the plan. Converting to draft so it reads as what it is — something concrete to react to when you pick the feature up. Happy to rework it toward whatever you land on, or for you to cherry-pick the useful parts.

David-AU-github:
So looking forward to this; this will allow Unsloth training of Qwen 3.5/.6 and Gemma 4 sparse moes on reasonable hardware vs H100/200.

pjordanandrsn:
Updated this branch with backward/gradient + QLoRA-training test coverage and a small OLMoE-1B-7B demo validating quant-correctness and the ~4.7 GB footprint on a 12 GB card (per your June note). Still the correctness-first per-expert loop — no grouped-GEMM.

No rush given v0.50.0 isn't out yet; happy to leave this in draft until you pick it up.

Separately I have a memory-lowering `matmul_4bit` forward (gated so it only engages on bitsandbytes ≥ 0.50, else falls back to the bit-exact dequantize path) — glad to add it here, keep it as a follow-up, or leave it for your MoE-kernel work, whichever you prefer.

pjordanandrsn:
Rebased on current main; added a regression test for the exact #1849 silent-skip (fused 3-D module with no `nn.Linear` → asserts `from_float` actually 4-bit-quantizes it) plus shape-parametrized forward coverage — 38 tests, CPU+CUDA. Still draft, no rush.

pjordanandrsn:
@matthewdouglas, a couple updates:

Training path no longer needs v0.50.0: backward is now a small `_FrozenLinearRecomputeBackward` autograd Function—re-dequantizes each expert instead of saving it, bit-exact gradients on any current bitsandbytes, no `matmul_4bit` gate. Purely additive: `nn/experts.py` + export, docs, example, tests. A downstream QLoRA package built on it runs on stock `bitsandbytes>=0.43` (on PyPI, in-tree Kaggle-T4 notebook).

Also folded `Experts4bit` into an `ExpertsNbit` base (4-bit + 8/16-bit storage)—past this PR's 4-bit title, so I can split the N-bit into a follow-up or leave it, your call.

Still draft, no rush.

pjordanandrsn:
Marking ready for review — scope is strictly 4-bit (the N-bit generalization moved to the fork's `experts-nbit-followup` branch), and the training path needs no unreleased kernels: `_FrozenLinearRecomputeBackward` re-dequantizes per expert in backward, bit-exact on stock bitsandbytes. Since the draft: device-movement, dtype-cast, strict/non-strict `state_dict`, safetensors, and CUDA-numerics tests — 53 tests green on CPU and an RTX A2000. Built to the plain-`nn.Parameter` + module-buffer spec from #1849 — happy to adjust the API surface.

pjordanandrsn:
v0.50.0 is out, so this is unparked — rebased onto current `main`. CI has never run here; the workflows are still at `action_required`. Locally the test file is green on macOS/ARM64, on linux-x64 at torch 2.4.1 and 2.13.0, and on an A2000, plus lint and docs.

One flag I'd rather raise myself: `nn.parametrize.replace_parameter_4bit` already covers much of this (#1849; #1999 fixed its checkpointing case). `Experts4bit` adds per-expert dequant instead of the whole fused stack, and an explicit recompute-in-backward contract. If you'd rather have those in `parametrize`, I'll close this and do that instead.

No rush.

pjordanandrsn:
@matthewdouglas this implements the approach from the #1849 thread — packed weights as plain `nn.Parameter`s with per-expert `absmax` as module buffers (no `Params4bit` subclassing), `in_features % blocksize` enforced at construction, no `compress_statistics`. Purely additive, +1218/−0.

The four workflows have been at `action_required` since the Jul 26 rebase — mind approving them so there's real CI to review against?

pjordanandrsn:
@matthewdouglas — reading this back, I suspect the blocker isn't CI so much as +1218 lines being a lot to ask anyone to find time for.

Happy to cut it down to the minimum that closes #1849 — the quantization primitive and its regression test — and move the backward contract, the serialization and dtype coverage, and the QLoRA demo out into follow-ups. Say the word and I'll push that. Otherwise I'll leave it be and stop filling your inbox.

PR #1915, opened 2026-04-06
[ROCm] Windows workflow for creating wheels with ROCm 7.2.1 support

sstamenk:
### Summary

- Add CI workflow support for building bitsandbytes with ROCm on Windows, using pip-installable ROCm SDK wheels
- Update documentation to reflect Windows ROCm support across installation guide, README, and troubleshooting

### CI Build

- **`.github/scripts/build-rocm.sh`**: Add Windows branch that installs ROCm SDK wheels from `repo.radeon.com`, expands the devel tarball via `rocm-sdk init`, and builds with Ninja + Clang from the SDK. Targets RDNA 3/3.5/4 consumer GPU architectures.
- **`.github/workflows/python-package.yml`**: Add `windows-2025` + ROCm 7.2.1 to the `build-rocm` matrix via `include` (gated to 7.2.1 only). Add conditional MSVC setup step and Linux-only disk cleanup guard.

### CMakeLists.txt fixes

- **Path normalization**: Use `file(TO_CMAKE_PATH ...)` when reading `$ENV{ROCM_PATH}` to convert Windows backslashes to forward slashes, preventing CMake escape errors in generated HIP compiler files.
- **HIP_PLATFORM detection**: Set `HIP_PLATFORM=amd` explicitly because the pip SDK's `hip-config.cmake` cannot auto-detect it.
- **DLL output directory**: Extend `RUNTIME_OUTPUT_DIRECTORY` / `LIBRARY_OUTPUT_DIRECTORY` to all `WIN32` builds (not just MSVC), so Clang/Ninja HIP builds place the DLL in `bitsandbytes/`.

### Documentation

- **`docs/source/installation.mdx`**: Add Windows row to ROCm wheel table, add Windows compile-from-source tab with pip SDK workflow, update PyTorch ROCm links to cover Windows, add warning about GPU arch coverage in pre-built wheels.
- **`README.md`**: Add AMD GPU row to Windows platform support table.
- **`docs/source/errors.mdx`**: Add generic CUDA/ROCm version mismatch troubleshooting section, ROCm GPU arch mismatch section, and general diagnostics guidance.

### Minor fixes

- **`bitsandbytes/diagnostics/cuda.py`**: Add `amdhip64*.dll` Windows pattern to HIP runtime library search.
- **`bitsandbytes/cuda_specs.py`**: Mark `get_rocm_warpsize()` as dead code (no callers in the codebase).
[...]

PR #1833, opened 2025-12-22
Update cuda_specs.py

asmertpc-cloud:
fix for windows

PR #1969, opened 2026-06-10
Fix ROCm GPU arch detection: prefer torch device properties, robust hipInfo.exe lookup on Windows

danielhanchen:
## Problem

On Windows ROCm setups, `get_rocm_gpu_arch()` probes `hipinfo.exe` with `subprocess.run(["hipinfo.exe"], ...)`, which resolves through `PATH` only. In practice `hipInfo.exe` is rarely reachable that way:

- Hosts without the HIP SDK don't have it on `PATH` at all.
- AMD's PyTorch wheels for Windows ship `hipInfo.exe` into the environment's `Scripts` directory (next to `python.exe`), which is only on `PATH` while the venv is activated — not when the interpreter is invoked directly, via `uv run`, or embedded.

The probe then raises `FileNotFoundError`, and **every** `import bitsandbytes` logs:

```
ERROR:bitsandbytes.cuda_specs:Could not detect ROCm GPU architecture: [WinError 2] The system cannot find the file specified
WARNING:bitsandbytes.cuda_specs:
ROCm GPU architecture detection failed despite ROCm being available.
```

while `ROCM_GPU_ARCH` silently degrades to `"unknown"` — even though the GPU works fine.

## Fix

1. **Prefer `torch.cuda.get_device_properties(0).gcnArchName`** — torch already knows the architecture on both Linux and Windows, with no subprocess at all. Feature-flag suffixes (e.g. `gfx90a:sramecc+:xnack-`) are stripped to keep the existing `"gfx..."` format. This introduces no new device initialization: importing bitsandbytes already initializes the device context in `cextension.py` via `get_cuda_specs()` → `torch.cuda.get_device_capability()`.
2. **Keep the `rocminfo` / `hipInfo.exe` parsing as a fallback**, and on Windows additionally try `hipInfo.exe` next to `python.exe` (where AMD's wheels place it) before giving up.
3. The ERROR/WARNING logging is preserved for genuine failures; behavior on non-ROCm builds is unchanged.

## Validation

On a Strix Halo (gfx1151) Windows 11 machine with AMD's wheels (`torch 2.11.0+rocm7.13.0`), venv **not** activated, so `Scripts` is not on `PATH` and `shutil.which("hipinfo.exe")` is `None`:

- **Before:** `ROCM_GPU_ARCH == "unknown"` plus the ERROR/WARNING above on import.
- **After:** the torch-properties path returns `gfx1151` with no log output; forcing the subprocess fallback (mocking `torch.cuda.is_available` to `False`) also returns `gfx1151` via the `Scripts`-relative `hipInfo.exe`.

Added two mocked unit tests that run on any backend. On the ROCm box:

```
tests/test_cuda_setup_evaluator.py: 6 passed, 4 skipped (CUDA-only)
```

All `pre-commit` hooks pass on the changed files.

## Context

- #1846 introduced the Windows `hipinfo.exe` probe (superseding #1843 / #1833, which reported this same failure mode); #1915 added the Windows ROCm wheels.
- Downstream, Unsloth currently works around this by prepending the venv `Scripts` directory to `PATH` before importing bitsandbytes (unslothai/unsloth#5940, commit f48fc9a). This PR fixes it at the source so downstream consumers don't need that.

### DIFF
--- a/bitsandbytes/cuda_specs.py
+++ b/bitsandbytes/cuda_specs.py
@@ -1,9 +1,11 @@
 import dataclasses
 from functools import lru_cache
 import logging
+import os
 import platform
 import re
 import subprocess
+import sys
 from typing import Optional
 
 import torch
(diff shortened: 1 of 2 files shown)


danielhanchen:
Closing this for now.

matthewdouglas:
@danielhanchen This seems like a reasonable change - what's the reason for closing?

PR #1970, opened 2026-06-10
Improve Windows guidance for CUDA override environment variables

wunianze666-netizen:
## Summary
- make `cextension.py` generate platform-aware set/clear instructions for override environment variables
- replace Linux-only `export`/`unset` guidance in CUDA/ROCm override errors and warnings with Windows-specific instructions when running on Windows
- add regression tests for the new instruction helpers

## Why
The current override guidance is shown to Windows users as well, but it only suggests Unix shell commands. This change keeps the existing Unix guidance while making the same messages actionable on Windows.

## Validation
- ran a local stub-based regression script to verify the Windows instruction strings and a Windows-targeted CUDA override error path
- ran `python -m py_compile bitsandbytes/cextension.py tests/test_cuda_setup_evaluator.py`

### DIFF
--- a/bitsandbytes/cextension.py
+++ b/bitsandbytes/cextension.py
@@ -2,6 +2,7 @@ import ctypes as ct
 import functools
 import logging
 import os
+import platform
 from pathlib import Path
 import re
 from typing import Optional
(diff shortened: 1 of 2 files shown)


matthewdouglas:
Thanks, but this looks like a duplicate of #1957.

PR #1971, opened 2026-06-11
fix: ensure absmax_offset is of type float32 before passing to gemm_4bit

kathsucurry:
# TLDR

`gemm_4bit` kernel functions expect a float32 `absmax_offset`, but the offsets from some pre-quantized models are bfloat16. The discrepancy leads to gibberish offset values in the kernel, which further affects the outputs.

Relevant package versions:
- torch: 2.10.0+cu128
- accelerate: 1.13.0
- transformers: 5.3.0

# Background

I was playing around with [Unsloth_Puzzles.ipynb](https://colab.research.google.com/drive/1JqKqA1XWeLHvnYAc0wzrR4JBCnq43HyH?usp=sharing) part B on RTX5070Ti (sm120) with the same model, `unsloth/meta-Llama-3.1-8B-Instruct-bnb-4bit`, when I noticed that the training loss values changed drastically after commit 5453368 "[CUDA] New 4bit GEMM kernels for inference (#1949)". Specifically, the loss starts off much larger and quickly becomes 0 (NaN values):

| Step | Training loss before 5453368 | Training loss after 5453368 |
|----------|----------|----------| 
| 1   |2.109375 | 7.015625 |
| 2   |2.343750 | 7.828125 |
| 3   |2.750000 | 11.593750 |
| 4   |2.015625 | 0.000000 |
| 5   |1.960938 | 0.000000 |
| 6   |1.945312 | 0.000000 |
| 7   |1.656250 | 0.000000 |
| 8   |1.750000 | 0.000000 |
| 9   |1.617188 | 0.000000 |
| 10 |1.882812 | 0.000000 |

*Note that the training loss before the mentioned commit is similar to that in the notebook.*

By trying out different models, I found that the loss values show the same issue when 1) a pre-quantized base model is used and 2) `gemm_4bit` is called (when [`use_custom` is `True`](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/bitsandbytes/backends/cuda/ops.py#L871)). Some of the models I used are as follows:

| Model | Is affected | Is pre-quantized | `use_custom` is `True` |
|----------|----------|----------| ----------| 
|unsloth/meta-Llama-3.1-8B-Instruct-bnb-4bit| Yes|Yes|Yes |
|unsloth/gemma-3-270m-bnb-4bit|Yes|Yes|Yes|
|unsloth/gemma-3-270m|No|No|Yes|
|unsloth/Llama-3.2-3B-Instruct-bnb-4bit|No|Yes|No|

# Findings

The bug appears to be caused by the type of `absmax_offset`. The `absmax_offset` values coming from the pre-quantized models I've used so far are bfloat16, while the kernel function expects a float32 tensor. Consequently, the offset value becomes gibberish and extremely large, which causes overflow.

I've made a small change along with the corresponding test. The training loss for all the models mentioned above is now more stable. The model `unsloth/meta-Llama-3.1-8B-Instruct-bnb-4bit`, for instance, now gives the following outputs:

| Step | Training loss before 5453368 | Training loss after 5453368 | Training loss after the bug fix |
|----------|----------|----------| ----------| 
| 1   |2.109375 | 7.015625 | 2.109375 |
| 2   |2.343750 | 7.828125 | 2.335938 |
| 3   |2.750000 | 11.593750 | 2.755859 |
| 4   |2.015625 | 0.000000 | 2.013672 |
| 5   |1.960938 | 0.000000 | 1.948242 |
| 6   |1.945312 | 0.000000 | 1.929688 |
| 7   |1.656250 | 0.000000 | 1.637695 |
| 8   |1.750000 | 0.000000 | 1.728516 |
| 9   |1.617188 | 0.000000 | 1.561523 |
| 10 |1.882812 | 0.000000 | 1.808594 |

I've only tried Unsloth's pre-quantized models, so do let me know if there are other pre-quantized models you'd like me to try (or if there are other tests you'd like me to do).

### DIFF
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -337,10 +337,7 @@ class Test4bitBlockwiseQuantOps:
     @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
     @pytest.mark.parametrize("offset_dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
     def test_gemm_4bit_non_float32_offset(self, device, dtype, offset_dtype):
-        """Regression test: offset tensors not in float32 must still produce correct results.
-
-        Pre-quantized models (e.g. Unsloth bnb-4bit) may store qs.offset in non-float32 dtype.
-        """
+        """Regression test: offset tensors not in float32 must still produce correct results."""
         N, K, blocksize = 64, 64, 64
         A = torch.randn(4, K, dtype=dtype, device=device)
         B = torch.randn(N, K, dtype=dtype, device=device)
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -333,6 +333,50 @@ class Test4bitBlockwiseQuantOps:
                 kwargs={"bias": bias},
             )
 
+    @pytest.mark.parametrize("device", get_available_devices())
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
+    @pytest.mark.parametrize("offset_dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
+    def test_gemm_4bit_non_float32_offset(self, device, dtype, offset_dtype):
+        """Regression test: offset tensors not in float32 must still produce correct results.
+
+        Pre-quantized models (e.g. Unsloth bnb-4bit) may store qs.offset in non-float32 dtype.
+        """
+        N, K, blocksize = 64, 64, 64
+        A = torch.randn(4, K, dtype=dtype, device=device)
+        B = torch.randn(N, K, dtype=dtype, device=device)
+        B_q, qs = bitsandbytes.functional.quantize_4bit(
+            B, blocksize=blocksize, quant_type="nf4", compress_statistics=True
+        )
+
+        # Simulate a pre-quantized model where offset may not be float32.
+        offset_non_f32 = qs.offset.to(dtype=offset_dtype)
+
+        # Reference: explicitly use the rounded float32 value.
+        offset_as_f32 = offset_non_f32.to(dtype=torch.float32)
+        ref = torch.ops.bitsandbytes.gemm_4bit.default(
+            A,
+            B_q,
+            list(B.shape),
+            qs.state2.absmax,
+            blocksize,
+            "nf4",
+            absmax_8bit=qs.absmax,
+            absmax_code=qs.state2.code,
+            absmax_offset=offset_as_f32,
+        )
+        out = torch.ops.bitsandbytes.gemm_4bit.default(
+            A,
+            B_q,
+            list(B.shape),
+            qs.state2.absmax,
+            blocksize,
+            "nf4",
+            absmax_8bit=qs.absmax,
+            absmax_code=qs.state2.code,
+            absmax_offset=offset_non_f32,
+        )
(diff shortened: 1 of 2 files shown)


matthewdouglas:
Thanks for catching this! Out of curiosity, can you please provide some more detail on your environment? Specifically versions of torch, accelerate, and transformers.

This seems reasonable as a defensive measure, but I don't think the root issue is in the serialized models themselves. There's a quirk in the 4bit serialization where this offset value is actually stored in a utf-8 encoded uint8 tensor as a JSON string. When deserialized, it uses the Python `float` type:

https://github.com/bitsandbytes-foundation/bitsandbytes/blob/9dad66596114659177ef4743e38d8c2a990d11eb/bitsandbytes/functional.py#L521-L523

I can see how in some circumstance (e.g. `torch.set_default_dtype()` is applied in context where we're deserializing) that this may materialize as a fp16 or bf16 tensor. I think this may be the more likely cause, so I may open a separate PR here to force that to be fp32.

The other possibility is  `QuantState.to()` which is supposed to mostly only handle device movement, not casting. It's called when a weight (`Params4bit.to()`) is moved. It is possible this may be casting the offset if it is called as `weight.to(torch.bfloat16)`, so I may also address this in a PR as well.

One last, but less likely, possibility here is that you have an AVX512BF16 enabled CPU, and moved between CPU and GPU at some point after running a forward pass on CPU. This may have inadvertently cast the offset.

Since you only observe the issue with pre-quantized models, it leans heavily toward being related to the deserialization.

tests/test_ops.py
```diff
@@ -333,6 +333,50 @@ def test_gemm_4bit(self, device, dtype, quant_type, compress_statistics, has_bia
...
             )
 
+    @pytest.mark.parametrize("device", get_available_devices())
+    @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
+    @pytest.mark.parametrize("offset_dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
+    def test_gemm_4bit_non_float32_offset(self, device, dtype, offset_dtype):
+        """Regression test: offset tensors not in float32 must still produce correct results.
+
+        Pre-quantized models (e.g. Unsloth bnb-4bit) may store qs.offset in non-float32 dtype.
+        """
```
matthewdouglas: This is inaccurate as the serialized checkpoints don't have a specific dtype for the offset. Strictly speaking, it's stored in JSON as a string of digits, and when loading it is reading as a Python `float` which is `fp64`. So it's more about tensor creation/manipulation than it is about storage.


```suggestion
        """Regression test: offset tensors not in float32 must still produce correct results."""
```

kathsucurry:
Ah I should have included the versions in the PR description. Below are the versions (and I'll add them in the description as well):

- torch:  2.10.0+cu128
- accelerate: 1.13.0
- transformers: 5.3.0

Thank you so much for the comments! The detailed breakdown gives really helpful context on how the offset gets stored and deserialized. I still lack context on the serialization/deserialization path or `QuantState.to()`, so I'll be happy to look into the possibilities you outlined if you'd like me to!

tests/test_ops.py
kathsucurry: Noted. Thank you for the explanation!

matthewdouglas:
Thanks! I think we're good here with this. The remaining things are something I can address in tandem with another PR that I'll do for improving host-side overhead.

PR #1974, opened 2026-06-15
Fix diagnostics package version lookup

wunianze666-netizen:
## Summary
- switch diagnostics package version lookup to importlib.metadata`n- handle PackageNotFoundError through the imported metadata module
- add a focused regression test for installed and missing packages

## Why
get_package_version() currently calls importlib.metadata without importing the submodule first, so the helper can raise AttributeError before reaching its fallback path.

## Testing
- compiled the updated source files locally
- ran an isolated Python verification script that stubs heavy runtime dependencies and checks both installed and missing package paths

Issue linkage: not tied to an existing tracked issue.

### DIFF
--- a/bitsandbytes/diagnostics/main.py
+++ b/bitsandbytes/diagnostics/main.py
@@ -1,7 +1,7 @@
-import importlib
 import platform
 import sys
 import traceback
+from importlib import metadata as importlib_metadata
 
 import torch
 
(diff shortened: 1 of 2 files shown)


wunianze666-netizen:
Closing this after re-validating the import sequence on the PR's base commit. itsandbytes/__init__.py calls rom importlib.metadata import entry_points before itsandbytes.diagnostics.main can be imported, which loads the submodule and makes importlib.metadata available on the parent package. The added tests also import importlib.metadata first, so they do not fail against the original implementation. I therefore cannot demonstrate the reported AttributeError through a supported import path, and keeping this PR open would not be justified. Sorry for the noise.

PR #1983, opened 2026-07-01
MPS: improved backend

matthewdouglas:
Makes the MPS backend usable without the `kernels-community/bitsandbytes-mps` hub kernel - either because the user is on macOS < 26, doesn't have kernels installed, the hub kernel lacks support for the shape/blocksize, or the hub kernel fails to load for any reason. Previously it was not usable in these situations.

Changes:
* Most ops are now unconditionally registered, and will attempt to use the HF Hub kernels before falling back to a slower implementation.
* Hub kernel is only attempted on macOS 26+
* Fix validation of shapes/blocksizes to match what Hub kernels support
* Improved performance of the fallback MPS op implementations

At this point, MPS works with all of the 4bit or LLM.int8() configurations. All tests pass on M4 + macOS 15, with exception of the optimizer support, which will be a future PR.

The MPS path in general requires torch >= 2.9, and for the hub kernels will currently require torch 2.10-2.12. Documentation updates regarding this will be addressed in a future PR.

### DIFF
--- a/bitsandbytes/backends/default/ops.py
+++ b/bitsandbytes/backends/default/ops.py
@@ -1,6 +1,6 @@
 from collections.abc import Sequence
 from functools import cache, wraps
-from math import sqrt
+from math import prod, sqrt
 from typing import Optional
 
 import torch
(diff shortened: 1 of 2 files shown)


hellopahe:
@matthewdouglas are you planning the optimizer support yourself, or is it up for grabs? Happy to take it on.

PR #1985, opened 2026-07-05
Add BF16 support for int8_vectorwise_quant / LLM.int8 activation quant

### LINKED ISSUE #1868: Question: intentional FP16-only path for int8_vectorwise_quant / LLM.int8 activation quant? (BF16 support + removing casts)
Hi guys, first of all, incredible work👍

Just a quick design question about the `LLM.int8` activation quantization path.

### Context

In `MatMul8bitLt.forward`, activations A are always cast to FP16 before quantization:

```py
#bitsandbytes/autograd/_functions.py::MatMul8bitLt.forward
CA, SCA, outlier_cols = F.int8_vectorwise_quant(A.to(torch.float16), threshold=state.threshold)
```

and the CUDA kernel implementation currently hard-requires FP16:
```py
# bitsandbytes/backends/cuda/ops.py
@register_kernel("bitsandbytes::int8_vectorwise_quant", "cuda")
def _(A, threshold=0.0):
    torch._check(A.dtype == torch.float16, ...)
    lib.cint8_vector_quant(get_ptr(A), ...)
```
On the native side, the exported ABI and launcher are also half-only:

- `csrc/pythonInterface.cpp`: cint8_vector_quant(half* A, ...)
- `csrc/ops.cu`: int8VectorQuant(half* A, ...)
- `csrc/kernels.cu`: instantiations only for half
So BF16 inputs get an extra bf16 -> fp16 cast + warning even though the rest of the pipeline tries to preserve output dtype (int8_scaled_mm(..., dtype=A.dtype)).

### Question

Was the FP16-only design for `int8_vectorwise_quant / LLM.int8` activation quantization intentional (e.g. for kernel simplicity, CUB reduction constraints, arch compatibility), or is it mainly an unimplemented gap?

I’m asking because I see a lot of LLM inference/training frameworks use BF16 activations by default these days, and this path currently forces an FP16 cast for quantization.

kru2710shna:
Templates int8VectorQuant on T and adds bf16 kernel instantiations plus a cint8_vector_quant_bf16 C ABI entry point, mirroring the existing gemm_4bit_inference_naive fp16/bf16/fp32 pattern. The blockwise absmax reduction now accumulates in float rather than T: required for bf16 to compile cleanly and slightly improves fp16 accuracy (rowStats was already float, so downstream is unaffected). Removes the forced A.to(torch.float16) casts in MatMul8bitLt so bf16 activations quantize natively.

Closes #1868.

### DIFF
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -134,10 +134,10 @@ class MatMul8bitLt(torch.autograd.Function):
         # 1. Quantize A. Note that as a side-effect, outliers are suppressed in CA/CAt.
         if ctx.needs_input_grad[1]:
             # Slower path
-            CA, CAt, SCA, SCAt, outlier_cols = F.int8_double_quant(A.to(torch.float16), threshold=state.threshold)
+            CA, CAt, SCA, SCAt, outlier_cols = F.int8_double_quant(A, threshold=state.threshold)
         else:
             # Fast path
-            CA, SCA, outlier_cols = F.int8_vectorwise_quant(A.to(torch.float16), threshold=state.threshold)
+            CA, SCA, outlier_cols = F.int8_vectorwise_quant(A, threshold=state.threshold)
             CAt = SCAt = None
 
         has_grad = False
@@ -152,7 +152,7 @@ class MatMul8bitLt(torch.autograd.Function):
                 state.reset_grads()
 
                 # 2. Quantize B
-                state.CB, state.SCB, _ = F.int8_vectorwise_quant(B.to(torch.float16))
+                state.CB, state.SCB, _ = F.int8_vectorwise_quant(B)
 
         # Handle sparse decomposition
         if state.threshold > 0.0:
@@ -258,7 +258,7 @@ class MatMul8bitFp(torch.autograd.Function):
 
             if (state.is_training and not has_grad) or state.CB is None or state.SCB is None:
                 state.reset_grads()
-                state.CB, state.SCB, _ = F.int8_vectorwise_quant(B.to(torch.float16))
+                state.CB, state.SCB, _ = F.int8_vectorwise_quant(B)
                 B = state.CB
 
         CB = state.CB.data.to(A.dtype).mul_(state.SCB.unsqueeze(1).mul(1.0 / 127.0))
--- a/bitsandbytes/autograd/_functions.py
+++ b/bitsandbytes/autograd/_functions.py
@@ -124,10 +124,6 @@ class MatMul8bitLt(torch.autograd.Function):
 
         input_shape = A.shape
 
-        # Cast A to fp16
-        if A.dtype != torch.float16 and not _is_compiling():
-            logger.warning("MatMul8bitLt: inputs will be cast from %s to float16 during quantization", A.dtype)
-
         if len(A.shape) == 3:
             A = A.reshape(-1, A.shape[-1])
 
--- a/csrc/kernels.cu
+++ b/csrc/kernels.cu
@@ -1384,10 +1384,10 @@ template __global__ void kInt8VectorQuant<bnb_bfloat16, 1024, 1>(
 );
 #define MM_DEQUANT_CONST 6.200012e-05f // 1.0f/(127.0f*127.0f)
 
-template <int ITEMS_PER_THREAD, int THREADS>
+template <typename T, int ITEMS_PER_THREAD, int THREADS>
 __global__ void kdequant_mm_int32_fp16(
-    int* __restrict__ const A, float* __restrict__ const rowStats, float* __restrict__ const colStats, half* out,
-    half* __restrict__ const bias, const int numRows, const int numCols, const int n
+    int* __restrict__ const A, float* __restrict__ const rowStats, float* __restrict__ const colStats, T* out,
+    T* __restrict__ const bias, const int numRows, const int numCols, const int n
 ) {
     const int n_out = numRows * numCols;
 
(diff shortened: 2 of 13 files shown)


csrc/kernels.cu
```diff
@@ -1324,52 +1324,40 @@ template <typename T, int THREADS, int SPARSE_DECOMP>
...
     void kInt8VectorQuant(T* __restrict__ A, int8_t* out, float* rowStats, float threshold, int rows, int cols) {
 
-    using BlockReduceT = bnb_cub::BlockReduce<T, THREADS>;
-
-    // One block per row.
-    // Threads load column values in a striped arrangement.
-    // e.g. t0 reads row[0], row[0+nthreads], ..
-    // and  t1 reads row[1], row[1+nthreads], ..
-    // Each thread will determine its local absmax.
-    // We then do a blockwise reduction to determine the row's absmax.
```
matthewdouglas, line 1334: Why was this comment removed?

csrc/kernels.cu
```diff
@@ -1324,52 +1324,40 @@ template <typename T, int THREADS, int SPARSE_DECOMP>
...
 
     // Quantize row-wise.
     const float scale = __fdividef(127.0f, smem_row_absmax);
     for (int i = threadIdx.x; i < cols; i += THREADS) {
-        float val = row_data[i];
+        float val = (float)row_data[i];
 
         if constexpr (SPARSE_DECOMP) {
-            // For sparse decomposition, we do not want to quantize the outliers.
-            // Instead they're zeroed out.
```
matthewdouglas, line 1371: Why remove this comment?

bitsandbytes/autograd/_functions.py
matthewdouglas: This warning would likely no longer apply.

csrc/kernels.cu
kru2710shna: Not intentional.  These got dropped when I rewrote the kernel body for float accumulation. They're accurate and unrelated to the change, so I will restored them verbatim in the next commit

csrc/kernels.cu
kru2710shna: Same cause as above, accidental during the float-reduction rewrite. Restored. Thanks for catching both.

matthewdouglas:
I think it would be worthwhile to update some of the other relevant tests to exercise the new bf16 path. E.g. in `test_functional.py`, `test_linear8bit.py`, `test_modules.py`, `test_autograd.py` where appropriate.

kru2710shna:
Will do
 I'll add bf16 coverage to the higher-level tests (test_functional.py, test_linear8bit.py, test_modules.py, test_autograd.py) in a follow-up commit on this branch.

kru2710shna:
Done in #1985 - e888a86.  Added bf16 coverage across the relevant tests:

test_matmullt: dropped the forced fp16 cast so the existing bf16 param now exercises the real path end-to-end.
test_int8_linear_matmul_half and test_linear8bitlt_inference: parametrized over fp16/bf16.
New test_linear8bitlt_forward_dtypes for the module forward in both dtypes.

Note: test_linear8bit.py didn't exist, so I covered test_linear8bitlt.py instead. Also rebased onto latest main. 

Happy to adjust anything.

matthewdouglas:
There's still a test failure and a couple build issues to resolve.

`bitsandbytes.functional.int8_mm_dequant` right now hard-codes the output dtype to be `torch.float16`, so I think at the very least in this test we'd consider casting this. A more thorough support of `bfloat16` throughout would involve implementing `bfloat16` output and bias addition directly in the `int8_mm_dequant` kernel and adding support for the `dtype` kwarg to `bitsandbytes.functional.int8_mm_dequant` as well.

kru2710shna:
Went ahead and implemented this properly rather than just patching the test. int8_mm_dequant now natively supports bfloat16 output end-to-end:

Templated kdequant_mm_int32_fp16 and its launcher on output dtype, added a bnb_bfloat16 instantiation and a cdequant_mm_int32_bf16 C ABI entry.
The Python backend dispatches on the requested dtype and allocates out directly in that dtype.
Fused bias now works for bf16 too (previously fp16-only), as it fuses in-kernel whenever bias.dtype matches the output dtype, otherwise added afterward as before.
bitsandbytes.functional.int8_mm_dequant now takes a dtype kwarg instead of hardcoding fp16.
Reverted the test workaround, this test_int8_linear_matmul_half now asserts a real bf16-to-bf16 comparison.

This should close the last known gap for bf16 support in this PR.

csrc/ops.cu
```diff
@@ -481,6 +483,21 @@ template void gemm_4bit_inference_naive<float, 32>(
...
+    bnb_bfloat16* __restrict__ A, int8_t* out, float* rowStats, float threshold, int rows, int cols, bnb_stream_t stream
+);
+
+template void dequant_mm_int32_fp16<half>(
+    int* A, float* rowStats, float* colStats, half* out, half* bias, int numRows, int numCols, bnb_stream_t stream
+);
+template void dequant_mm_int32_fp16<bnb_bfloat16>(
+    int* A, float* rowStats, float* colStats, bnb_bfloat16* out, bnb_bfloat16* bias, int numRows, int numCols,
+    bnb_stream_t stream
+);
```
matthewdouglas, line 499: Now that it supports bf16 and fp16, the naming on this with the `_fp16` suffix is a little confusing. It may make more sense to just rename it to `dequant_mm_int32<T>`.

tests/test_linear8bitlt.py
```diff
@@ -364,3 +365,32 @@ def test_linear8bitlt_device_movement(device):
 
     # Accelerator outputs should match both times.
     torch.testing.assert_close(out_accelerator_2, out_accelerator, rtol=1e-8, atol=1e-8)
+
+
+@pytest.mark.parametrize("device", get_available_devices(no_cpu=True))
```
matthewdouglas, line 370: Why skip on CPU?

matthewdouglas:
Hi @kru2710shna,

There still remains build failures on CUDA < 12.4 and on ROCm. Additionally there are test failures where the build succeeded.

PR #1993, opened 2026-07-08
Fix Lion to use decoupled weight decay (default + CUDA 32-bit)

### LINKED ISSUE #1992: Lion applies coupled (L2) weight decay instead of decoupled in 32-bit CUDA, Triton, and the default backend
### Summary

The **Lion** optimizer applies **coupled (L2) weight decay** — folding `weight_decay * p` into the gradient _before_ the sign update — in three code paths, when Lion requires **decoupled** (AdamW-style) weight decay: shrink the parameter directly (`p *= 1 - lr*wd`), _outside_ the `sign()`. This corrupts the update whenever `weight_decay > 0`.

This is a correctness bug, not just a numerical detail: coupled decay changes what `sign()` sees (`sign(momentum_of(g + wd*p))` vs. the intended `sign(momentum_of(g))`), so it alters the direction of the update, not merely its magnitude. It affects **32-bit Lion on CUDA** (the main user path), 32-bit Lion on **Triton**, and Lion on the **`default`** backend (MPS and any device without a dedicated kernel).

The reference (Chen et al. 2023, _Symbolic Discovery of Optimization Algorithms_) applies decoupled decay: `θ_t = θ_{t-1} - η(sign(c_t) + λθ_{t-1}) = θ_{t-1}(1 - ηλ) - η·sign(c_t)`.

### The smoking gun: bitsandbytes is inconsistent with itself

The CUDA **8-bit blockwise** Lion kernel already does the _correct_ decoupled decay, while the CUDA **32-bit** kernel does the _incorrect_ coupled decay. Nobody chooses that split on purpose — it's a bug in the 32-bit path, not a design decision.

### Affected code (line numbers approximate; cite by function)

| Path                      | Location                                                                                                                                                                                                    | Lion weight decay | Correct? |
| ------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------- | -------- |
| CUDA 32-bit               | `csrc/kernels.cu` → `kOptimizer32bit1State` (~L861: `g_vals[j] = g + p*wd` applied to all 1-state optimizers, incl. `case LION`)                                                                            | **coupled**       | ❌       |
| Triton 32-bit             | `bitsandbytes/backends/triton/kernels_optim.py` → `_optimizer_update_1state_triton_kernel` (~L273: `g_vals = g_vals + p_vals * weight_decay` before `OPTIMIZER_ID == 4`)                                    | **coupled**       | ❌       |
| `default` backend         | `bitsandbytes/backends/default/ops.py` → `_optimizer_update_32bit` (~L453: `if optimizer_id in [0, 1, 2, 4] ...: g_vals = g_vals + p_vals * weight_decay`; the `LION` branch never applies decoupled decay) | **coupled**       | ❌       |
| CPU                       | `bitsandbytes/backends/cpu/ops.py` (lion branch: `p_float.mul_(1.0 - lr * weight_decay)`)                                                                                                                   | decoupled         | ✅       |
| CUDA 8-bit blockwise      | `csrc/kernels.cu` (`case LION: p_vals[j] = p * (1 - lr*wd)`)                                                                                                                                                | decoupled         | ✅       |
| Triton 8-bit (high-level) | `kernels_optim.py` (lion: `p_fp32.mul_(1.0 - lr*weight_decay)`)                                                                                                                                             | decoupled         | ✅       |

### Why this went unnoticed

`tests/test_optimizer32bit` constructs both the reference and the bnb optimizer with the **default `weight_decay=0`**, where coupled and decoupled decay are identical. The bug lives entirely in the `weight_decay > 0` regime, which the 32-bit test never exercises.

### Reproduction (CUDA; also reproduces on MPS)

```python
import torch, bitsandbytes as bnb
from lion_pytorch import Lion  # reference: decoupled weight decay, per the Lion paper

torch.manual_seed(0)
dev = "cuda"  # buggy 32-bit CUDA kernel; use "mps" for the default backend. "cpu" is already correct.
p_ref = (torch.randn(1024, 1024, device=dev) * 0.1)
p_bnb = p_ref.clone()
p_ref.requires_grad_(); p_bnb.requires_grad_()

opt_ref = Lion([p_ref], lr=1e-3, weight_decay=0.1)
opt_bnb = bnb.optim.Lion([p_bnb], lr=1e-3, weight_decay=0.1)

for _ in range(20):
    g = torch.randn(1024, 1024, device=dev) * 0.01
    p_ref.grad, p_bnb.grad = g.clone(), g.clone()
    opt_ref.step(); opt_bnb.step()

print((p_ref - p_bnb).abs().max().item())  # large divergence; ~0 only when weight_decay=0
```

### Suggested fix

Treat Lion like the other decoupled optimizers: **exclude LION from the coupled-decay group**, and apply `p *= 1 - lr*weight_decay` in the Lion branch (exactly as the CPU backend and the CUDA 8-bit blockwise kernel already do). Concretely, in each buggy path:

- Remove `4` (LION) from the `optimizer_id in [0, 1, 2, 4]` / equivalent coupled-decay condition.
- In the `LION` case, add the decoupled shrink to the parameter before the sign update.

I've implemented and verified this fix for the **`default` backend** (with a new `weight_decay > 0` regression test that fails before / passes after, validated on Apple Silicon where `default` is the active Lion path). Happy to open a PR covering the `default` backend, and — with a maintainer's confirmation on the intended semantics change and access to CUDA CI — the CUDA and Triton kernels as well.

### Note on the semantics change

Fixing this changes the training trajectory for anyone currently using `Lion(weight_decay > 0)` on CUDA/Triton/MPS. It makes them consistent with the CPU backend, the CUDA 8-bit path, and the paper, but it is a behavior change worth calling out in release notes.

eaglstun:
Fixes #1992.

## Problem

Lion with `weight_decay > 0` applies **coupled (L2)** weight decay, folding decay into the gradient before the `sign()` in the 32-bit optimizer paths. Lion requires **decoupled** (AdamW-style) decay applied to the parameter directly (`p *= 1 - lr*wd`), _outside_ the sign update (Chen et al. 2023).

Coupling decay into the gradient changes the input to `sign()`, so it corrupts the update _direction_, not just its magnitude.

The smoking gun that this is a bug and not a design choice: the CUDA backend is already inconsistent with itself. The **8-bit blockwise** kernel and the **cpu** backend do decoupled decay correctly; only the **32-bit** paths got it wrong.

## Fix

| Backend     | Path                                     | Before    | After             |
| ----------- | ---------------------------------------- | --------- | ----------------- |
| default     | `backends/default/ops.py`                | coupled   | ✅ decoupled      |
| CUDA 32-bit | `csrc/kernels.cu::kOptimizer32bit1State` | coupled   | ✅ decoupled      |
| CUDA 8-bit  | `kOptimizerStatic8bit1State`             | decoupled | (already correct) |
| cpu         | `backends/cpu/ops.py`                    | decoupled | (already correct) |

The CUDA change mirrors the existing 8-bit kernel exactly: exclude `LION` from the coupled gradient fold, and apply `p *= (1 - lr*wd)` in the `LION` branch before the sign update (un-scaled by the `max_unorm` clip, matching the reference).

I verified `kOptimizer32bit1State` is the only 32-bit CUDA site that applies Lion decay: the 2-state kernel is ADAM/ADEMAMIX (already decoupled), and the 1-state precondition kernel only accumulates the update norm for Lion.

## Test

Adds `test_lion32bit_weight_decay` to `tests/test_optim.py`, comparing `bnb.optim.Lion` against the `lion-pytorch` reference with `weight_decay=0.1`. It's parametrized over `get_available_devices()`, so **it exercises the CUDA 32-bit kernel on GPU CI**. The existing `test_optimizer32bit` never caught this because it uses the default `weight_decay=0`, where coupled and decoupled coincide.

Under the bug the params diverge far past the error budget; with the fix they track the reference.

## Notes / scope

- I can run the default-backend + CPU tests locally, but **not** the CUDA kernel, it's unbuildable on Apple Silicon. Relying on GPU CI to validate the 32-bit CUDA path (happy to help debug any failures).
- The **Triton** 1-state kernel (`backends/triton/kernels_optim.py`) carries the same coupled-decay bug, but it's XPU-only with no CI hardware here. Left as a follow-up, happy to open a companion PR if someone can validate on XPU.

### DIFF
--- a/bitsandbytes/backends/default/ops.py
+++ b/bitsandbytes/backends/default/ops.py
@@ -522,9 +522,6 @@ def _optimizer_update_32bit(
         # Lion uses decoupled weight decay: shrink the param directly (p *= 1 - lr*wd)
         # rather than folding wd into the gradient. Matches the cpu backend, the CUDA
         # 8-bit blockwise kernel, and the Lion paper (Chen et al. 2023).
-        # NOTE: the Triton 1-state kernel (backends/triton/kernels_optim.py) still
-        # applies *coupled* decay to Lion — the same bug this fixes — and needs a
-        # separate fix. The CUDA 32-bit kernel is fixed alongside this change.
         if weight_decay > 0.0:
             p_vals = p_vals * (1.0 - lr * weight_decay)
 
--- a/bitsandbytes/backends/default/ops.py
+++ b/bitsandbytes/backends/default/ops.py
@@ -450,7 +450,11 @@ def _optimizer_update_32bit(
 
     p_vals = p.float()
     g_vals = (gnorm_scale * g).float()
-    if optimizer_id in [0, 1, 2, 4] and weight_decay > 0.0:
+    # Coupled (L2) weight decay: fold wd into the gradient. This is correct for
+    # MOMENTUM/RMSPROP/ADAGRAD, but NOT for LION (id 4), which uses *decoupled*
+    # (AdamW-style) weight decay applied to the param directly (see the LION branch
+    # below and Chen et al. 2023). LION is intentionally excluded here.
+    if optimizer_id in [0, 1, 2] and weight_decay > 0.0:
         g_vals = g_vals + p_vals * weight_decay
 
     update_scale = 1.0
@@ -515,6 +519,15 @@ def _optimizer_update_32bit(
         state1.copy_(s1_vals)
 
     elif optimizer_id == 4:  # LION
+        # Lion uses decoupled weight decay: shrink the param directly (p *= 1 - lr*wd)
+        # rather than folding wd into the gradient. Matches the cpu backend, the CUDA
+        # 8-bit blockwise kernel, and the Lion paper (Chen et al. 2023).
+        # NOTE: the CUDA 32-bit kernel (csrc/kernels.cu::kOptimizer32bit1State) and the
+        # Triton 1-state kernel (backends/triton/kernels_optim.py) still apply *coupled*
+        # decay to Lion — the same bug this fixes — and need a separate upstream fix.
+        if weight_decay > 0.0:
+            p_vals = p_vals * (1.0 - lr * weight_decay)
+
         momentum_update = state1 * beta1 + (1.0 - beta1) * g_vals
         update_val = update_scale * lr * torch.sign(momentum_update)
         p_vals = p_vals - update_val
--- a/bitsandbytes/backends/default/ops.py
+++ b/bitsandbytes/backends/default/ops.py
@@ -522,9 +522,9 @@ def _optimizer_update_32bit(
         # Lion uses decoupled weight decay: shrink the param directly (p *= 1 - lr*wd)
         # rather than folding wd into the gradient. Matches the cpu backend, the CUDA
         # 8-bit blockwise kernel, and the Lion paper (Chen et al. 2023).
-        # NOTE: the CUDA 32-bit kernel (csrc/kernels.cu::kOptimizer32bit1State) and the
-        # Triton 1-state kernel (backends/triton/kernels_optim.py) still apply *coupled*
-        # decay to Lion — the same bug this fixes — and need a separate upstream fix.
+        # NOTE: the Triton 1-state kernel (backends/triton/kernels_optim.py) still
+        # applies *coupled* decay to Lion — the same bug this fixes — and needs a
+        # separate fix. The CUDA 32-bit kernel is fixed alongside this change.
         if weight_decay > 0.0:
             p_vals = p_vals * (1.0 - lr * weight_decay)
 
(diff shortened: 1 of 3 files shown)


bitsandbytes/backends/default/ops.py
```diff
@@ -515,6 +519,15 @@ def _optimizer_update_32bit(
         state1.copy_(s1_vals)
 
     elif optimizer_id == 4:  # LION
+        # Lion uses decoupled weight decay: shrink the param directly (p *= 1 - lr*wd)
+        # rather than folding wd into the gradient. Matches the cpu backend, the CUDA
+        # 8-bit blockwise kernel, and the Lion paper (Chen et al. 2023).
+        # NOTE: the Triton 1-state kernel (backends/triton/kernels_optim.py) still
+        # applies *coupled* decay to Lion — the same bug this fixes — and needs a
+        # separate fix. The CUDA 32-bit kernel is fixed alongside this change.
```
matthewdouglas: Minor nit: I don't think this part of the comment is entirely necessary; we can track in an issue and separate PR for that.
```suggestion
```

matthewdouglas (approved):
Looks good, thanks for the PR! Appreciate that it is well scoped and has a reasonable test addition. All of the CUDA tests are passing.

If you'd like, feel free to open an equivalent PR for the Triton backend. I wouldn't expect it to be particularly invasive and probably safe to merge whether we can validate it or not.

Issue #1261, opened 2024-06-30
Wrong doc and function signature for 8-bit optim


gau-nernst:
### System Info

NA

### Reproduction

Argument `optim_bits` and `amsgrad` is not used for 8-bit Adam

https://github.com/TimDettmers/bitsandbytes/blob/dada530149212d64d4b69534716202659ef37ec8/bitsandbytes/optim/adam.py#L77-L95

The document for `optim_bits` is also wrong in this case, because we are using 8 bits here.

https://github.com/TimDettmers/bitsandbytes/blob/dada530149212d64d4b69534716202659ef37ec8/bitsandbytes/optim/adam.py#L107-L110

I'm thinking of 2 ways to solve this:
- Remove `optim_bits` and `amsgrad` arguments (actually even 32-bit Adam also doesn't use `amsgrad` argument)
- If we want to keep the function signature the same, there should be a check that `optim_bits == 8` and `amsgrad == False`

### Expected behavior

NA

nighting0le01:
? so does it always run in 8 bit?? @gau-nernst

gau-nernst:
Yes, `Adam8bit` will always run in 8 bit. The `optim_bits` argument has no effect

nighting0le01:
@gau-nernst i see, thanks also, did you notice any difference varying min_8bit_size? would it make sense to start with such a high value if your hidden dim ==4096 and only goes above in projection layers?

gau-nernst:
I didn't play around with that parameter.

kritikagarg:
Hi! I'd like to work on this issue.

Looking at the current state of the code, `Adam8bit` (and `AdamW8bit`, `PagedAdam8bit`, `PagedAdamW8bit`) already have validation guards that reject non-default values for `optim_bits` and `amsgrad`. However, the docstrings are still misleading — they describe these as functional parameters with a small "Note:" appended, rather than clearly marking them as deprecated.

I'm planning to:
1. Replace the misleading docstring entries with proper `.. deprecated::` directives that clearly state these params exist only for signature compatibility and will be removed
2. Add test coverage for the Adam8bit/AdamW8bit parameter guards (currently only LAMB8bit and Adagrad8bit have guard tests)

No breaking changes — the params stay in the signature for backwards compatibility, just properly documented as deprecated.

Happy to adjust the approach based on feedback.

PR #1998, opened 2026-07-11
fix(optim): honor user-provided max_unorm in LAMB (+ 8-bit arg guards)

egeozkoc:
### Problem

The `LAMB`, `LAMB8bit`, and `LAMB32bit` constructors accept a `max_unorm` argument but pass a hardcoded `1.0` to the base optimizer:

```python
super().__init__("lamb", ..., min_8bit_size, max_unorm=1.0)  # ignores the argument
```

On the **32-bit LAMB path** (`LAMB` with the default `optim_bits=32`, and `LAMB32bit`), `max_unorm` drives the trust-ratio clipping of the update (`optimizer_update_32bit` → `_compute_update_norm_and_scale`). So the setting was silently ignored: a tighter `max_unorm` had no effect, and it could not be changed from `1.0`.

Two related inconsistencies in the 8-bit optimizers, in the same spirit as the `Adam8bit`/`AdamW8bit` guards that already exist (relates to #1261):
- `LAMB8bit` accepts `amsgrad`, which the base optimizer never uses.
- `Adagrad8bit` accepts `optim_bits`, but always runs 8-bit.

### Fix

- Thread `max_unorm` through in all three LAMB classes. Default behavior is unchanged (the default is `1.0`, which equals the previously hardcoded value).
- `LAMB8bit`: raise on `amsgrad=True`; `Adagrad8bit`: raise on `optim_bits != 8`; add docstring notes. Mirrors the existing `Adam8bit`/`AdamW8bit` guards.

### Notes

- The 8-bit blockwise update kernel does **not** apply update-norm clipping, so `max_unorm` has no numerical effect on `LAMB8bit`. The value is now stored consistently and this limitation is documented in the docstring.
- No public API changes; the guards only reject values that were already silently ignored, and no code in the repo constructs these classes with the rejected arguments.

### Tests (`tests/test_optim.py`)

- `test_lamb_max_unorm_changes_update` — behavioral: on the 32-bit path a tight vs loose `max_unorm` now yields different updates (fails on `main`, where the value was ignored). Runs on CPU.
- `test_lamb_max_unorm_threaded_to_config` — the value reaches `optimizer.args` for all three LAMB classes.
- `test_lamb8bit_rejects_amsgrad`, `test_adagrad8bit_rejects_non_8_optim_bits` — the guards.

`pre-commit run --all-files` passes; verified the new assertions fail on `main` and pass here.

Relates to #1261

### DIFF
--- a/bitsandbytes/optim/adagrad.py
+++ b/bitsandbytes/optim/adagrad.py
@@ -95,6 +95,7 @@ class Adagrad8bit(Optimizer1State):
                 The epsilon value prevents division by zero in the optimizer.
             optim_bits (`int`, defaults to 8):
                 The number of bits of the optimizer state.
+                Note: This parameter is not used in Adagrad8bit as it always uses 8-bit optimization.
             args (`object`, defaults to `None`):
                 An object with additional arguments.
             min_8bit_size (`int`, defaults to 4096):
@@ -110,6 +111,10 @@ class Adagrad8bit(Optimizer1State):
             raise ValueError("Initial accumulator value != 0.0 not supported!")
         if lr_decay != 0.0:
             raise ValueError("Lr Decay != 0.0 not supported!")
+        if optim_bits != 8:
+            # We allow the default value of 8 to maintain compatibility with the function signature,
+            # but any other value is invalid since Adagrad8bit always uses 8-bit optimization
+            raise ValueError("Adagrad8bit only supports optim_bits=8 (default value for compatibility)")
         super().__init__(
             "adagrad",
             params,
--- a/bitsandbytes/optim/lamb.py
+++ b/bitsandbytes/optim/lamb.py
@@ -60,7 +60,7 @@ class LAMB(Optimizer2State):
             optim_bits,
             args,
             min_8bit_size,
-            max_unorm=1.0,
+            max_unorm=max_unorm,
         )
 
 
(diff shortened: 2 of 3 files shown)


matthewdouglas:
Thanks for this. The 32-bit LAMB fix is the real win — `max_unorm` was stuck at `1.0` and couldn't be changed, and this fixes it. Keep the fix, the docstring correction, the test, and the guards.

The one thing I'd change is `LAMB8bit`. The 8-bit path has no update-norm clipping — the kernel just doesn't implement it — so `max_unorm` does nothing there no matter what. Threading it through doesn't change any 8-bit behavior: a user who passes `max_unorm=0.1` to `LAMB8bit` still gets a silent no-op with no error, and reasonably assumes it's doing something.

I'd rather make that explicit — raise (or warn) when a non-default `max_unorm` is passed to `LAMB8bit`, the same way you're already guarding `amsgrad`. As written, the diff is a little inconsistent: it rejects one unsupported argument (`amsgrad`) but silently accepts another (`max_unorm`). Rejecting both is cleaner.

Implementing the clip in the 8-bit kernel would be the proper fix, but that's a bigger change and out of scope here.

So: keep everything on the 32-bit side, and just have `LAMB8bit` reject a non-default `max_unorm` instead of quietly accepting it.

egeozkoc:
Done, that makes sense. LAMB8bit now raises on any non-default max_unorm, matching the amsgrad guard and the existing optim_bits pattern (the default 1.0 is allowed for signature compatibility). Tests updated to match: the config-threading test covers LAMB and LAMB32bit only, and there's a new guard test for LAMB8bit. Agreed that implementing the clip in the 8-bit kernel would be the proper fix and out of scope here.

matthewdouglas (approved):

matthewdouglas:
LGTM, thanks!

PR #1999, opened 2026-07-13
fix(parametrize): cache-gate hooks leak the global parametrize cache under activation checkpointing (all dequantized params retained)

### LINKED ISSUE #2005: Memory leak: nn.parametrize.replace_parameter_4bit + HF gradient_checkpointing_enable() — dequant cache not released on early-stopped recompute
> **Disclosure:** this bug was found during an AI-assisted debugging session and this report was drafted with AI assistance (Claude). All measurements come from real runs on our hardware; the account owner posts and vouches for this report.

### System Info

- bitsandbytes 0.49.2, transformers 5.13.1, torch 2.13.0+cu126 (CUDA 12.6)
- Windows 11, RTX 3050 8 GB
- Model: `allenai/OLMoE-1B-7B-0924`

### Context

Follow-up to #1849: transformers v5 stores fused-MoE expert weights as 3-D `nn.Parameter`s that the 4-bit walker skips, and until `Experts4bit` (#1965) lands, `bitsandbytes.nn.parametrize.replace_parameter_4bit` is the in-tree way to quantize them. That works fine for inference and for training **without** gradient checkpointing. The problem is the interaction with HF's `gradient_checkpointing_enable()` — which is the natural companion of 4-bit training on small GPUs, so anyone following the #1849 workaround path for training is likely to hit this.

**Prior-report check (why we're filing this as new):** before filing we searched GitHub issues for `replace_parameter_4bit` (zero hits in any repository) and bitsandbytes issues around gradient checkpointing / parametrize. The closest match is #1927 (INT8 CB/SCB state machine breaking under GC recompute), which is a related family but a different mechanism — that one corrupts quantization state in `autograd/_functions.py`; this one leaks the 4-bit parametrization dequant cache. We found no existing report of this leak, so we're filing it.

### Reproduction

```python
import torch
from transformers import AutoModelForCausalLM, BitsAndBytesConfig
from bitsandbytes.nn import parametrize as bnb_parametrize

bnb = BitsAndBytesConfig(
    load_in_4bit=True, bnb_4bit_compute_dtype=torch.bfloat16,
    bnb_4bit_quant_type="nf4", llm_int8_skip_modules=["gate", "lm_head"],
)
    "allenai/OLMoE-1B-7B-0924", quantization_config=bnb, device_map={"": 0},
)
# experts are fused 3-D nn.Parameters -> quantize them via parametrize (see #1849)
for layer in model.model.layers:
    bnb_parametrize.replace_parameter_4bit(layer.mlp.experts, "gate_up_proj", quant_type="nf4")
    bnb_parametrize.replace_parameter_4bit(layer.mlp.experts, "down_proj", quant_type="nf4")

# freeze everything, train only the 16 router gates (2.1M params)
for p in model.parameters():
    p.requires_grad_(False)
for n, p in model.named_parameters():
    if ".mlp.gate." in n:
        p.requires_grad_(True)

model.gradient_checkpointing_enable()   # <-- the trigger
model.config.use_cache = False
# ... then a plain training loop: AdamW, batch 1, seq len 300, LM loss
```

**Observed:** allocated CUDA memory grows monotonically every training step until OOM on the 8 GB card.
**Expected:** flat memory across steps, as with any frozen-weight training.

Removing either ingredient (no `replace_parameter_4bit`, or no `gradient_checkpointing_enable()`) makes the leak disappear.

### Root cause (confirmed from source)

`_register_parametrization_hooks` ([`bitsandbytes/nn/parametrize.py#L149-L153`](https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/bitsandbytes/nn/parametrize.py)) manages the global parametrization cache with a **pre-hook / post-hook pair**:

```python
def _enable_parametrization_cache(module, inputs):
    P._cache_enabled += 1

def _disable_parametrization_cache(module, inputs, output):
    P._cache_enabled -= 1
    if not P._cache_enabled:
        P._cache = {}
```

The post-hook is registered with plain `register_forward_hook(...)`, and **PyTorch does not run regular forward post-hooks when the module's forward raises**. Meanwhile, the non-reentrant `torch.utils.checkpoint` implements early-stopping of recomputation by raising `_StopRecomputationError` from a saved-tensor pack hook (`torch/utils/checkpoint.py`), which unwinds straight through the module's forward.

So under HF's `gradient_checkpointing_enable()` (which checkpoints whole decoder layers, so the parametrized experts module is called via `Module.__call__` *inside* the recomputed region):

1. recompute enters the experts module → pre-hook runs → `P._cache_enabled += 1`, dequantized weight gets cached in the global `P._cache`;
2. early-stop fires mid-forward → `_StopRecomputationError` unwinds → **post-hook never runs**;
3. `P._cache_enabled` is now permanently ≥ 1, so **`P._cache` is never cleared again for the rest of the process** — every subsequent forward pins its dequantized bf16 expert stacks (hundreds of MB each for fused-expert models) in the cache;
4. memory grows until OOM.

This also explains why our workaround below is leak-free: when the checkpointed region is the experts' bare `forward` function (not `Module.__call__`), the recompute path never touches the hook pair, so the counter stays balanced.

### Workaround (validated: 900+ steps, flat memory)

Checkpoint at the experts-module boundary instead of the decoder-layer boundary, and leave HF checkpointing off:

```python
import torch.utils.checkpoint as cpk

for layer in model.model.layers:
    ex = layer.mlp.experts
    orig = ex.forward

    def make_ckpt_fwd(orig):
        def fwd(*args, **kwargs):
            if torch.is_grad_enabled():
                return cpk.checkpoint(orig, *args, use_reentrant=False, **kwargs)
            return orig(*args, **kwargs)
        return fwd

    ex.forward = make_ckpt_fwd(orig)
```

With this, the same run is stable: 5.35 GB peak, ~1.0 s/step over 900 steps (router-only fine-tune of OLMoE-1B-7B on the 8 GB card).

### Suggested fix

Make the cache-disable hook exception-safe:

1. Register it with `register_forward_hook(_disable_parametrization_cache, always_call=True)` — available since torch 2.1, designed exactly for hooks that must run when forward raises; and
2. clamp the counter (`P._cache_enabled = max(0, P._cache_enabled - 1)`) so an `always_call` invocation whose matching pre-hook never ran (exception raised before pre-hooks) cannot drive it negative.

Equivalently, the pair could be replaced with a `try/finally` wrapper mirroring `torch.nn.utils.parametrize.cached()`.

Happy to share the full training script or run additional diagnostics if useful.

pjordanandrsn:
### The bug

`replace_parameter_4bit` / `replace_parameter_4bit_prequantized` register a
`forward_pre_hook`/`forward_hook` pair that gates torch's global parametrization
cache by counter: pre-hook does `parametrize._cache_enabled += 1`, post-hook
decrements and clears `parametrize._cache` when it reaches 0.

Activation checkpointing with `use_reentrant=False` (the recommended mode, and
what HF `gradient_checkpointing_enable` uses) aborts its backward recompute
mid-forward **by design** — early stop raises an internal exception the moment
the last needed activation has been rematerialized. A plain `forward_hook` is
skipped when the forward is interrupted, so for the module holding the last
recomputed save the pair fires pre-only: the global counter leaks +1 per
checkpointed region per step and never returns to zero. From then on the cache
is permanently enabled and never cleared — **every dequantized parameter the
model touches stays resident for the rest of training**, i.e. a memory leak of
the full dequantized model size (4× the packed 4-bit bytes).

Observed in the wild training MoE models whose experts are quantized via this
path with gradient checkpointing on:

- OLMoE-1B-7B: ~12.9 GiB retained (steady ~13 GiB `max_active` regardless of store); plain resident training OOMs a 16 GB card at step 0 with the leak and fits with the fix applied
- Qwen3-30B-A3B: ~53 GiB retained — training peaks at 57–60 GiB on a 96 GB card
  and **OOMs any 24 GB card** at step 1 (allocated 22.37 GiB = base + ~14 blocks
  of retained dequants, `Tried to allocate 384 MiB` inside
  `bitsandbytes/nn/parametrize.py::forward → dequantize_4bit`).

Minimal repro of the torch-level mechanism (no bnb needed): a 3-Linear chain
under `checkpoint(..., use_reentrant=False)`, pre/post hooks counting on the
tail module → `pre=2, post=1` after one fwd/bwd.

### The fix

- Register the disable hook with `always_call=True` (torch ≥ 2.0), so it also
  runs when the recompute is aborted — verified to balance the pair (2/2) under
  early stop.
- Clamp the decrement at zero. With `always_call=True` the hook can also fire
  when an *earlier* pre-hook raised before `_enable` ran; a negative counter is
  truthy, so `if not P._cache_enabled` would never clear the cache again.

### Tests

`TestParametrizationCacheCounterUnderCheckpointing`:
1. `test_counter_balanced_under_checkpoint_early_stop` — real hook pair on the
   tail module of a checkpointed chain, 3 fwd/bwd steps, asserts counter == 0 and
   cache empty. **Fails on main, passes with the fix** (verified by mutation).
2. `test_counter_never_goes_negative` — pins the clamp.

CPU-only, no quantization kernels needed; runs in ~1.2 s.

---

Developed with [Claude Code](https://claude.com/claude-code) (Anthropic) — the diagnosis, fix, and regression tests were produced with AI assistance and validated by the author as described above.

### DIFF
--- a/bitsandbytes/nn/parametrize.py
+++ b/bitsandbytes/nn/parametrize.py
@@ -127,7 +127,11 @@ def replace_parameter_4bit(
 
 
 def _disable_parametrization_cache(module: nn.Module, inputs: tuple[Any, ...], output: Any):
-    P._cache_enabled -= 1
+    # Clamp instead of a bare decrement: with ``always_call=True`` this hook also runs
+    # when the forward raised before the pre-hook incremented (e.g. an earlier pre-hook
+    # failed), and the counter must never go negative — a negative value is truthy, so
+    # ``if not P._cache_enabled`` would stop clearing the cache forever.
+    P._cache_enabled = max(0, P._cache_enabled - 1)
     if not P._cache_enabled:
         P._cache = {}
 
@@ -149,8 +153,18 @@ def _register_parametrization_hooks(module: nn.Module, param_name: str):
     # Register hooks to enable caching for the dequantization parametrization.
     # This helps preserve time and memory when the same quantized parameter
     # is accessed multiple times in the forward computation.
+    #
+    # ``always_call=True`` is load-bearing: activation checkpointing with
+    # ``use_reentrant=False`` aborts its backward recompute mid-forward by design
+    # (early stop, via an internal exception) once the last needed activation has
+    # been rematerialized. A plain forward hook is skipped in that case, so the
+    # global ``parametrize._cache_enabled`` counter leaks upward once per
+    # checkpointed region per step, after which the cache is enabled (and never
+    # cleared) for the remainder of training — every dequantized parameter this
+    # module produces stays resident, i.e. a memory leak of the full dequantized
+    # model size (4x the packed 4-bit bytes).
     module.register_forward_pre_hook(_enable_parametrization_cache)
-    module.register_forward_hook(_disable_parametrization_cache)
+    module.register_forward_hook(_disable_parametrization_cache, always_call=True)
 
 
 def _parametrized_state_dict_post_hook(
--- a/tests/test_parametrize.py
+++ b/tests/test_parametrize.py
@@ -431,3 +431,69 @@ def test_gradient_behavior(device, dtype):
     # The dequantized output should also not require gradients
     reconstructed = module.weight_2d
     assert not reconstructed.requires_grad, "Dequantized parameter should not require gradients"
+
+
+class TestParametrizationCacheCounterUnderCheckpointing:
+    """The cache-gate hooks must stay balanced under activation checkpointing.
+
+    ``use_reentrant=False`` checkpointing aborts its backward recompute mid-forward by
+    design (early stop) once the last needed activation is rematerialized. A plain
+    forward hook is skipped for the module holding that last save, so the global
+    ``parametrize._cache_enabled`` counter leaked +1 per checkpointed region per step,
+    the cache was never cleared again, and every dequantized parameter stayed resident
+    (4x the packed model bytes). ``always_call=True`` keeps the pair balanced.
+    """
+
+    def _reset(self):
+        import torch.nn.utils.parametrize as P
+
+        P._cache_enabled = 0
+        P._cache = {}
+
+    def test_counter_balanced_under_checkpoint_early_stop(self):
+        import torch.nn.utils.parametrize as P
+        from torch.utils.checkpoint import checkpoint
+
+        from bitsandbytes.nn.parametrize import _register_parametrization_hooks
+
+        self._reset()
+
+        class Chain(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.a = nn.Linear(8, 8)
+                self.tail = nn.Linear(8, 8)
+
+            def forward(self, x):
+                return self.tail(self.a(x))
+
+        m = Chain()
+        # Register the real hook pair on the tail module — the one whose recompute
+        # holds the last save, where early stop aborts before the forward hook.
+        # (torch < 2.5 skips the state-dict hook inside; the cache pair is what we need.)
+        _register_parametrization_hooks(m.tail, "weight")
+
+        for _ in range(3):
+            x = torch.randn(2, 8, requires_grad=True)
+            checkpoint(m, x, use_reentrant=False).sum().backward()
+
+        assert P._cache_enabled == 0, (
+            f"parametrize cache counter leaked to {P._cache_enabled}; dequantized "
+            "parameters would be retained for the rest of training"
+        )
+        assert P._cache == {}
+        self._reset()
+
+    def test_counter_never_goes_negative(self):
+        import torch.nn.utils.parametrize as P
+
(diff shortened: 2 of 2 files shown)


matthewdouglas:
Thanks for the PR. This should fix #2005. 

One thing I should maybe note is that I never really took the effort to highlight this feature in release notes, examples, demos, etc mainly for the reason that it was just a sort of experiment. Overall, it's not an ideal solution for MoEs, even if it sort of works. I think Axolotl might use it in some way, so the PR here is OK.

pjordanandrsn:
Thanks for the quick review and merge. Hit it benchmarking Qwen3-30B QLoRA on a 24 GB 3090, where the retained dequant cache quietly erases most of 4-bit's memory win — glad it lands before anyone else meets #2005's path at scale. Agreed on the MoE fit of parametrize generally: the recompute interaction is intrinsic to caching dequantized params under checkpointing, rather than something a patch fully cures.

PR #2001, opened 2026-07-14
Fix Lion to use decoupled weight decay in the Triton 32-bit kernel

eaglstun:
Follow-up to #1993, which fixed the same coupled-vs-decoupled weight-decay bug for Lion in the default backend and the CUDA 32-bit kernel. matthewdouglas invited a Triton follow-up when approving #1993.

_optimizer_update_1state_32bit_triton_kernel folded weight decay into the gradient (coupled L2) for every 1-state optimizer, Lion included. Lion requires *decoupled* (AdamW-style) decay applied to the param directly (p *= 1 - lr*wd), outside the sign update (Chen et al. 2023); folding decay into the gradient corrupts the input to sign(), changing the update direction, not just its magnitude. This mirrors the CUDA 32-bit fix exactly: exclude LION (id 4) from the gradient fold, and shrink the param in the LION branch before the sign update.

The existing test_lion32bit_weight_decay (added in #1993) is parametrized over get_available_devices(), so it already exercises this kernel on XPU hardware in CI. Like #1993's CUDA change, the Triton kernel is XPU-only and cannot be built or run on the available hardware.

### DIFF
--- a/bitsandbytes/backends/triton/kernels_optim.py
+++ b/bitsandbytes/backends/triton/kernels_optim.py
@@ -270,7 +270,11 @@ def _optimizer_update_1state_32bit_triton_kernel(
     s1_vals = tl.load(state1_ptr + offsets, mask=mask, other=0.0)
 
     g_vals = gnorm_scale * g_vals
-    if weight_decay > 0.0:
+    # Coupled (L2) weight decay: fold wd into the gradient. This is correct for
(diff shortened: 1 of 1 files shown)


matthewdouglas:
@jiqing-feng Would you mind checking the optimizer tests for this on XPU? At first glance it looks correct to me. Thanks in advance!

jiqing-feng:
Verified it's okay on XPU with the command `python -m pytest tests/test_optim.py -k "xpu and lion" -q`:
image

PR #2012, opened 2026-07-19
fix(rocm): avoid SIMT workgroup-count resonance

sstamenk:
Add one bounds-checked workgroup when a large HIP SIMT grid is an exact multiple of 256, avoiding scheduler underutilization observed on RDNA3 and RDNA4.

The extra workgroup is guarded and does no work:
[csrc/gemm_4bit_simt.cu#L196-L197](https://github.com/sstamenk/bitsandbytes/blob/worktree-camping-mainline/csrc/gemm_4bit_simt.cu#L196-L197)
```
if (warp_n >= N)
    return;
```

This addresses the suboptimal performance on certain shapes that was originally noticed in https://github.com/bitsandbytes-foundation/bitsandbytes/pull/1979 and achieves up to 3.5x performance improvement at M=1 for specific shapes across fp16, bf16 and fp32.

Not exactly sure of the mechanism that causes the lower-than-expected performance.
### fp16
gfx1201_fp16_strongest

### bf16
gfx1201_bf16_strongest

### fp32
gfx1201_fp32_strongest

### DIFF
--- a/csrc/common.cuh
+++ b/csrc/common.cuh
@@ -4,10 +4,23 @@
 
 #include "compat.cuh"
 
+// AMD GPU architecture detection. Use explicit device macros rather than the
+// broad __GFX9__/__GFX11__/__GFX12__ macros, which span distinct architectures.
+#define IS_RDNA3 (defined(__gfx1100__) || defined(__gfx1101__) || defined(__gfx1102__) || defined(__gfx1103__))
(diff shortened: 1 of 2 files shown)


matthewdouglas:
The test failure is unrelated. LGTM, thanks!

matthewdouglas (approved):

PR #2018, opened 2026-07-23
Fix int8 GEMM by using dedicated BLAS Lt handle on ROCm/CUDA

zjin-lcf:
## Summary
- Fix int8 GEMM on ROCm (and CUDA) by creating a dedicated `(hip|cublas)Lt` handle in `Context` instead of casting the regular BLAS handle
- Route `cigemmlt_32`, `cigemmlt_8`, and `cigemmlt_8_rowscale` through `m_lt_handle` for `igemmlt`
- Add `igemmlt_32_gemmex_fallback` for ROCm builds without hipBLASLt (`NO_HIPBLASLT`)
- Skip CUDA-only `BNB_CUDA_VERSION` evaluator tests on HIP; add int8 regression tests

## Problem
`cigemmlt_*` passed `Context::m_handle` (a `rocblas_handle` on HIP) cast to `hipblasLtHandle_t`. hipBLASLt then ran with invalid state, causing OOB GPU writes, corrupted quant stats, and 62 test failures on gfx950 (50 Linear8bitLt, 10 int8 functional, 2 CUDA-only evaluator tests).

## Test plan
- [x] Re-run all 62 originally failing tests on gfx950/ROCm 7.2: **60 passed, 2 skipped**
- [x] Full suite: `pytest tests/ -v --tb=short -n 4` → **3452 passed, 16 skipped, 32 xfailed, 0 failed** (~96s)
- [ ] CI CUDA matrix (same latent handle bug exists on CUDA path)

### DIFF
--- a/csrc/ops.cu
+++ b/csrc/ops.cu
@@ -8,8 +8,6 @@
 #include <limits>
 #include <ops.cuh>
 
-#define ERR_NOT_IMPLEMENTED 100
-
 #if BNB_HIP
 #include <atomic>
 #include <hip/hip_runtime.h>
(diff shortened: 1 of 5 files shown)


matthewdouglas:
Hi,

Thanks for raising this issue and the PR.

I don't think this issue is valid on the CUDA side, but looks valid on ROCm. Additionally, I'm planning on removing the `NO_HIPBLASLT` related code as we don't build with that option anymore. 

I want to take a moment to think a little more about this. I will likely put up a PR or two for the deprecations/removals.

PR #2023, opened 2026-07-27
Add ROCm CDNA5 (gfx1250) support

zjin-lcf:
## Summary

Adds ROCm CDNA5 (gfx1250) support to bitsandbytes.

- Include `gfx1250` in the ROCm 7.14+ build architecture list (`.github/scripts/build-rocm.sh`).
- Add an `IS_CDNA5` macro (`gfx1250`/`gfx1251`) in `csrc/common.cuh`. CDNA5 runs in **Wave32** mode and is deliberately kept **out** of `IS_CDNA` (Wave64), so the default warp size stays 32 (unchanged from upstream).
- Route CDNA5 into the fp32-FMA 4-bit GEMM path (`IS_CDNA || IS_CDNA5`) rather than the RDNA `v_dot2` path, which CDNA5 does not implement (no `dot12-insts` / `v_dot2_f32_*` builtins — they fail to compile for gfx1250).
- Route `gfx125x` through the existing gfx12 SIMT-vs-dequant crossover (`M<=8`) in `bitsandbytes/backends/cuda/ops.py`; dedicated calibration is left as a follow-up.
- Document `gfx1250` in `README.md` and `docs/source/installation.mdx`.

## Rationale

Despite its `gfx125x` target number, CDNA5 behaves like CDNA (Wave32 but no SIMT dot instructions), so it needs the fp32-FMA math path for correct/accurate bf16/fp16 4-bit GEMM rather than the RDNA `v_dot2` path.

## Test plan

- [ ] Build with ROCm 7.14+ including `gfx1250` in the arch list.

### DIFF
--- a/.github/scripts/build-rocm.sh
+++ b/.github/scripts/build-rocm.sh
@@ -40,6 +40,11 @@ if rocm_version_at_least "7.14"; then
     bnb_rocm_arch="${bnb_rocm_arch};gfx908;gfx1030;gfx1031;gfx1032;gfx1033;gfx1034;gfx1035;gfx1036"
 fi
 
+# ROCm 7.14+ - Add CDNA5 (gfx1250).
+if rocm_version_at_least "7.14"; then
+    bnb_rocm_arch="${bnb_rocm_arch};gfx1250"
+fi
+
 if [ "${RUNNER_OS}" == "Linux" ]; then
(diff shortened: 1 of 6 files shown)


matthewdouglas (approved):

matthewdouglas:
LGTM, thanks!

PR #2031, opened 2026-08-02
Warn once, and only when it matters, for unaligned gemm_4bit

### LINKED ISSUE #2027: `gemm_4bit`: blocksize-alignment warning emitted on every call, and now also during training (new in 0.50.0)
### System Info

```
bitsandbytes: 0.50.0, and also reproduced on current main (0.50.1.dev0, continuous-release_main
              wheel bitsandbytes-1.33.7.preview, main @ 1f4007e)
torch:        2.13.0+cu130 (CUDA 13.0)
python:       3.10.18
GPU:          NVIDIA H100 80GB HBM3 (sm90)
transformers: 5.15.0.dev0, peft: 0.20.0
platform:     Linux x86_64
```

### Reproduction

Minimal, bitsandbytes only. The shape is taken from a real checkpoint: the Qwen2.5-VL vision-tower `down_proj`, `Linear(3420 -> 1280)`, where `3420 % 64 == 28`:

```python
import warnings

import torch

K, N = 3420, 1280
W = torch.randn(N, K, dtype=torch.float16, device="cuda")
Wq, state = F.quantize_4bit(W, blocksize=64, quant_type="nf4")
A = torch.randn(16, K, dtype=torch.float16, device="cuda")

with warnings.catch_warnings(record=True) as rec:
    warnings.simplefilter("always")
    for _ in range(10):
        torch.ops.bitsandbytes.gemm_4bit.default(
            A, Wq, state.shape, state.absmax, state.blocksize, state.quant_type
        )

print(len([w for w in rec if "not aligned" in str(w.message)]))  # -> 10
```

The warning comes from the `bitsandbytes::gemm_4bit` dispatch added in #1949 — `bitsandbytes/backends/cuda/ops.py:948` on main, with the same warning at `bitsandbytes/backends/xpu/ops.py:190`:

```python
if M > _gemm_4bit_custom_max_m:
    use_custom = False
elif K % blocksize != 0:
    warn(
        f"inner dimension ({K}) is not aligned for fast kernel "
        f"with blocksize={blocksize}, falling back to slower implementation.",
        UserWarning,
    )
    use_custom = False
else:
    use_custom = _gemm_4bit_use_custom_fn(A.device.index, A.dtype, M, N, K)
```

Two things make this noisier than it may look:

**1. It now also fires during training, which it did not before 0.50.0.**

Previously the check lived in `matmul_4bit` (`bitsandbytes/autograd/_functions.py`) and was gated to the inference gemv path (`A.numel() == A.shape[-1] and A.requires_grad == False`), so training never reached it. In 0.50.0, `matmul_4bit` routes every forward with `needs_grad == False` to `gemm_4bit` for all `M`, and in a QLoRA setup that includes many training-time forwards:

- gradient-checkpointed forwards — PEFT's `prepare_model_for_kbit_training()` enables gradient checkpointing, and the checkpointed region runs with grad disabled;
- no-grad reference-model forwards (e.g. DPO/KTO).

Concretely, in TRL's PEFT + 4-bit tests (2-layer test model, 14 quantized `Linear4bit`, one training step): **0** such warnings with 0.49.2 and **26** with 0.50.0 — one per misaligned layer per grad-free forward. Full details: https://github.com/huggingface/trl/issues/6580

(In #1949 you mentioned that the new kernels would not typically be dispatched for QLoRA training; these grad-free forwards look like the case where they are.)

**2. For several real checkpoints the dimension comes from the architecture, so the message is not actionable.**

Vision towers whose `intermediate_size` is not a multiple of 64, hence a misaligned `down_proj`:

| checkpoint(s) | vision `intermediate_size` | `% 64` |
| --- | --- | --- |
| `Qwen/Qwen2.5-VL-3B/7B-Instruct` | 3420 | 28 |
| `Qwen/Qwen3-VL-8B-Instruct`, `HuggingFaceTB/SmolVLM(2)`, `google/siglip-so400m-patch14-384` (so also Idefics3 / PaliGemma-2 / Gemma-3 towers) | 4304 | 16 |

transformers' bitsandbytes quantizer does not exclude vision towers by default (`quantizers/base.py`: default skips are `get_keys_to_not_convert()` plus `_keep_in_fp32_modules`), so a plain QLoRA fine-tune of these checkpoints warns about a "slower implementation" that the user cannot avoid — the only knob would be the blocksize, and no blocksize divides 3420 or 4304. Since the warning has no `stacklevel`, it also points at `bitsandbytes/backends/cuda/ops.py` rather than the caller, and does not say which module is affected.

### Expected behavior

I'm not suggesting the message should go away — the information is useful once. A few options, all matching patterns already used in the codebase:

1. Emit at most once per `(K, blocksize)` (or once per process) rather than once per call — e.g. the `_cpu_paged_warned` guard style in `bitsandbytes/optim/optimizer.py:376`, or a `@functools.cache`d helper like the other cached helpers in `backends/cuda/ops.py`.
2. Pass `stacklevel=` so the warning points at the caller, as done at `backends/cuda/ops.py:159`, `autograd/_functions.py:87` and `:444`, and `optim/optimizer.py:379`. Adding the shape (`N`, `K`) to the message would also help users locate the layer.
3. Consider moving the alignment check into the `else` branch, i.e. warn only when the custom kernel would otherwise have been selected. Today the check runs before `_gemm_4bit_use_custom_*`, so the warning also appears where the heuristic would have returned `False` regardless of alignment (e.g. `A.dtype == torch.float32` with `M >= 8`, sm87/sm110, sm100 with large `N`).
4. Apply the same change to the XPU copy at `backends/xpu/ops.py:190`.
5. Optionally demote it to a `logging` call (logging is already used in `utils.py`, `cextension.py`, `cuda_specs.py`), given that for architecture-fixed dimensions there is no user action to take.

Correctness and performance are not in question here: the dequantize + `F.linear` fallback is exactly the path ≤ 0.49.2 used for all 4-bit training, so this is only about the warning.

Happy to open a PR for 1 + 2 + 4 if that sounds reasonable.

guptaishaan:
Fixes #2027

In the CUDA `gemm_4bit` dispatch the `K % blocksize != 0` check sat in an `elif` ahead of the kernel-selection heuristic, so any call with `M <= _gemm_4bit_custom_max_m` (1536) and a misaligned `K` warned, even though the heuristic caps the fused kernel at `M <= 512` and would have taken the `dequant + F.linear` fallback anyway. That is why the warning shows up during training, where `M` is `batch * seq_len`. On top of that it was emitted per call, so a model whose `K` is inherently misaligned (Qwen2.5-VL vision tower, `K = 3420`) got one warning per forward.

Changes:

- `backends/cuda/ops.py`: run the heuristic first and warn only when misalignment is what actually cost us the fused kernel.
- `backends/utils.py`: the warning body moves into a `functools.cache`d `_warn_gemm_4bit_unaligned(K, blocksize)`, so it fires at most once per shape per process.
- `backends/xpu/ops.py`: calls the same helper. The warning text there was a verbatim copy.

No behaviour change beyond the warning. Every call that used to warn took the fallback then and takes it now, so numerics are identical.

Verified on an A40 (sm_86), CUDA 12.6, torch 2.13.0+cu126. With the issue's repro (`K = 3420`, `blocksize = 64`, 10 calls) the counts go from 10/10/10 to 1/0/0 for M=1 inference, batched, and a `Linear4bit` training forward+backward. New test `tests/test_ops.py::Test4bitBlockwiseQuantOps::test_gemm_4bit_unaligned_warning` fails on CUDA before the patch and passes after. `tests/test_ops.py` and the 4-bit subset of `tests/test_functional.py` pass, and the pre-commit hooks pass on the changed files.

Not verified on hardware I do not have: the XPU change (mechanical substitution, no control-flow change) and ROCm, which shares the CUDA dispatch and can only see fewer warnings after this. Arch coverage is sm_86 only. The fix keys on the heuristic's return value rather than any threshold, so it is arch-independent, but the exact `M` at which warnings stop is not.

Thanks to @albertvillanova for the report and the precise diagnosis.

### DIFF
--- a/tests/test_ops.py
+++ b/tests/test_ops.py
@@ -1,4 +1,5 @@
 from math import prod
+import warnings
 
 import pytest
 import torch
@@ -332,6 +333,28 @@ class Test4bitBlockwiseQuantOps:
                 (A_op, B_q, list(B.shape), qs.absmax, blocksize, quant_type),
                 kwargs={"bias": bias},
             )
+
+    @pytest.mark.parametrize("device", get_available_devices())
+    def test_gemm_4bit_unaligned_warning(self, device):
+        """Regression test for #2027: the blocksize-alignment warning must not be emitted
+        on every call, nor at all when the fused kernel was not going to be used anyway."""
+        N, K, blocksize = 128, 3420, 64  # 3420 % 64 != 0 (Qwen2.5-VL vision tower)
+        B = torch.randn(N, K, dtype=torch.float16, device=device)
+        B_q, qs = bitsandbytes.functional.quantize_4bit(B, blocksize=blocksize, quant_type="nf4")
+
+        def run(M):
+            A = torch.randn(M, K, dtype=torch.float16, device=device)
+            with warnings.catch_warnings(record=True) as caught:
+                warnings.simplefilter("always")
+                for _ in range(2):
+                    torch.ops.bitsandbytes.gemm_4bit.default(A, B_q, list(B.shape), qs.absmax, blocksize, "nf4")
+            return [w for w in caught if "not aligned" in str(w.message)]
+
+        # Large M always takes the dequant+F.linear path, aligned or not.
+        assert run(1024) == []
+
+        # When alignment does decide it, warn at most once per (K, blocksize).
+        assert len(run(1)) <= 1
 
     @pytest.mark.parametrize("device", get_available_devices())
     @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16], ids=describe_dtype)
--- a/bitsandbytes/backends/cuda/ops.py
+++ b/bitsandbytes/backends/cuda/ops.py
@@ -3,7 +3,6 @@ import ctypes as ct
 import functools
 from math import prod
 from typing import Optional
-from warnings import warn
 
 import torch
 
@@ -11,6 +10,7 @@ from bitsandbytes.functional import CUBLAS_Context, _cuda_device_of, get_ptr
 
 from ..._ops import register_kernel
 from ...cextension import lib
+from ..utils import _warn_gemm_4bit_unaligned
 
 
 def _setup_ctypes(names, argtypes, restype=None):
@@ -943,15 +943,13 @@ def _(
     # fallback.
     if M > _gemm_4bit_custom_max_m:
         use_custom = False
-    elif K % blocksize != 0:
-        warn(
-            f"inner dimension ({K}) is not aligned for fast kernel "
-            f"with blocksize={blocksize}, falling back to slower implementation.",
-            UserWarning,
-        )
-        use_custom = False
     else:
         use_custom = _gemm_4bit_use_custom_fn(A.device.index, A.dtype, M, N, K)
+        if use_custom and K % blocksize != 0:
+            # Only warn when misalignment is what costs us the fused kernel; when the
+            # heuristic picks the fallback anyway (larger M, e.g. training), it doesn't.
+            _warn_gemm_4bit_unaligned(K, blocksize)
+            use_custom = False
 
     if not use_custom:
         return _dequant_linear_fallback(
--- a/bitsandbytes/backends/utils.py
+++ b/bitsandbytes/backends/utils.py
@@ -1,4 +1,6 @@
+import functools
 from importlib.metadata import metadata
+from warnings import warn
 
 from packaging import version
 import torch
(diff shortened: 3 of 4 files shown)


tests/test_ops.py
```diff
@@ -333,6 +334,28 @@ def test_gemm_4bit(self, device, dtype, quant_type, compress_statistics, has_bia
...
+                for _ in range(2):
+                    torch.ops.bitsandbytes.gemm_4bit.default(A, B_q, list(B.shape), qs.absmax, blocksize, "nf4")
+            return [w for w in caught if "not aligned" in str(w.message)]
+
+        # Large M always takes the dequant+F.linear path, aligned or not.
+        assert run(1024) == []
+
+        # When alignment does decide it, warn at most once per (K, blocksize).
+        assert len(run(1)) <= 1
+
```
matthewdouglas, line 358: I don't think this test is entirely necessary. To me it's noise. I understand we're testing this changed logging behavior, but I still don't think the value is really there. Especially since this is written to run on all devices too. I would prefer to just remove it.

matthewdouglas:
Thanks for the PR. On the issue I had commented that I would like the warning to be emitted per `(N, K, blocksize)` combination. Can you please make this adjustment?

Additionally, I was leaning on going to use `logger` instead of `warnings` but I'm not super particular on that point.

PR #2032, opened 2026-08-03
fix: missing compilation flag on  recent msvc version

mfuntowicz:
In recent MSVC compiler versions it needs to specify which preprocessor to use. 

This PR addresses this adding the required flag for MSVC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,12 +338,12 @@ if(WIN32)
 endif()
 
 if(MSVC)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast /Zc:preprocessor")
+
     # /arch:AVX2 is only valid for x86/x64 targets, not ARM64
     string(TOLOWER "${CMAKE_SYSTEM_PROCESSOR}" _msvc_arch)
     if(_msvc_arch MATCHES "x86|x64|amd64")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2 /fp:fast")
-    else()
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -338,7 +338,13 @@ if(WIN32)
 endif()
 
 if(MSVC)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast /Zc:preprocessor")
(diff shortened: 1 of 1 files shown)


Titus-von-Koeller:
Thanks for this — the `/fp:fast` de-duplication is a nice simplification and behavior-identical, so that part is good to go. Two things on the new flag, plus one ask.

**1. It also hits compilers that don't want it.** CMake's `MSVC` variable is true for any MSVC-*emulating* compiler, which includes `icx` (used by the XPU backend) and `clang-cl`. Clang's preprocessor is already conformant, so the flag is rejected. This PR's own `build-xpu (windows-2022, 2026)` job logs it on every translation unit:

```
icx: warning: argument unused during compilation: '-Zc:preprocessor' [-Wunused-command-line-argument]
```

Suggest gating on the compiler ID instead:

```cmake
if(MSVC)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /fp:fast")

    # Only real cl.exe needs (and accepts) the conformant preprocessor switch;
    # clang-cl / icx are already conformant and warn on it.
    if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /Zc:preprocessor")
    endif()

    # /arch:AVX2 is only valid for x86/x64 targets, not ARM64
    string(TOLOWER "${CMAKE_SYSTEM_PROCESSOR}" _msvc_arch)
    if(_msvc_arch MATCHES "x86|x64|amd64")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX2")
    endif()
endif()
```

**2. It may not fix the failure it's aimed at.** The flag lands in `CMAKE_CXX_FLAGS`, which never reaches the `.cu` translation units — CMake compiles those with `CMAKE_CUDA_FLAGS`. Given the branch name (`fix/cuda_121_win`), if the error was in a CUDA file this patch is a no-op for it. The CUDA-side equivalent already exists, but is gated to CUDA 13.2+:

https://github.com/bitsandbytes-foundation/bitsandbytes/blob/a2b90e6/CMakeLists.txt#L197-L200

If the failure is nvcc-side, loosening that gate (e.g. keying on `MSVC_VERSION` rather than the CUDA version) is the actual fix. Either way, the two halves of the build currently disagree on preprocessor conformance for CUDA < 13.2, which is worth aligning.

**3. Could you add the error to the description?** All Windows jobs are green on `main` — including `build-cpu (windows-2025)` and `build-cuda (windows-2025, 12.1.1)` — so nothing in CI reproduces this, and it's hard to tell which of the two fixes above is correct. The failing file, the full error text, and your `cl.exe` + CUDA versions would settle it.

If it turns out to be `BNB_OMP_PARALLEL_FOR` in `csrc/cpu_ops.cpp` (it uses `_Pragma("omp parallel for")`, historically MSVC-conformant-preprocessor-only), then `__pragma(omp parallel for)` guarded by `defined(_MSC_VER) && !defined(__clang__)` would be a more surgical fix than a global conformance switch — though I'm fine with the switch too.

PR #2040, opened 2026-08-14
Fix decoupled weight decay ordering in the CUDA Adam/AdEMAMix kernels

### LINKED ISSUE #2010: Adam/AdamW/LAMB/AdEMAMix: weight decay applied in wrong order in default and Triton backends, diverging from CUDA kernel
### System Info

bitsandbytes 0.49.2 (current PyPI release, `pip install bitsandbytes`), Python 3.12.10, Windows, CPU only (`torch` 2.x). Also confirmed present on the current `main` branch source. No GPU/CUDA needed to reproduce, since this is about the `default` and `triton` backend implementations diverging from the CUDA reference kernel, not about CUDA itself.

### Reproduction

For Adam/AdamW/LAMB (`optimizer_id == 3`, i.e. `adam`/`lamb` share the same code path) and AdEMAMix (`optimizer_id == 5`), the `default` backend (`bitsandbytes/backends/default/ops.py`, used for MPS and any device without a dedicated compiled kernel) and the `triton` backend (`bitsandbytes/backends/triton/kernels_optim.py`) apply weight decay to the parameter **before** adding the optimizer update, while the CUDA reference kernel (`csrc/kernels.cu`) applies it **after**. This is a real, systematic numerical divergence between backends for the same optimizer with the same hyperparameters, not a floating-point rounding artifact.

CUDA reference, `csrc/kernels.cu` lines 700-710 (`kOptimizer32bit2State`, `case ADAM`):
```cpp
s1_vals[j] = s1_vals[j] * beta1 + ((1.0f - beta1) * (float)g_vals[j]);
s2_vals[j] = s2_vals[j] * beta2 + ((1.0f - beta2) * (float)g_vals[j] * (float)g_vals[j]);
p_vals[j] = ((float)p_vals[j]) +
            (update_scale * step_size * (s1_vals[j] / (sqrtf(s2_vals[j]) + (eps * correction2))));
if (weight_decay > 0.0f)
    p_vals[j] = ((float)p_vals[j]) * (1.0f - (lr * weight_decay));   // decay AFTER the update
```
i.e. `p_new = (p_old + step_size·update) · (1 − lr·wd)`. The same "decay after the update" placement is used in the 8-bit blockwise CUDA kernel (`kOptimizerStatic8bit2StateBlockwise`, lines 1097-1104) and for ADEMAMIX (lines 690-697).

`default` backend, `bitsandbytes/backends/default/ops.py` lines 460-471 (installed package, `optimizer_id == 3`):
```python
if optimizer_id == 3:  # ADAM
    s1_vals = state1 * beta1 + (1.0 - beta1) * g_vals
    s2_vals = state2 * beta2 + (1.0 - beta2) * g_vals * g_vals
    correction1 = 1.0 - beta1**step
    correction2 = sqrt(1.0 - beta2**step)
    step_size = -lr * correction2 / correction1

    if weight_decay > 0.0:
        p_vals = p_vals * (1.0 - lr * weight_decay)   # decay BEFORE the update

    update_val = update_scale * step_size * (s1_vals / (torch.sqrt(s2_vals) + eps * correction2))
    p_vals = p_vals + update_val
```
i.e. `p_new = p_old·(1 − lr·wd) + step_size·update`. The Triton backend (`backends/triton/kernels_optim.py` lines 196-211) has the identical ordering, and so does ADEMAMIX in both files. I confirmed all of this directly against the files shipped inside the installed `bitsandbytes==0.49.2` wheel, not just against `main`.

The two expressions expand to:
- CUDA: `p_old·(1−lr·wd) + step_size·update·(1−lr·wd)`
- default/triton: `p_old·(1−lr·wd) + step_size·update`

They differ by a factor of `(1−lr·wd)` on the update term, a genuine per-step bias of magnitude `≈ |update|·lr·wd`, so anyone training with `weight_decay > 0` (the normal AdamW case) gets a measurably different trajectory depending on whether their run happens to execute on the CUDA kernel vs. the `default`/Triton backend.

I verified this numerically with a standalone script implementing both formulas exactly as shown above (CPU only, `torch` tensors, no bitsandbytes compilation needed):

```python
import math, torch

def cuda_step(p, g, s1, s2, step, lr, beta1, beta2, eps, weight_decay):
    correction1 = 1.0 - beta1**step
    correction2 = math.sqrt(1.0 - beta2**step)
    step_size = -lr * correction2 / correction1
    s1.mul_(beta1).add_(g, alpha=1 - beta1)
    s2.mul_(beta2).addcmul_(g, g, value=1 - beta2)
    p.add_(step_size * s1 / (s2.sqrt() + eps * correction2))
    if weight_decay > 0.0:
        p.mul_(1.0 - lr * weight_decay)  # matches csrc/kernels.cu:708-709

def default_backend_step(p, g, s1, s2, step, lr, beta1, beta2, eps, weight_decay):
    correction1 = 1.0 - beta1**step
    correction2 = math.sqrt(1.0 - beta2**step)
    step_size = -lr * correction2 / correction1
    if weight_decay > 0.0:
        p.mul_(1.0 - lr * weight_decay)  # matches backends/default/ops.py:468-469
    s1.mul_(beta1).add_(g, alpha=1 - beta1)
    s2.mul_(beta2).addcmul_(g, g, value=1 - beta2)
    p.add_(step_size * s1 / (s2.sqrt() + eps * correction2))

lr, wd, beta1, beta2, eps = 1e-2, 0.1, 0.9, 0.999, 1e-8
n = 4096
torch.manual_seed(0)
p0 = torch.randn(n)
p_cuda, s1_cuda, s2_cuda = p0.clone(), torch.zeros(n), torch.zeros(n)
p_def, s1_def, s2_def = p0.clone(), torch.zeros(n), torch.zeros(n)

torch.manual_seed(1)
for step in range(1, 501):
    g = torch.randn(n) * 0.01
    cuda_step(p_cuda, g.clone(), s1_cuda, s2_cuda, step, lr, beta1, beta2, eps, wd)
    default_backend_step(p_def, g.clone(), s1_def, s2_def, step, lr, beta1, beta2, eps, wd)

print("max abs diff:", (p_cuda - p_def).abs().max().item())
print("mean abs diff:", (p_cuda - p_def).abs().mean().item())
```

Output:
```
max abs diff: 0.0006633400917053223
mean abs diff: 0.00014304943033494055
```

As a control, running the same loop with `weight_decay=0.0` gives `max abs diff: 0.0` exactly, both formulas are identical when there's no decay, which isolates the discrepancy to the weight-decay term specifically.

### Expected behavior

The `default` and `triton` backends should produce (up to float precision) the same optimizer trajectory as the CUDA kernel for the same optimizer and hyperparameters. Weight decay should be applied in the same order (after the update, per the CUDA reference) in all backends.

### Not a duplicate of #1992

#1992 ("Lion applies coupled (L2) weight decay instead of decoupled in 32-bit CUDA, Triton, and the default backend") is about Lion specifically using the wrong *kind* of weight decay (coupled/L2 folded into the gradient vs. decoupled/AdamW-style applied to the parameter). That was fixed in #1993. This is a different bug, affecting Adam/AdamW/LAMB/AdEMAMix specifically, both of which already correctly use decoupled decay in every backend, the issue here is the *order* decoupled decay is applied relative to the update, not which kind of decay is used. I checked PR #1993's diff and confirmed it didn't touch the ADAM/ADEMAMIX branches in any of these files.

### Additional context

I also checked the CPU-specific backend (`bitsandbytes/backends/cpu/ops.py`) on the current `main` branch, it has the same decay-before-update bug for Adam/LAMB/AdEMAMix as well, but that particular file's optimizer support doesn't exist yet in the released 0.49.2 wheel (it's newer, in-development code), so I'm not claiming that part is live for released-package users today, only that it has the identical bug pattern and would need the same fix once it ships. The `default` and `triton` findings above, by contrast, are confirmed present in the current stable release.

yentur:
Closes #2010.

The CUDA 2-state optimizer kernels apply weight decay after adding the update:

```
p = (p + step_size*update) * (1 - lr*wd)
```

which expands to `p*(1 - lr*wd) + step_size*update*(1 - lr*wd)`. The extra factor on the update term is not in AdamW Algorithm 2 (Loshchilov and Hutter, arXiv:1711.05101) or in the AdEMAMix update rule (Pagliardini et al., arXiv:2409.03137). Both take the decay and the gradient-based update from the same previous parameter, which rearranges to `p*(1 - lr*wd) + step_size*update`. `torch.optim.AdamW` does the same, and so do the cpu, default and triton backends here.

Three sites in `csrc/kernels.cu`: the `ADAM` and `ADEMAMIX` branches of `kOptimizer32bit2State`, and the 2-state branch of `kOptimizerStatic8bit2StateBlockwise`. The change is a reorder, 9 insertions and 8 deletions. The 1-state kernels already use coupled L2 for momentum/rmsprop/adagrad and decoupled decay for Lion, matching the Python backends, so they are untouched.

@ErenAta16 reported this and worked out in the thread that CUDA is the side that diverges. @egeozkoc confirmed it against the paper, pointed out that the 8-bit blockwise kernel is affected too, and offered to write the expanded weight-decay tests. I went ahead because the issue had been sitting for about a month. Happy to close this if either of you would rather carry it.

## This changes results for existing users

CUDA runs using Adam, AdamW, LAMB or AdEMAMix with `weight_decay > 0` will follow different training trajectories after this. The per-step difference is `lr*wd*|update|`. Runs with `weight_decay = 0` are unaffected, as are the cpu, default, triton and XPU backends, which already used this ordering.

## Why the current tests do not catch it

They run these optimizers at the default `weight_decay` of 0, where the two orderings coincide. At the lr=1e-3, wd=0.01 an Adam user would typically pick, 20 steps leave the orderings 1.7e-7 apart, inside the 1e-6 tolerance. The entire existing `test_optim.py` passes on the unfixed kernel:

```
$ BNB_TEST_DEVICE=cuda pytest tests/test_optim.py -q -k "not weight_decay"   # unfixed kernel
213 passed, 12 skipped, 35 deselected, 194 warnings in 61.97s
```

The two orderings separate as `lr*lr*wd`, while the kernels' own arithmetic error grows only as `lr`. So `test_optimizer32bit_weight_decay` runs at lr=0.1, wd=0.1, where the orderings are 1.0e-3 apart and agreement with the reference is 3.9e-5.

`test_optimizer8bit_weight_decay` cannot use a full run at all: the state quantization error is 6.6e-4 per step while the orderings differ by 1.6e-6, so the signal is buried. It instead takes one step from a fresh optimizer, where the state is still zero and therefore quantizes exactly, and the update reduces to `-lr*g/(|g| + eps)` (AdEMAMix also mixes in `alpha*(1-b3)*g`). At those hyperparameters the orderings are 1.0e-3 apart, and the fixed kernels sit 3.0e-6 from the closed form, which is `__powf` and `__fdividef` rather than the decay. Both tests are fp32; the 16-bit paths in `test_optimizer32bit` resynchronise the parameters every step, so at most one step of divergence is visible there and it stays inside the looser 16-bit tolerances.

## Verification

RTX 3090, sm_86, CUDA 12.4, torch 2.5.1+cu124, built from source with `-DCOMPUTE_CAPABILITY=86`. Only `csrc/kernels.cu` was reverted between the two runs; the tests were identical.

```
$ BNB_TEST_DEVICE=cuda pytest tests/test_optim.py -q -k weight_decay     # kernels.cu reverted
E  Greatest absolute difference: 0.0010029971599578857 at index (7, 2) (up to 0.0001 allowed)
E  Greatest absolute difference: 0.0010035037994384766 at index (329, 20) (up to 1e-05 allowed)
8 failed, 252 deselected in 16.44s

$ BNB_TEST_DEVICE=cuda pytest tests/test_optim.py -q -k weight_decay     # fix applied
8 passed, 252 deselected in 18.81s

$ BNB_TEST_DEVICE=cuda pytest tests/test_optim.py -q                     # fix applied, full module
221 passed, 12 skipped, 27 deselected, 194 warnings in 83.19s
```

Isolating the ordering on one step, same hardware, with the fix applied:

```
AdamW-32bit: |p-decay_first|=3.040e-06   |p-decay_after|=9.971e-04
AdamW-8bit:  |p-decay_first|=3.040e-06   |p-decay_after|=9.971e-04
```

On cpu and mps, which already had the correct ordering, the new tests pass unchanged (`8 passed` on cpu, `4 passed, 4 skipped` on mps; the 8-bit ones skip because `optimizer_update_8bit_blockwise` has no MPS kernel). To check they are not vacuous without a GPU, I also re-registered the cpu kernels with the CUDA ordering and confirmed all 8 fail.

`pre-commit run --all-files` passes.

## Not covered

`bnb.optim.Adam` applies decoupled decay in every backend, so with `weight_decay > 0` it matches `torch.optim.AdamW` rather than `torch.optim.Adam` (on cpu, 60 steps at lr=1e-2, wd=0.1: max 5.3e-1 from `torch.optim.Adam`, 1.5e-7 from `torch.optim.AdamW`). That is the separate point you raised in the issue and it is not touched here. The `skip_zeros` path still skips the decay along with the update when a gradient is exactly zero, which is existing behaviour I left alone. I verified on sm_86 only.

### DIFF
--- a/csrc/kernels.cu
+++ b/csrc/kernels.cu
@@ -690,11 +690,11 @@ __launch_bounds__(TH, 1) __global__ void kOptimizer32bit2State(
                 // nu update: nu = beta2 * nu + (1-beta2) * g^2
                 s2_vals[j] = (s2_vals[j] * beta2) + ((1.0f - beta2) * (float)g_vals[j] * (float)g_vals[j]);
 
-                p_vals[j] = (float)p_vals[j] - lr * (((s1_vals[j] / correction1) + (alpha * s3_vals[j])) /
-                                                     ((sqrtf(s2_vals[j]) / correction2) + eps));
-
                 if (weight_decay > 0.0f)
                     p_vals[j] = ((float)p_vals[j]) * (1.0f - (lr * weight_decay));
+
+                p_vals[j] = (float)p_vals[j] - lr * (((s1_vals[j] / correction1) + (alpha * s3_vals[j])) /
+                                                     ((sqrtf(s2_vals[j]) / correction2) + eps));
 
                 break;
             case ADAM:
@@ -702,11 +702,12 @@ __launch_bounds__(TH, 1) __global__ void kOptimizer32bit2State(
                 if (!skip_zeros || (skip_zeros && ((float)g_vals[j] != 0.0f))) {
                     s1_vals[j] = s1_vals[j] * beta1 + ((1.0f - beta1) * ((float)g_vals[j]));
                     s2_vals[j] = s2_vals[j] * beta2 + ((1.0f - beta2) * (((float)g_vals[j]) * ((float)g_vals[j])));
-                    p_vals[j] = ((float)p_vals[j]) +
-                                (update_scale * step_size * (s1_vals[j] / (sqrtf(s2_vals[j]) + (eps * correction2))));
 
                     if (weight_decay > 0.0f)
                         p_vals[j] = ((float)p_vals[j]) * (1.0f - (lr * weight_decay));
+
+                    p_vals[j] = ((float)p_vals[j]) +
+                                (update_scale * step_size * (s1_vals[j] / (sqrtf(s2_vals[j]) + (eps * correction2))));
                 }
                 break;
             }
@@ -1090,6 +1091,9 @@ __launch_bounds__(256, 3) __global__ void kOptimizerStatic8bit2StateBlockwise(
         for (unsigned int j = 0; j < N_PER_TH; j++) {
             // if(!skip_zeros || (skip_zeros && ((float)g_vals[j] != 0.0f)))
             if (!isnan((float)g_vals[j]) && !isinf((float)g_vals[j])) {
+                if (weight_decay > 0.0f)
+                    p_vals[j] = ((float)p_vals[j]) * (1.0f - (lr * weight_decay));
+
                 if (OPTIMIZER == ADEMAMIX) {
                     p_vals[j] =
                         T((float)p_vals[j] - lr * (((s1_vals[j] / correction1) + (alpha * s3_vals[j])) /
@@ -1099,9 +1103,6 @@ __launch_bounds__(256, 3) __global__ void kOptimizerStatic8bit2StateBlockwise(
                         (T)(((float)p_vals[j]) +
                             ((step_size * (__fdividef(s1_vals[j], (sqrtf(s2_vals[j]) + (correction2 * eps)))))));
                 }
-
-                if (weight_decay > 0.0f)
-                    p_vals[j] = ((float)p_vals[j]) * (1.0f - (lr * weight_decay));
             }
         }
 
--- a/tests/test_optim.py
+++ b/tests/test_optim.py
@@ -292,6 +292,107 @@ def test_lion32bit_weight_decay(dim1, dim2, gtype, device):
             p2.copy_(p1.data)
 
 
+# Decoupled weight decay multiplies the *previous* parameter and leaves the
+# gradient-based update unscaled: p_new = p_old*(1 - lr*wd) + step_size*update.
+# See Loshchilov & Hutter, arXiv:1711.05101 Algorithm 2 for AdamW and
+# Pagliardini et al., arXiv:2409.03137 for AdEMAMix; torch.optim.AdamW does the same.
+str2optimizers_weight_decay = {
+    "adamw": (torch.optim.AdamW, bnb.optim.AdamW),
+    "ademamix": (bnb.optim.ademamix._ReferenceAdEMAMix, bnb.optim.AdEMAMix),
+}
+
+str2optimizers_weight_decay_8bit = {
+    "adamw8bit_blockwise": bnb.optim.AdamW8bit,
+    "ademamix8bit_blockwise": bnb.optim.AdEMAMix8bit,
+}
+
+
+@pytest.mark.parametrize("optim_name", list(str2optimizers_weight_decay), ids=id_formatter("opt"))
+@pytest.mark.parametrize("dim1", [1024], ids=id_formatter("dim1"))
+@pytest.mark.parametrize("dim2", [32, 1024], ids=id_formatter("dim2"))
+@pytest.mark.parametrize("device", get_available_devices(), ids=id_formatter("device"))
+def test_optimizer32bit_weight_decay(dim1, dim2, optim_name, device):
+    """AdamW/AdEMAMix must apply decoupled weight decay to the previous parameter, not to
+    the parameter after the update has been added.
+
+    The two orderings differ by lr*wd*|update| per step, so the separation grows with
+    lr*lr*wd while the kernel's own arithmetic error only grows with lr. test_optimizer32bit
+    runs these optimizers at the default weight_decay of 0, and at the lr=1e-3, wd=0.01 an
+    Adam user would typically pick the orderings stay 1.7e-7 apart, inside its 1e-6
+    tolerance. At lr=0.1, wd=0.1 they are ~1e-3 apart while agreement with the reference
+    stays at 3.9e-5 on CUDA, which is the gap this tolerance sits in.
+
+    fp32 only: the ordering is a property of the float math inside the kernel and does not
+    depend on the parameter storage dtype, and the 16-bit variants of test_optimizer32bit
+    resynchronise the parameters after every step, so at most one step of divergence is
+    ever visible and it stays inside the looser 16-bit tolerances.
+    """
+    lr, weight_decay = 0.1, 0.1
+
+    p1 = torch.randn(dim1, dim2, device=device) * 0.1
+    p2 = p1.clone()
+
+    ref_cls, bnb_cls = str2optimizers_weight_decay[optim_name]
+    torch_optimizer = ref_cls([p1], lr=lr, weight_decay=weight_decay)
+    bnb_optimizer = bnb_cls([p2], lr=lr, weight_decay=weight_decay)
+
+    for _ in range(k):
+        g = torch.randn(dim1, dim2, device=device) * 0.01
+        p1.grad = g.clone()
+        p2.grad = g.clone()
+
+        bnb_optimizer.step()
+        torch_optimizer.step()
+
+        assert_most_approx_close(p1, p2, atol=1e-4, rtol=1e-3, max_error_count=0)
+
+
+@pytest.mark.parametrize("optim_name", list(str2optimizers_weight_decay_8bit), ids=id_formatter("opt"))
+@pytest.mark.parametrize("dim1", [1024], ids=id_formatter("dim1"))
+@pytest.mark.parametrize("dim2", [32, 1024], ids=id_formatter("dim2"))
+@pytest.mark.parametrize("device", get_available_devices(), ids=id_formatter("device"))
+def test_optimizer8bit_weight_decay(dim1, dim2, optim_name, device):
+    """Same decoupled weight decay requirement for the 8-bit blockwise kernels.
+
+    Comparing a full run against a 32-bit reference cannot see this: the state
+    quantization error is ~6.6e-4 per step while the two decay orderings differ by
+    ~1.6e-6. The first step is exempt, because the optimizer state is still zero and
+    therefore quantizes exactly, so one step from a fresh optimizer has a closed form:
+
+        m = (1-b1)g, v = (1-b2)g^2  =>  update = -lr*g/(|g| + eps)
+
+    and AdEMAMix also mixes in alpha*(1-b3)*g. Decoupled decay then gives
+    p1 = p0*(1 - lr*wd) + update, while applying decay afterwards scales the update
+    term by an extra (1 - lr*wd), which at these hyperparameters is 1.0e-3 away.
+
+    The tolerance is set above the 3.0e-6 the CUDA kernels leave against the closed form,
+    which comes from their fast intrinsics (__powf, __fdividef) rather than from the decay.
+    """
+    if device == "mps":
+        pytest.skip("8-bit optimizers are not implemented for MPS")
+
+    lr, weight_decay, eps = 0.1, 0.1, 1e-8
+
+    p0 = torch.randn(dim1, dim2, device=device) * 0.1
+    p = p0.clone()
+    g = torch.randn(dim1, dim2, device=device) * 0.01
+
+    bnb_optimizer = str2optimizers_weight_decay_8bit[optim_name]([p], lr=lr, weight_decay=weight_decay, eps=eps)
+    p.grad = g.clone()
+    bnb_optimizer.step()
+
+    update = -lr * g / (g.abs() + eps)
+    if optim_name.startswith("ademamix"):
+        group = bnb_optimizer.param_groups[0]
+        alpha = group["alpha"]
+        beta3 = group["betas"][2]
+        update = update * (1.0 + alpha * (1.0 - beta3))
+
+    expected = p0 * (1.0 - lr * weight_decay) + update
+
+    torch.testing.assert_close(p, expected, atol=1e-5, rtol=1e-4)
+
+
 @pytest.mark.parametrize("dim1", [1024], ids=id_formatter("dim1"))
 @pytest.mark.parametrize("dim2", [32, 1024, 4097], ids=id_formatter("dim2"))
 @pytest.mark.parametrize("gtype", [torch.float32, torch.float16], ids=describe_dtype)


ErenAta16 (commented):
This fixes the right side, and I want to be explicit about that because #2010 pointed at the other one.

## Which ordering is correct

I settled this against a source neither implementation consults, `torch.optim` itself. In torch 2.11.0, `AdamW` is `Adam(..., decoupled_weight_decay=True)`, and `_single_tensor_adam` does:

```python
if weight_decay != 0:
    if decoupled_weight_decay:
        # Perform stepweight decay
        param.mul_(1 - lr * weight_decay)
```

That runs before the moment updates and before the final `addcdiv_`, so the parameter is shrunk first and the gradient-based update is added afterwards, unscaled:

```
p_new = p_old * (1 - lr*wd) + step_size * update
```

which is Algorithm 2 of Loshchilov & Hutter and exactly what this PR moves the CUDA kernels to. The old CUDA ordering multiplied the whole thing, update included, by `(1 - lr*wd)`, so the effective step size shrank with the decay rate.

## Correcting my own issue

#2010 framed the CUDA kernel as the reference and the `default`/`triton` backends as the deviation. The divergence I reported was real and the arithmetic in that issue is right, but the conclusion about which side to change was mine to get right and I got it backwards. `default` and `triton` already matched torch; CUDA was the outlier. Fixing the kernels, as you have done here, is the correct direction, and the Python backends should be left alone.

## The fix is complete, which I checked rather than assumed

`kernels.cu` applies `weight_decay` in four places. This PR changes two of them. The other two should stay exactly as they are, so nobody needs to wonder whether they were missed:

```
kOptimizer32bit2State                693, 706   changed here
kOptimizer32bit1State                866, 885   correct as is
kOptimizerStatic8bit2StateBlockwise  1094       changed here
kOptimizerStatic8bit1StateBlockwise  1227       correct as is
```

The 1-state kernels split by optimizer rather than applying one rule:

```cpp
if (weight_decay > 0.0f && OPTIMIZER != LION)
    g_vals[j] = (float)g_vals[j] + (((float)p_vals[j]) * weight_decay);
```

MOMENTUM, RMSPROP and ADAGRAD fold the decay into the gradient, which is coupled L2 decay and the correct form for those optimizers; there is no ordering question there because the decay never touches the parameter directly. LION is excluded from that branch and applies decoupled decay to the parameter **before** its sign update, which is already the ordering this PR is establishing for the 2-state kernels. The 8-bit 1-state kernel at 1227 makes the same split with the same result.

So the two-of-four count is the right count, not a partial fix.

## Why the existing suite could not have caught it

Your docstring argues the separation grows with `lr*lr*wd` while the kernel's own arithmetic error grows with `lr`. That reproduces. Running both orderings side by side, 50 steps over 4096 parameters, and taking the max absolute divergence:

```
lr=0.001  wd=0.01  ->  2.8e-07   inside a 1e-6 tolerance
lr=0.01   wd=0.01  ->  2.8e-05   outside
lr=0.1    wd=0.1   ->  2.1e-02   outside
```

My absolute figures differ from the ones in your docstring because the harness differs, step count and gradient scale included, so treat this as corroborating the shape rather than reproducing the exact number. The conclusion holds either way: at the hyperparameters an Adam user actually picks, the two orderings sit inside the tolerance `test_optimizer32bit` uses, and that suite runs these optimizers at `weight_decay=0` anyway. Choosing `lr=0.1, wd=0.1` for the new test is what makes it discriminating rather than a second copy of the existing one.

Citing the papers next to the reference implementations in the test file is worth keeping. The next person to look at this will have the same "which side is right" question I had, and the answer is now in the tree instead of in a review thread.

Nothing blocking from me. Thanks for picking this up, and for taking it in the direction the reference actually supports rather than the one the issue suggested.

ErenAta16 (commented):
I filed #2010, so this is the fix for my report. Please carry it. You have a CUDA card and I do not, and the GPU verification here is the half I could not have done.

I checked the ordering independently rather than re-read the diff, using `torch.optim.AdamW` as the referee since neither ordering consults it. Float64, lr=1e-3, wd=0.01, 20 steps, same gradients into both:

```
decay BEFORE update (this PR)        max|diff to torch.optim.AdamW| = 0.000e+00
decay AFTER  update (current kernel) max|diff to torch.optim.AdamW| = 1.126e-07
```

Exactly zero, not merely closer. That is the part I would put in the PR description: the two orderings are not two approximations of AdamW where one is more accurate, one of them **is** AdamW and the other is a different algorithm. The extra `(1 - lr*wd)` on the gradient term is a factor Algorithm 2 does not contain, and no tolerance argument makes it correct.

The one-step algebra lands where you say:

```
decay_after = (p - ss*u)*(1 - lr*wd)
decay_first =  p*(1 - lr*wd) - ss*u
gap         = ss*u*lr*wd      predicted 1.500e-06, measured 1.500e-06
```

**Your scaling claim is the load-bearing one and it holds.** This is what explains why the suite never caught it, so it is worth having measured rather than argued:

```
wd=0.01, 20 steps
  lr=1e-4   gap=1.126e-09
  lr=1e-3   gap=1.126e-07
  lr=1e-2   gap=1.125e-05
  lr=1e-1   gap=1.113e-03
```

A hundredfold per tenfold in lr, so the gap really does grow as `lr^2 * wd` while the kernels' own error grows as `lr`. That is exactly why `test_optimizer32bit_weight_decay` has to sit at lr=0.1 and why the 8-bit test has to take a single step from zeroed state: at the hyperparameters a user actually picks, the defect is smaller than the arithmetic noise, and any test written at those values would be measuring nothing.

My number is 1.126e-07 where you quote 1.7e-07 for the same configuration. Different bias-correction placement in my scratch reimplementation, most likely; same order and same conclusion, so I mention it only so nobody reads the mismatch later as a disagreement.

Two things I would keep exactly as they are:

Re-registering the cpu kernels with the CUDA ordering to confirm the new tests fail is the step most people skip. Without it "8 passed on cpu" only tells you the tests run, not that they discriminate.

Scoping out `bnb.optim.Adam` matching `torch.optim.AdamW` rather than `torch.optim.Adam` is right. That is a naming and contract question and folding it into a kernel reorder would make both harder to review.

One small thing, non-blocking. The decay now sits inside the `if (!skip_zeros || ...)` guard in the ADAM branch, which it also did before, so `skip_zeros=True` still skips the decay for an exactly-zero gradient. Unchanged behaviour and correctly left alone, but now that the decay is the first statement rather than the last it reads more like a deliberate coupling than it did. One line saying it is inherited, not chosen, would save the next reader the archaeology.

@egeozkoc since you confirmed the 8-bit blockwise site, the reorder there covers both the ADEMAMIX and ADAM branches because it moved above the `if (OPTIMIZER == ADEMAMIX)`.

matthewdouglas:
Hi @yentur, thanks for the PR.

I wanted to explain why the issue remained open for a while. [In the issue #2010, I had stated](https://github.com/bitsandbytes-foundation/bitsandbytes/issues/2010#issuecomment-5005343450):
> I want to think on this a little more. If I do make a correctness fix here, it's likely going to result in real behavior changes that we'll need to call out. With that said, right now I do lean towards correcting it.

So, I had not yet committed to this change, just indicated where I was leaning. I wasn't prepared to introduce a change like this into our release cycle at the time. Then I went on vacation. I will however consider this PR for v0.51.0.

ErenAta16:
@matthewdouglas on the part you flagged as needing a call-out, here are numbers for how large the behaviour change actually is, in case they help size the release note.

The two orderings differ by `lr*wd` applied to the update term, so the gap scales with the product rather than with either alone. Running both orderings side by side over 4096 parameters, max absolute divergence and the same figure relative to the parameter scale:

```
     lr     wd   steps      max|dp|    relative
  0.001   0.01     100    3.949e-07      0.00%
  0.001   0.01    1000    1.192e-06      0.00%
  0.001    0.1     100    3.934e-06      0.01%
  0.001    0.1    1000    1.153e-05      0.01%
 0.0001   0.01     100    4.191e-09      0.00%
 0.0001   0.01    1000    1.211e-08      0.00%
 0.0003    0.1     100    3.548e-07      0.00%
 0.0003    0.1    1000    1.062e-06      0.00%
```

At the hyperparameters people actually train AdamW with, the trajectory difference stays at or below 0.01% of parameter scale even after a thousand steps, and it grows sub-linearly with step count rather than compounding. It only becomes visible at deliberately large `lr*wd`: at `lr=0.1, wd=0.1` the same measurement is around 2e-2, which is what makes the new test in this PR discriminating.

So the correctness argument and the disruption argument point in different directions here: the ordering was genuinely wrong against `torch.optim.AdamW`, and almost nobody's run will move enough to notice. That seems like a comfortable thing to put in a release note rather than a scary one.

One correction I owe you, since it is my issue you quoted. #2010 named the CUDA kernel as the reference and the `default`/`triton` backends as the deviation. That was backwards. `torch.optim`'s `_single_tensor_adam` shrinks the parameter before adding the unscaled update, which is the Python backends' form, so CUDA was the outlier and this PR fixes the right side. I have corrected the issue body accordingly.

I also checked whether this PR is complete: `kernels.cu` applies `weight_decay` in four places, this changes two, and the other two are the 1-state kernels where MOMENTUM/RMSPROP/ADAGRAD use coupled L2 decay folded into the gradient and LION already decays before its sign update. Details are in my earlier comment on this PR.

PR #2044, opened 2026-08-19
Fix mixed INT8 FakeTensor output metadata

tandede:
## Summary

- make the mixed INT8 FakeTensor implementation return `subA` with the same shape and dtype contract as eager execution
- keep `subA` two-dimensional when there are zero outlier columns, including when `outlier_cols` is omitted
- add CPU and MPS operation checks for non-empty, empty, and omitted outlier-column inputs

## Root cause

The fake implementation created `subA` as a fresh one-dimensional `int64` tensor. Eager execution instead returns the selected input columns, which are two-dimensional and inherit the input dtype. This caused `torch.library.opcheck` to fail with mismatched output metadata. The eager no-outlier branch also returned a one-dimensional empty tensor, making the output rank depend on input data.

## Impact

The custom operation now has stable output metadata for FakeTensor and graph-capture consumers. The numerical mixed INT8 result is unchanged.

## Tests

- `pytest -q tests/test_ops.py` (649 passed, 69 skipped, 60 xfailed)
- focused mixed INT8 operation checks on CPU and MPS (4 passed)
- full-graph `Linear8bitLt` compile checks with threshold 6.0 on CPU and MPS (2 passed)
- `pre-commit run --files bitsandbytes/_ops.py bitsandbytes/backends/default/ops.py tests/test_ops.py`

### DIFF
--- a/bitsandbytes/_ops.py
+++ b/bitsandbytes/_ops.py
@@ -27,8 +27,8 @@ def _(
 
     out = torch.empty(shapeC, device=A.device, dtype=A.dtype)
 
-    outlier_cols = torch.library.get_ctx().new_dynamic_size()
-    subA = A.new_empty(outlier_cols, dtype=torch.int64)
+    num_outlier_cols = 0 if outlier_cols is None else outlier_cols.shape[0]
+    subA = A.new_empty((A.shape[0], num_outlier_cols))
 
     return out, subA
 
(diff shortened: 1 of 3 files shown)


tandede:
Hi maintainers, I’m following up on this PR in case it was missed. It fixes the output metadata produced for mixed INT8 FakeTensor operations and is ready for review. I’m happy to address any feedback or provide additional validation if needed. Thank you!

PR #2048, opened 2026-08-20
Fix CPU dequantize_4bit output shape for 1-D inputs

### LINKED ISSUE #2047: CPU dequantize_4bit returns shape (1, n) for even-length 1-D inputs; all other backends return (n,)
The native CPU path of `bitsandbytes::dequantize_4bit` reshapes a 1-D `shape` argument to `(1, shape[0])` before allocating the output (`bitsandbytes/backends/cpu/ops.py`, `if len(shape) == 1:`), so a quantize/dequantize round trip of a 1-D tensor comes back 2-D on CPU.

```python
import torch, bitsandbytes.functional as F
A = torch.randn(4096)
q, state = F.quantize_4bit(A, blocksize=64, quant_type="nf4")
F.dequantize_4bit(q, state).shape   # torch.Size([1, 4096]) on CPU
```

Every other implementation returns exactly `shape`: the CUDA backend allocates `torch.empty(shape)`, the default and MPS backends reshape to `shape`, and the registered fake kernel returns `torch.empty(shape)` — so eager CPU also contradicts the op's fake-tensor contract under torch.compile. The smoking gun that this is a bug and not a design choice: the CPU backend is inconsistent with itself — an odd-length 1-D input (n=4095) takes the generic fallback and correctly returns `(4095,)`, while n=4096 returns `(1, 4096)`.

Present since the native CPU 4-bit kernel landed in #1789. Reproduced on macOS arm64 (NEON path), torch 2.13, main @ 95f9af3. Fix is a 2-line deletion — `m = prod(shape[:-1])` already evaluates to 1 for 1-D inputs and the kernel writes a flat contiguous buffer; PR incoming.

2sumtech:
Fixes #2047

The native CPU `dequantize_4bit` mutated a 1-D `shape` to `(1, n)` before allocating the output, so even-length 1-D inputs dequantized to 2-D. The `(1, n)` view is only needed for the C++ kernel's 2-D indexing, and `m = prod(shape[:-1])` is already 1 for 1-D inputs — this PR drops the mutation so the output keeps the caller-provided shape, matching the CUDA backend, the default/MPS backends, the registered fake kernel, and the CPU backend's own odd-length fallback path.

- Output values are bit-for-bit identical before and after over 36 configurations (fp32/fp16/bf16 × nf4/fp4 × blocksize 64/128 × n 128/256/4096); only the shape changes, `(1, n) -> (n,)`.
- New regression test `test_dequantize_4bit_1d_shape` (24 params): fails 12/12 on CPU without the fix, passes 24/24 (cpu+mps) with it.
- `Test4bitBlockwiseQuantOps`: 536 passed, 24 skipped, 60 xfailed. `test_functional.py -k 4bit`: 1958 passed. Full pre-commit suite passes on touched files.
- Verified on macOS arm64 (NEON path) + MPS; the fix is backend-dispatch Python only, no csrc change. I could not run the CUDA/AVX-512 paths on this machine, but they are untouched.

Disclosure: prepared with AI assistance (Claude Code); I reviewed the change and take responsibility for it.


matthewdouglas:
Thank you for the PR. Happy to accept, but do not think the regression test is necessary. Please remove that part.

2sumtech:
done, removed the test and rebased on main

PR #2065, opened 2026-08-30
Fix CPU Hub kernel backend selection

### LINKED ISSUE #2064: CPU `gemm_4bit_forward` kernel is requested without `backend="cpu"`, so it never loads on a CUDA torch build
`bitsandbytes/backends/cpu/ops.py` requests the fused CPU 4-bit gemm kernel without specifying a
backend. `kernels` therefore infers the backend from the installed torch build, which on any CUDA
wheel is a CUDA backend — and `kernels-community/quantization-bitsandbytes` publishes CPU-only build
variants. Every variant is rejected, `get_kernel` raises, and the `except Exception` handler logs a
warning on every process start on any host whose CPU reports AVX512-BF16.

This is separate from #1972, which fixed the repo id and the missing `version` in the same call. With
0.50.2 the id and version are correct; the backend mismatch remains.

### Versions

- `bitsandbytes` 0.50.2
- `kernels` 0.16.1
- `torch` 2.12.1+cu130 (`torch.version.cuda == "13.0"`, `compiled_with_cxx11_abi() == True`)
- Python 3.12, Linux x86_64
- CPU reports `avx512_bf16`, so `has_avx512bf16()` is True and the guarded block executes

### What happens

```
Failed to load CPU gemm_4bit_forward from kernels-community:
Cannot find a build variant for this system in kernels-community/quantization-bitsandbytes
(revision: 4fc7d1e1abacca3985171c276723e1c4b4f30a81):
torch212-cxx11-cpu-x86_64-linux: backend (cpu) does not match ...
```

### Reproduction

```python
from kernels import get_kernel, has_kernel
from kernels.backends import _select_backend

REPO = "kernels-community/quantization-bitsandbytes"
print(_select_backend(None).variant_str)      # -> cu130
print(has_kernel(REPO, version=1))            # -> False

get_kernel(REPO, version=1)                   # -> FileNotFoundError, backend (cpu) does not match
get_kernel(REPO, version=1, backend="cpu")    # -> loads torch212-cxx11-cpu-x86_64-linux cleanly
```

The second call is the whole report: the variant exists and is loadable; only the inferred backend is
wrong for it.

### Why the inference cannot be right here

The kernel is registered for the CPU device —
`@register_kernel("bitsandbytes::gemv_4bit", "cpu")` — so the backend it needs is known statically at
the call site. `kernels` cannot infer it, because it infers from the torch build rather than from the
device the caller intends. A CUDA torch build is the normal case for anyone who also uses this
library on a GPU, so on those installations the CPU kernel is unreachable by construction.

### Suggested fix

Pass the backend explicitly at the call site:

```python
gemm_4bit_forward_kernel = get_kernel(
    "kernels-community/quantization-bitsandbytes", version=1, backend="cpu"
)
```

### Impact

Cosmetic but persistent: the fused path silently falls back to the unfused reference implementation,
and the warning appears on every process start — including in headless/containerised runs, where it
lands in every log a batch job produces. Downstream projects that use bitsandbytes only for CUDA
quantization end up suppressing the logger, which also hides `bitsandbytes.cextension`'s CUDA-setup
warnings unless they take care to target the submodule.

### Remark

This issue was formated and investigated with the help of AI.

SiluPanda:
## Description

The fused CPU 4-bit GEMM loader knows it is registering a CPU kernel, but it let `kernels` infer the backend from the installed PyTorch build. With a CUDA PyTorch wheel, that selects CUDA even though `kernels-community/quantization-bitsandbytes` publishes a CPU build, so the fused kernel is rejected and bitsandbytes falls back after logging a warning.

Pass `backend="cpu"` explicitly when loading the Hub kernel. Explicit backend selection requires `kernels >= 0.13.0`; older or missing versions continue to use the existing best-effort warning and fallback behavior.

## Tests

- `.venv/bin/pytest tests/test_ops.py::Test4bitBlockwiseQuantOps::test_gemv_4bit -k cpu -v --tb=short` (`48 passed, 12 skipped, 60 deselected`)
- `pre-commit run --all-files` (all hooks passed)

Fixes #2064

AI assistance: This change was implemented with OpenAI Codex and independently reviewed and validated before submission.


bitsandbytes/backends/cpu/ops.py
```diff
@@ -16,6 +16,24 @@
...
+        from kernels import get_kernel
+
+        return get_kernel(
+            "kernels-community/quantization-bitsandbytes",
+            version=1,
+            backend="cpu",
+        ).gemm_4bit_forward
+    except Exception as exc:  # pragma: no cover - best effort fallback
+        logger.warning(
+            "Failed to load CPU gemm_4bit_forward from kernels-community: %s. Please make sure you already `pip install kernels` and the kernels >= 0.11.1",
```
matthewdouglas: Realistically this new kwarg bumps the minimum `kernels` version to 0.13.0. Can you update to reflect that?

Secondly, since we do this only once anyway, I'm fine to just keep this whole thing inline as it was.

tests/test_cpu_ops.py
```diff
@@ -0,0 +1,44 @@
...
+
+    kernels.get_kernel = fake_get_kernel
+    monkeypatch.setitem(sys.modules, "kernels", kernels)
+
+    with caplog.at_level(logging.WARNING, logger=cpu_ops.__name__):
+        kernel = cpu_ops._load_gemm_4bit_forward_kernel()
+
+    assert kernel is None
+    assert "Failed to load CPU gemm_4bit_forward from kernels-community" in caplog.text
+    assert "missing cpu variant" in caplog.text
```
matthewdouglas: These tests are just noise to me. For this kind of change I think it's completely unnecessary to do a regression test on it.

matthewdouglas:
Hi, thanks for flagging this issue. I understand this could happen when users on a CUDA-enabled PyTorch build are doing operations on both their accelerators and their CPU. It seems like a reasonable change to me. Just a few nits before I'm ready to merge.

bitsandbytes/backends/cpu/ops.py
SiluPanda: thanks. I’ve put the loader back inline and updated the minimum `kernels` version to 0.13.0 in both the warning and the PR description.

tests/test_cpu_ops.py
SiluPanda: Makes sense. I removed the dedicated tests and kept the change focused. I also ran the existing CPU `gemv_4bit` coverage and the full pre-commit suite after the cleanup.

SiluPanda:
> Hi, thanks for flagging this issue. I understand this could happen when users on a CUDA-enabled PyTorch build are doing operations on both their accelerators and their CPU. It seems like a reasonable change to me. Just a few nits before I'm ready to merge.

Hello @matthewdouglas 

Thank you so much for your review. Whenever possible for you, would appreciate if you could take another look at the pr. Thank you

Issue #2011, opened 2026-07-17
Offer to help improve the MPS backend fallbacks


egeozkoc:
I'd like to help push the MPS backend off the 🐢 status in the README, and I'm set up to work on it directly (Apple Silicon, can build and test locally). I've contributed on the MPS side before (#1994) and on optimizers (#1998), so I'm somewhat familiar with the layout.

From reading the current backend (#1983, #2004), my understanding of the design is: try the `kernels-community/bitsandbytes-mps` Hub kernels when available, and fall back to pure PyTorch otherwise, with those fallbacks being what the 🐢 reflects. I don't want to cut across your in-flight work, so rather than guess, I wanted to ask where contributions would be most useful.

A few directions I could take on, if any are helpful:
- Profiling and speeding up specific PyTorch fallbacks in `backends/mps/ops.py`, in the spirit of #1960.
- Filling in MPS-specific fallbacks for ops that currently only hit the generic slow path.
- Extending MPS test coverage and correctness for the fallback paths.

If the 8-bit optimizer path (the 🚧) is something you'd want help with, I'm glad to look at that too, though I'd want to check it fits your plans first. Let me know what would be most useful, or if you'd rather I stay in the smaller correctness and coverage lane for now.

matthewdouglas:
Hi @egeozkoc,

Thanks for asking!

The README's 🐢 is a little out of date. With the newer Hub kernels the 4bit quantize/dequantize/gemv ops are in a much better place. I plan to release what we have in v0.50.0 very soon, so will likely be updating this in the README even if best performance needs macOS 26 + `kernels`.

I do have a plan already for the optimizers, which will be part of a future v0.51.0 release. I would ask for now that you hold off on that. Correctness is going to come first, then further perf if needed. I'll be implementing it similar to the CPU backend, which will rely on the separate `quantize_blockwise`, `dequantize_blockwise` ops, and do the rest in PyTorch. I think this will be good enough to start before jumping further to e.g. a fully fused kernel like we have on the CUDA side.

I did spend some time benchmarking as part of #1960 and #1983. But if you want to share any benchmarks, profile data, etc, then that could be useful. Especially if you've got different hardware than I do (M4, 10-core). 

One obvious area that underperforms is the LLM.int8() ops. It's not exactly a priority, but there's opportunity there I'm sure. Another op that I know has room for improvement is `quantize_blockwise`.

[...]

PR #2069, opened 2026-09-03
fix(optim): fall back to non-paged state on all devices without managed memory

egeozkoc:
## Summary

Paged optimizers (`PagedAdamW`, `PagedAdam`, `PagedLion`, `PagedAdEMAMix`, and their 8-bit variants) only fell back to regular state buffers on `cpu`. On any other device without managed-memory support they reached `F.get_paged`, which calls `lib.cget_managed_ptr`, and failed with a misleading error:

```
RuntimeError: Method 'cget_managed_ptr' not available in CPU-only version of bitsandbytes.
Reinstall with GPU support or use CUDA-enabled hardware.
```

On MPS this affects every `paged_*` optimizer in `test_optim.py::test_optimizer32bit` (45 cases with `BNB_TEST_DEVICE=mps`), and any user who picks e.g. `optim="paged_adamw_32bit"` in a Transformers `TrainingArguments` on Apple Silicon.

This PR generalizes the existing CPU fallback in `Optimizer8bit.get_state_buffer` to any device type whose native library does not provide `cget_managed_ptr`. Managed memory is implemented for CUDA/ROCm (`cudaMallocManaged`) and XPU (`sycl::malloc_shared`), so those keep the paged path; everything else gets a plain tensor plus a one-time warning naming the device.

## Changes

- `bitsandbytes/optim/optimizer.py`: fallback condition is now `device.type not in ("cuda", "xpu")`; warning text names the device; the once-only flag is renamed from `_cpu_paged_warned` to `_paged_fallback_warned`.
- `tests/test_optim.py`: the `paged_*` skip in `test_optimizer32bit` now applies to all non-paging devices instead of only CPU.

## Testing

Apple M5, macOS 26, torch 2.12.1:

- `BNB_TEST_DEVICE=mps pytest tests/test_optim.py -k "paged or test_optimizer32bit"`: 72 passed, 72 skipped (the 45 paged cases previously errored).
- `BNB_TEST_DEVICE=cpu pytest tests/test_optim.py -k "paged or test_optimizer32bit"`: 72 passed, 72 skipped.
- `pre-commit run --all-files`: clean.

Not tested on CUDA/XPU hardware; the paged path there is unchanged apart from the condition being expressed as an allow-list of device types.

Ref #2011 (MPS backend work). The 8-bit optimizer op itself is still unimplemented on MPS; that is a separate item planned for v0.51.0 and is not touched here.


matthewdouglas:
Thanks for the PR. Given that paged optimizers are only actually supported on CUDA/ROCm/XPU, it seems superfluous. But considering we'd later implement optimizers for MPS or fallback, I suppose it's reasonable in that respect.

With that said, the extra test for regression against previous behavior, literally saying "previously" in the comments, IMHO seems just unnecessary. I do not understand why the agents love adding these pointless tests so much, but please remove that.

egeozkoc:
Removed the test, and dropped a now-stale reference to it from the skip message. The remaining test change is just widening the existing `paged_*` skip in `test_optimizer32bit` from CPU to any non-paging device.

